PR #15, opened 2019-08-08
Add docs for MaybeSentinel and RemovalSentinel

bgw:
## Summary

These were referred to by other parts of the documentation, and provide
core parts of our functionality, so we should have explicit
documentation for them.

## Test Plan

https://1371-200896124-gh.circle-artifacts.com/0/doc/nodes.html#maybe-sentinel
https://1371-200896124-gh.circle-artifacts.com/0/doc/visitors.html#libcst.RemovalSentinel

### DIFF
--- a/libcst/_maybe_sentinel.py
+++ b/libcst/_maybe_sentinel.py
@@ -9,9 +9,48 @@ from enum import Enum, auto
 
 class MaybeSentinel(Enum):
     """
-    A MaybeSentinal value is used as the default constructor for some attributes to
-    denote that on generating code we should optionally include this element in order
-    to generate valid code.
+    A :class:`MaybeSentinel` value is used as the default value for some attributes to
+    denote that when generating code (when :attr:`Module.code` is evaluated) we should
+    optionally include this element in order to generate valid code.
+
+    :class:`MaybeSentinel` is only used for "syntactic trivia" that most users shouldn't
+    care much about anyways, like commas, semicolons, and whitespace.
+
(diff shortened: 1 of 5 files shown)


libcst/_maybe_sentinel.py
```diff
@@ -9,9 +9,48 @@
...
+
+    This makes manual node construction easier, but it also means that we safely add
+    arguments to a preexisting function call without manually fixing the commas:
+
+    >>> import libcst as cst
+    >>> fn_call = cst.parse_expression("fn(1, 2)")
+    >>> new_fn_call = fn_call.with_changes(
+    ...     args=[*fn_call.args, cst.Arg(cst.Integer("3"))]
+    ... )
+    >>> dummy_module = cst.parse_module("")  # we need to use Module.code_for_node
```
jimmylai: Shall we provide some helper to render code of a Node (not Module) easily?
What's the reason we don't provide the code property in Node or Statement?

bgw: We can only generate code in the context of a module because the module contains information about the default indentation level, newline, etc.

However (as this example demonstrates) it'd be useful to have a `.code` property on every node for debugging purposes. I just don't know how we can communicate these pitfalls to the developer.

It's also a little weird that `parse_expression` and `parse_statement` are lossy and not entirely reversible, but that's by design. If you have a module using `\r\n` as it's newline format and you use `parse_statement` to insert a line of code with `\n` in it, I'd want that `\n` to be replaced with `\r\n` (which is the current behavior).

PR #23, opened 2019-08-13
Codegen visitor functions

DragonMinded:
## Summary

Implement codegen for visit_* and leave_* methods so that pyre can warn about incorrect overloads and incorrect types. This includes code which does the actual codegen, a test that verifies that the codegen isn't out of date with the LibCST repo, and a tox environment/helper script for performing the codegen in a roughly atomic manner.

## Test Plan

`tox`, as well as creating scenarios that 1) caused codegen failure to verify that we do the right thing when failing to codegen and 2) make us need a codegen, thus triggering a test failure.


libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,164 @@
...
+        yield base
+
+
+def _get_base(node: Type[cst.CSTNode]) -> Type[cst.CSTNode]:
+    bases = set(_get_bases(node))
+    if not bases:
+        raise Exception(f"Couldn't get any bases for {node.__name__}!")
+    if len(bases) == 1:
+        for b in bases:
+            return b
```
rayjzeng: This is a small nit but in my opinion rewriting this as `return bases.pop()` instead of a loop improves clarity.
Also applies a few lines down in this function.

libcst/codegen/generate.py
```diff
@@ -0,0 +1,104 @@
...
+
+def main(cli_args: List[str]) -> int:
+    # Parse out arguments, run codegen
+    parser = argparse.ArgumentParser(description="Generate code for libcst.")
+    parser.add_argument(
+        "system",
+        metavar="SYSTEM",
+        help='System to generate code for. Valid values include: "visitors"',
+        type=str,
+    )
```
rayjzeng: Is the intention for this script to be used for other codegen within libcst? I guess it doesn't cost anything to add the argument anyhow.

libcst/_typed_visitor.py
```diff
@@ -0,0 +1,2177 @@
+# This source code is licensed under the MIT license found in the
```
jimmylai: This line is required on every source code. It should only apply to generated code since we checked it in unless you explicitly exclude it using blacklist.
`# Copyright (c) Facebook, Inc. and its affiliates.`

libcst/codegen/generate.py
DragonMinded: Yes. There's more stuff coming down the line that will need codegen, so I tried to set this up as the entrypoint for such pieces.

libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,164 @@
+# This source code is licensed under the MIT license found in the
```
jimmylai: # Copyright (c) Facebook, Inc. and its affiliates.

libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def _get_base(node: Type[cst.CSTNode]) -> Type[cst.CSTNode]:
+    bases = set(_get_bases(node))
+    if not bases:
+        raise Exception(f"Couldn't get any bases for {node.__name__}!")
+    if len(bases) == 1:
+        for b in bases:
+            return b
+    bases = {b for b in bases if b is not node}
```
jimmylai:
```suggestion
    bases.discard(node)
```

libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,164 @@
...
+
+    new_bases = set()
+    for base in bases:
+        new_bases.add(_get_base(base))
+
+    if len(new_bases) > 1:
+        raise Exception(f"Got too many bases for {node.__name__}!")
+    if len(new_bases) == 1:
+        for b in new_bases:
+            return b
```
jimmylai:
```suggestion:-1
            return b.pop()
```

jimmylai:
```suggestion:-2+1
            return b.pop()
```

libcst/codegen/generate.py
```diff
@@ -0,0 +1,104 @@
...
+    # So, this attempts to keep the repo in a working state for as many operations
+    # as possible.
+    base = os.path.abspath(
+        os.path.join(os.path.dirname(os.path.abspath(__file__)), "../")
+    )
+    visitors_file = os.path.join(base, "_typed_visitor.py")
+    with open(visitors_file, "r") as orig:
+        with open(f"{visitors_file}.bak", "w") as bkp:
+            bkp.write(orig.read())
+            bkp.close()
```
jimmylai:
```
import shutil
shutil.copyfile(visitors_file, f"{visitors_file}.bak")
```

PR #26, opened 2019-08-14
Add metadata documentation

rayjzeng:
## Summary
Adds documentation to the metadata interface for LibCST and cleans up some internal code.

## Test Plan
Ran tox and pyre.

### DIFF
--- a/docs/source/metadata.rst
+++ b/docs/source/metadata.rst
@@ -1,16 +1,44 @@
+.. _libcst-metadata:
+
 Metadata
 ========
 
-.. autoclass:: libcst.BatchableCSTVisitor
-.. autofunction:: libcst.visit_batched
+LibCST ships with a metadata interface that defines a standardized way to
+associate nodes in a CST with arbitrary metadata. The metadata interface is
+designed to be declarative and type safe.
+
+Providing Metadata
+------------------
+
+Metadata is generated through provider classes that can be declared as a dependency
+by a subclass of :class:`~libcst.MetadataDependent`. These providers are then
+resolved automatically using methods provided by :class:`~libcst.MetadataWrapper`.
 
 .. autoclass:: libcst.BaseMetadataProvider
 .. autoclass:: libcst.VisitorMetadataProvider
 .. autoclass:: libcst.BatchableMetadataProvider
 
-.. autoclass:: libcst._MetadataDepedent
+Accessing Metadata
+------------------
+
+.. autoclass:: libcst.MetadataDependent
+
+.. autoclass:: libcst.MetadataWrapper
+
+Position Metadata
+-----------------
+
+Position (line and column numbers) metadata are accessible through the metadata
+interface by declaring the one of the following providers as a dependency. For
+most cases, :class:`~libcst.SytacticPositionProvider` is what you probably want.
 
 .. autoclass:: libcst.BasicPositionProvider
 .. autoclass:: libcst.SyntacticPositionProvider
 
-.. autoclass:: libcst.MetadataWrapper
+Accessing position metadata through the :class:`~libcst.MetadataDepedent`
+interface will return a :class:`~libcst.CodeRange` object.
+
+.. autoclass:: libcst.CodeRange
+	:members:
+.. autoclass:: libcst.CodePosition
+	:members:
--- a/docs/source/metadata.rst
+++ b/docs/source/metadata.rst
@@ -7,8 +7,9 @@ LibCST ships with a metadata interface that defines a standardized way to
 associate nodes in a CST with arbitrary metadata while maintaining the immutability
 of the tree. The metadata interface is designed to be declarative and type safe.
 Here's a quick example of using the metadata interface to get line and column
-numbers of nodes:
+numbers of nodes through the :class:`~libcst.SyntacticPositionProvider`:
 
+.. _libcst-metadata-position-example:
 .. code-block:: python
 
     class NamePrinter(cst.CSTVisitor):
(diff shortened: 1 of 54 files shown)


docs/source/metadata.rst
```diff
@@ -0,0 +1,68 @@
...
+.. autoclass:: libcst.CodePosition
+    :members:
+
+Usage
+-----
+
+Here's an example of a visitor that prints out the starting position of every
+:class:`~libcst.Name` node.
+
+.. code-block:: python
```
jimmylai: Can you create a notebook for this example like the Usage page or add this to the usage page since it's small example? So that we run the code (to test the code still run), generate output in document (which is more straightforward) and allow user to play around easily.

docs/source/usage.ipynb
```diff
@@ -23,7 +24,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
```
jimmylai: We don't want to store the execution result in notebook. That way the nbsphinx won't actually run the notebook.
Please clear all output and then save again.

docs/source/visitors.rst
```diff
@@ -43,3 +43,14 @@ An example Python REPL using the above visitor is as follows::
     >>> _ = demo.visit(FooingAround())
     'abc'
     '123'
+
+
+Batched Visitors
+----------------
+
+A batchable visitor classs is provided to facilitate performing operations that
```
jimmylai: classs -> class

libcst/_nodes/_internal.py
```diff
@@ -47,19 +49,39 @@
...
 
-    @classmethod
-    def create(cls, start: Tuple[int, int], end: Tuple[int, int]) -> "CodeRange":
-        return CodeRange(CodePosition(start[0], start[1]), CodePosition(end[0], end[1]))
+    @overload
+    def __init__(self, start: CodePosition, end: CodePosition) -> None:
+        ...
+
+    @overload
+    def __init__(self, start: Tuple[int, int], end: Tuple[int, int]) -> None:
```
jimmylai: Why don't we just use CodePosition? Then we don't need to overload with tuple.
CodePosition already been used in multiple places.
https://github.com/Instagram/LibCST/blob/a843500b72ff1ea39265ebea840dba1e5ba50ffb/libcst/_nodes/_base.py#L263
https://github.com/Instagram/LibCST/blob/a843500b72ff1ea39265ebea840dba1e5ba50ffb/libcst/_nodes/tests/test_internal.py#L69

jimmylai: If we prefer to use tuple to be simple, maybe we can just remove CodePosition?

bgw: I'm fine with this.

`CodePosition` is convenient for access (it has named `start` and `end` fields), but tuples are convenient for construction.

However, it should still be possible to construct a `CodeRange` with `CodePosition` objects for parity between access and construction, so I think it makes sense to support both in `__init__`.

bgw:
Regarding Jimmy's comment about a usage notebook:

I think that'd be useful, but it's still nice to have a very small example on the same page as the reference itself.

I'd actually take the usage example (or an even more simplified example if that's possible) and move it to the top of the page, similar to what we do with the parser docs: https://libcst.readthedocs.io/en/latest/parser.html

I think notebooks should be reserved for more complex examples.

libcst/_nodes/_internal.py
rayjzeng: The primary use is for internal testing where it gets very tedious writing out `CodePosition` repeatedly. As Ben said, `CodePosition` is more descriptive when performing an access. Users shouldn't ever need to worry about constructing these objects anyhow as they're generated through the metadata provider.

jimmylai:
> Regarding Jimmy's comment about a usage notebook:
> 
> I think that'd be useful, but it's still nice to have a very small example on the same page as the reference itself.
> 
> I'd actually take the usage example (or an even more simplified example if that's possible) and move it to the top of the page, similar to what we do with the parser docs: https://libcst.readthedocs.io/en/latest/parser.html
> 
> I think notebooks should be reserved for more complex examples. E.g. it'd be great to have a metadata provider example in a notebook.

We can have the simple example in notebook and then include it from rst. That requires running a comment to generate the rst from notebook and then we can include it. In that case, we'll have executable code and simple example. The other way to always make sure the example code is executable is to use doctest.
https://github.com/spatialaudio/nbsphinx/issues/181#issuecomment-398080485

docs/source/metadata.rst
```diff
@@ -0,0 +1,60 @@
...
+Providing Metadata
+------------------
+
+Metadata is generated through provider classes that can be declared as a dependency
+by a subclass of :class:`~libcst.MetadataDependent`. These providers are then
+resolved automatically using methods provided by :class:`~libcst.MetadataWrapper`.
+
+.. autoclass:: libcst.BaseMetadataProvider
+.. autoclass:: libcst.VisitorMetadataProvider
+.. autoclass:: libcst.BatchableMetadataProvider
```
jimmylai: The difference between VisitorMetadataProvider and BatchableMetadataProvider (the concept of batchable) is not clear from the docstring. Even though it links to BatchableCSTVisitor, we can probably say something here to make it more clear it's for performance sake.

docs/source/metadata.rst
```diff
@@ -0,0 +1,60 @@
+.. _libcst-metadata:
```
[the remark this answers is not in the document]
rayjzeng: I did some reorganization of the docs and added to the tutorial. I only added an example for resolve() and not one for resolve_many() as I think it should be pretty easy to extrapolate how to use resolve_many() from that point.

PR #27, opened 2019-08-14
Implement attribute visitors

DragonMinded:
## Summary

It can be useful in a variety of situations to know what attribute of a node you are currently visiting children on. So, implement the ability to track this via attribute visit_ and leave_ support. This pull request finishes support to visit on attributes, adds a test to verify visitor behavior, and adds codegen for attribute visitor functions.

Note that this is on top of the existing pull request for codegen (https://github.com/Instagram/LibCST/pull/23).

## Test Plan

`tox`


libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,181 @@
...
+
+def _get_bases(node: Type[cst.CSTNode]) -> Generator[Type[cst.CSTNode], None, None]:
+    for base in classmap[node]:
+        if base.__name__ in DO_NOT_CARE_TYPES:
+            # Ignore bases that are Python implementation details.
+            continue
+        if base not in classmap:
+            # Ignore bases that aren't in our tracking list.
+            continue
+        if base.__name__.startswith("_Base"):
```
bgw: These underscore-prefixed members shouldn't be exported from the top-level libcst package, and therefore shouldn't be in `classmap` anyways.

libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,181 @@
...
+# pyre-strict
+import inspect
+from collections import defaultdict
+from dataclasses import fields
+from typing import Dict, Generator, List, Mapping, Sequence, Set, Type
+
+import libcst as cst
+
+
+DO_NOT_CARE_TYPES: Set[str] = {"CSTNode", "ABC", "object"}
```
bgw: Optional: You could check `issubclass(cls, CSTNode) and not isinstance(cls, CSTNode)` instead. It looks like you're already doing `issubclass(node, cst.CSTNode)` in `_get_nodes`.

libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def _get_base(node: Type[cst.CSTNode]) -> Type[cst.CSTNode]:
+    bases = set(_get_bases(node))
+    if not bases:
+        raise Exception(f"Couldn't get any bases for {node.__name__}!")
+    if len(bases) == 1:
+        for b in bases:
+            return b
+    bases = {b for b in bases if b is not node}
```
bgw: Maybe this should be `bases.remove(node)`

libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,163 @@
...
+            continue
+        yield base
+
+
+def _get_base(node: Type[cst.CSTNode]) -> Type[cst.CSTNode]:
+    bases = set(_get_bases(node))
+    if not bases:
+        raise Exception(f"Couldn't get any bases for {node.__name__}!")
+    if len(bases) == 1:
+        for b in bases:
```
bgw: You can just `return next(bases)` instead of using a for loop.

libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,163 @@
...
+
+    if len(new_bases) > 1:
+        raise Exception(f"Got too many bases for {node.__name__}!")
+    if len(new_bases) == 1:
+        for b in new_bases:
+            return b
+    raise Exception(f"Couldn't get any bases for {node.__name__}!")
+
+
+nodebases: Dict[Type[cst.CSTNode], Type[cst.CSTNode]] = {}
```
bgw: Between `nodes`, `classmap`, and `nodebases` it can be hard to figure out what variable does what. I think it would help if you gave these variables more detailed names.

libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,163 @@
...
+    # pyre-ignore __dir__ exists on all objects.
+    for name in cst.__dir__():
+        if name.startswith("__") and name.endswith("__"):
+            continue
+        if name == "CSTNode":
+            continue
+
+        node = getattr(cst, name)
+        try:
+            if issubclass(node, cst.CSTNode):
```
bgw: As-is, this seems to overlap a bit with your `DO_NOT_CARE_TYPES` check later. This is going to filter out `ABC` and `object`.

libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,163 @@
...
+        if base not in classmap:
+            # Ignore bases that aren't in our tracking list.
+            continue
+        if base.__name__.startswith("_Base"):
+            # Ignore bases that aren't exported.
+            continue
+        yield base
+
+
+def _get_base(node: Type[cst.CSTNode]) -> Type[cst.CSTNode]:
```
bgw: Maybe this should be `_get_most_generic_base`? As it is, it's hard to understand the difference between `_get_bases` and `_get_base`.

---

Separately: couldn't you just grab the last exported item in the MRO list before `CSTNode`? Unless I'm misunderstanding MRO (which is quite possible). That seems simpler than recursively traversing base classes.

I came up with this, which seemed to work on every node I tried it on:

```python
def _get_most_generic_base(cls):
    candidates = inspect.getmro(cls)
    
    # We only care about subclasses of CSTNode
    candidates = [base for base in candidates if issubclass(base, cst.CSTNode)]
    # But we don't want CSTNode itself
    candidates = [base for base in candidates if base is not cst.CSTNode]
    # And make sure it's actually an exported base
    candidates = [base for base in candidates if base.__name__ in cst.__all__]
    
    # return the most generic candidate
    return candidates[-1]
```

libcst/codegen/gen_visitor_functions.py
```diff
@@ -0,0 +1,163 @@
...
+generated_code.append("class CSTTypedTransformerFunctions(CSTTypedBaseFunctions):")
+generated_code.append("    pass")
+for node in sorted(nodebases.keys(), key=lambda node: node.__name__):
+    name = node.__name__
+    if name.startswith("Base"):
+        continue
+    generated_code.append("")
+    generated_code.append("    @mark_no_op")
+    generated_code.append(
+        f'    def leave_{name}(self, original_node: "{name}", updated_node: "{name}") -> Union["{nodebases[node].__name__}", RemovalSentinel]:'
```
bgw: While this is mostly correct, I noticed that there are a few cases where we accept a union of multiple potential values. For example, I might want to replace a compound statement with a simple statement, so the return type should actually be `Union["SimpleStatementLine", "BaseCompoundStatement"]`.

Or maybe I want to replace every comma with a `MaybeSentinel`.

I'm not sure that we should type the return value beyond `Union[CSTNode, MaybeSentinel, RemovalSentinel]`.

libcst/codegen/gen_visitor_functions.py
DragonMinded: They show up because we ask for all of the bases using inspect.getmro().

libcst/codegen/gen_visitor_functions.py
DragonMinded: I do that because I'm walking the tree until we get to nodes that are direct descendants of CSTNode, not to exclude nodes. I could issubclass() as well I guess.

libcst/codegen/gen_visitor_functions.py
DragonMinded: Nice, that will work better.

libcst/codegen/gen_visitor_functions.py
DragonMinded: Is that guaranteed by getmro? I'd rather not play code golf here and rewrite everything :\

libcst/codegen/gen_visitor_functions.py
DragonMinded: That's not actually correct though. If you replace a simple statement with a compound statement, you have no guarantee that the node isn't a child of a SimpleStatementLine.

libcst/_visitors.py
```diff
@@ -71,8 +72,32 @@ def on_leave(
 
         return updated_node
 
+    def on_visit_attribute(self, node: "CSTNode", attribute: str) -> None:
```
rayjzeng: The `on_visit_attribute` and `on_leave_attribute` code is shared between the visitor and transformer. Do we want to factor it out into helper functions to reuse code?

rayjzeng: Alternatively could they be moved into `CSTTypedVisitorFunctions`?

libcst/codegen/gen_visitor_functions.py
rayjzeng: Going all the way to allowing `CSTNode` would mean losing type safety in cases where you should be returning an expression but the user tries to return a statement. Looking briefly at the generated code and how inheritance is done with nodes it looks like its mostly an issue with statements as there's a lot of disjoint base classes for various kinds of statements.

libcst/tests/test_visitor.py
```diff
@@ -0,0 +1,99 @@
...
+                "visit_If",
+                "visit_If_test",
+                "visit_Name",
+                "leave_Name",
+                "leave_If_test",
+                "leave_If",
+            ],
+        )
+
+    def test_transofrmer(self) -> None:
```
rayjzeng: Typo in name `test_transformer`

libcst/codegen/gen_visitor_functions.py
bgw: Yes, but I think it's always valid to replace a `BaseCompoundStatement` with a `SimpleStatementLine`. I'd rather have the wider type because it's more functional even if it's not always as safe.

libcst/codegen/gen_visitor_functions.py
bgw: It works for everything I tested it on. I wrote this because I was having a really hard time understanding what the existing code did.

This whole thing is moot though if you make the return type of leave methods `Union[CSTNode, MaybeSentinel, RemovalSentinel]` like I suggest in my later comment.

libcst/codegen/gen_visitor_functions.py
bgw: Yes, but then you filter things out with `if base not in classmap`. No class beginning with `_Base` should be in `classmap`'s set of keys.

libcst/codegen/gen_visitor_functions.py
DragonMinded: I did some more thinking on this: Essentially, typing the response as CSTNode throws away the benefit of bothering to generate these functions in the first place. We're trying to generate functions that give a user the most constrained version available so that its less likely that invalid code is generated. Stating that all leave_* nodes can return this is not just wrong (not many node parents support a MaybeSentinel, and nobody needs to return a MaybeSentinel instead of a conrete type), but also useless from a typing perspective. If that's the case, we literally do not need to generate code since the whole purpose is for typing.

libcst/codegen/gen_visitor_functions.py
rayjzeng: I think if the existing code is fine if the function name is chanced to something like `_get_most_generic_base`. It's not too hard to follow the recursion once you realize what it's trying to do.

libcst/codegen/gen_visitor_functions.py
bgw: Even if the return type is `CSTNode`, it's still useful in ensuring that the arguments match the function name. Lint AST visitors never had return types, but we still codegen'ed them for this reason.

We could still check the return types using `validate_types_shallow` on the resulting parent node. In fact, we should probably default to running `validate_types_shallow` in `__post_init__` for people using LibCST without mypy or pyre.

jimmylai:
@DragonMinded, can you rebase on master so that codecov verifies the test coverage for you?

Issue #24, opened 2019-08-13
Make ParserSyntaxError more generic, use it in more places


bgw:
[`ParserSyntaxError`](https://github.com/Instagram/LibCST/blob/5d1514e416004f04f17ddf1e07fe90537046d65e/libcst/_exceptions.py#L16) is a little too coupled to how pgen2 works (my fault), which prevents us from using it everywhere we should. It's possible to get a parser error that isn't a `ParserSyntaxError`. That shouldn't be possible.

I propose that we:
- Remove the expected/encountered fields, and instead provide a helper function for creating an error message with those values. [Not every syntax error follows this format](https://github.com/Instagram/LibCST/blob/bf94bad977d53a82754ac04a1490a43b3beea00d/libcst/_parser/_conversions/expression.py#L1250).
- Initialize line and column numbers to a sentinel. We can make the exposed line/column properties raise an exception if this sentinel isn't initialized. While an exception may temporally lack position information during construction, it should always have a line/column number from the end user's perspective.
- Intercept exceptions caused by conversion functions in [`_base_parser`](https://github.com/Instagram/LibCST/blob/master/libcst/_parser/_base_parser.py), and attach line and column numbers before re-raising them.

PR #31, opened 2019-08-17
Make ParserSyntaxError more generic and use it in more places

bgw:
## Summary

`ParserSyntaxError` is a little too coupled to how pgen2 works (my fault), making some error messages bad and preventing us from using it everywhere we should.

A more detailed description of this problem is available in #24.

After refactoring `ParserSyntaxError`, this uses it in places where we should've been raising a `ParserSyntaxError`, but were raising an `Exception` instead.

## Test Plan

Unit tests!

### DIFF
--- a/libcst/_exceptions.py
+++ b/libcst/_exceptions.py
@@ -69,6 +69,25 @@ def _parser_syntax_error_unpickle(kwargs: Any) -> "ParserSyntaxError":
 
 
 @final
+class PartialParserSyntaxError(Exception):
+    """
+    An internal exception that represents a partially-constructed
+    :class:`ParserSyntaxError`. It's raised by our internal parser conversion functions,
+    which don't always know the current line and column information.
+
+    This partial object only contains a message, with the expectation that the line and
+    column information will be filled in by :class:`libcst._base_parser.BaseParser`.
+
+    This should never be visible to the end-user.
+    """
+
+    message: str
+
+    def __init__(self, message: str) -> None:
+        self.message = message
+
+
+@final
 class ParserSyntaxError(Exception):
     """
     Contains an error encountered while trying to parse a piece of source code. This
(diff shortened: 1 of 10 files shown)


libcst/tests/test_exceptions.py
```diff
@@ -4,54 +4,90 @@
...
+                ).strip(),
+            ),
+            "empty_file": (
+                cst.ParserSyntaxError(
+                    "some message", lines=[""], raw_line=1, raw_column=0
+                ),
+                dedent(
+                    """
+                    Syntax Error @ 1:1.
+                    some message
```
jimmylai: Why this case didn't show the position indicator `^`?

libcst/_exceptions.py
```diff
@@ -4,77 +4,225 @@
...
+                    "raw_column": self._raw_column,
+                },
+            ),
         )
 
     def __str__(self) -> str:
         """
-        A human-readable error message of where the syntax error is in their code.
+        A multi-line human-readable error message of where the syntax error is in their
+        code.
```
jimmylai: Can we add an example error message in this comment?

jimmylai:
This change adds many tests and I just setup test coverage in CI #35 to help us verify the coverage of new test cases. Let's merge #35 and then rebase this PR to verify it.

libcst/tests/test_exceptions.py
bgw: I added some logic that suppresses the context if it would just show an empty line, because I found an empty line followed by a caret indicator really confusing.

If you think this behavior is more confusing, I can remove it and unconditionally show the caret.

I'll add a comment to the test case to clarify the behavior a bit more.

libcst/_exceptions.py
bgw: :+1: Will do.

jimmylai:
@bgw, can you rebase on master to generate the test coverage report?

bgw:
Next step in the stack: converting conversion functions to ParserSyntaxError.

https://github.com/bgw/LibCST/pull/1

It doesn't show up as a PR to this repository because of the way github's forking model works, and I can't add people as reviewers...

Next, I'm going to work on some tweaks to the ParserSyntaxError class: I think I can clean up the partial-initialization logic by introducing a separate PartialParserSyntaxError class.

Issue #32, opened 2019-08-18
Can you ignore ParserSyntaxError (for reserved async keyword in 3.6)?


zhammer:
I'd like to use this on a project that is python3.6 syntax but not _fully_ 3.7 yet (as far as I know the only violation is that there are still some `async` keyword args which are reserved in 3.7).

Here's the exception I'm getting when visiting a module:
```
libcst._exceptions.ParserSyntaxError: Syntax Error: incomplete input @ 200:76.
Encountered 'async', but expected one of ["')'"].

        results = client.query_performers(query, "api_ingestion_consumer", async=False)
```

Would it be possible to suppress this error until the code is refactored to be totally 3.7 compliant? I realize "LibCST parses Python 3.7" is specifically noted in the README so understood if that's not possible.

zhammer:
Not sure if this is supported:
```py
source_tree = cst.parse_module(python_source, config=cst.PartialParserConfig(python_version="3.6"))
```
Tried running but still got the `async` error.

zhammer:
Ah just saw this: https://github.com/Instagram/LibCST/blob/68699e21d7fa24ee39ab9cea02c9118f4d50827b/libcst/_parser/_types/config.py#L85-L89.

bgw:
Hey, @zhammer. Thanks for your interest in LibCST!

Unfortunately we only support 3.7's syntax right now. Sorry for the confusing documentation around `python_version`. It's there because we intend to support multiple versions of Python syntax in the future. We should probably improve that and raise a `ValueError` if we see anything other than `3.7.x`.

This is actually a limitation in [Parso's tokenizer](https://github.com/davidhalter/parso/blob/master/parso/python/tokenize.py).

CPython 3.6's tokenizer contains a (really ugly) hack that tracks whether you're inside an async function or not, and decides which kind of token to emit for the `async` and `await` keywords.

Black's tokenizer (which is similar to Parso's) [contains a version of this logic](https://github.com/psf/black/blob/40e8b3a231bade22d838858435a33d60a8325306/blib2to3/pgen2/tokenize.py#L359). I think we'd need to port that logic from Black to Parso to fix this problem.

I've upstreamed this issue as https://github.com/davidhalter/parso/issues/83.

Out of curiosity: How many of these syntax errors do you have in your codebase? If it's small enough, it might be worth fixing them by hand, especially seeing as this is a requirement for upgrading to Python 3.7+ anyways.

zhammer:
[...]

PR #37, opened 2019-08-19
Update PartialParserConfig to more clearly reflect that we only support Python 3.7 (#32)

bgw:
## Summary

This PR has two parts:
1. We now raise an exception if the user tries to parse the code as anything other than 3.7.
2. The documentation has more details and should now properly reflect reality.

## Test Plan

Unit tests. Built and inspected the documentation.

https://1540-200896124-gh.circle-artifacts.com/0/doc/parser.html#libcst.PartialParserConfig
https://1540-200896124-gh.circle-artifacts.com/0/doc/nodes.html#module

### DIFF
--- a/libcst/_parser/_types/config.py
+++ b/libcst/_parser/_types/config.py
@@ -87,20 +87,26 @@ class PartialParserConfig:
         # We use object.__setattr__ because the dataclass is frozen. See:
         # https://docs.python.org/3/library/dataclasses.html#frozen-instances
         # This should be safe behavior inside of `__post_init__`.
-        object.__setattr__(
-            self,
-            "parsed_python_version",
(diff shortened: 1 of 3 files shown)


libcst/_parser/_types/config.py
```diff
@@ -63,21 +63,47 @@ def __repr__(self) -> str:
 @add_slots
 @dataclass(frozen=True)
 class PartialParserConfig:
-    """
+    r"""
     An optional object that can be supplied to the parser entrypoints (e.g.
     `parse_module`) to configure the parser.
```
jimmylai: Not related to this change but can you convert the `parse_module` as a link to the actual function?

PR #39, opened 2019-08-20
Fuzz with Hypothesmith

Zac-HD:
I wrote a fuzzer for you, because it seemed easier than reporting all the bugs 😁 

Happily, it passes for `libcst.parse_module`, though statements and expressions both fail - I'm using the `compile()` builtin as an oracle for what should be accepted, then testing that code always round-trips.  Apparently something about comments is different?

This fuzzer is based on an EBNF grammar plus some hand-tweaking plus filtering, but it's useful even at this early stage - I've now shaken out some bugs in black, lib2to3, libcst...  I also have some sketched-out prototypes that *use* libCST to generate code with semantic constraints, but sadly not quite enough free time to pursue them at the moment.


Zac-HD:
I'm not sure how to tell Pyre where to find Hypothesis and Hypothesmith...

But the test failures are "real bugs" in the sense that `libCST` is failing to parse something accepted by CPython (via the `compile()` builtin)!

bgw:
Wow, thanks @Zac-HD! Very impressive. This was more than I expected.

I'll look at this a bit more tomorrow, but a few quick notes:
- We probably don't care that `parse_expression` and `parse_statement` fail on certain inputs that `compile()` works on (maybe we could change the oracle). These functions disallow certain whitespace before/after themselves because there's nowhere to store it on the resulting node. We do care about fuzzing `parse_module` though.

  [This behavior is documented](https://libcst.readthedocs.io/en/latest/parser.html#libcst.parse_expression), but we could probably do a much better job of raising a descriptive error message (related to #24).

- I'm assuming that these tests are non-deterministic, and possibly slow. If that's the case, we need to make sure that they run on-demand (maybe via a separate tox command), but not in CI. We don't want to block PRs on a potentially slow or flaky test.

- You can add a placeholder stub ([example](https://github.com/Instagram/LibCST/blob/master/stubs/typing_inspect.pyi)) to make pyre shut up and treat the whole module as `Any`-typed. We can dig into why pyre isn't seeing the type annotations that already exist in your module later. I think there's some potential issues with our pyre search paths.

Zac-HD:
> We probably don't care that `parse_expression` and `parse_statement` fail on certain inputs that `compile()` works on (maybe we could change the oracle).

Easy enough: for these functions we can just test that *if* they parse an input, they can spit it out again.  I assume they should also reject anything rejected by `compile()`, so it's a tiny change.

> I'm assuming that these tests are non-deterministic, and possibly slow.

Around ten seconds total on my machine, to give you some idea.

The other question is a little trickier: while non-deterministic, Hypothesis tests are never *flaky* (unless the thing being tested is flaky, of course).  To be more specific, [there are several systems that ensure that once a test fails, it consistently fails until you fix the bug](https://hypothesis.readthedocs.io/en/latest/reproducing.html).  

On the other hand very rare edge cases *might* be triggered for the first time on an unrelated PR.  In my experience this is extremely rare - we average something like one per year on Hypothesis itself - and IMO better than not finding the bug at all.  This is exactly why I've provided the "fuzzing settings" though, so that you can run a thousand times the number of test cases, and be confident that if that passes your tests will almost certainly keep passing too.

DragonMinded:
> > We probably don't care that `parse_expression` and `parse_statement` fail on certain inputs that `compile()` works on (maybe we could change the oracle).
> 
> Easy enough: for these functions we can just test that _if_ they parse an input, they can spit it out again. I assume they should also reject anything rejected by `compile()`, so it's a tiny change.

This sounds ideal to me, and the assumption you put forth here sounds reasonable as well. We shouldn't accept input that Python also won't accept.

> > I'm assuming that these tests are non-deterministic, and possibly slow.
> 
> Around ten seconds total on my machine, to give you some idea.

This sounds reasonable to me. I would gladly take 10 seconds of slowdown over verifying round-trips across our internal repository which currently takes ~1.5 minutes. For local development, we can already filter on the test we're iterating against, so this seems completely fine.

> The other question is a little trickier: while non-deterministic, Hypothesis tests are never _flaky_ (unless the thing being tested is flaky, of course). To be more specific, [there are several systems that ensure that once a test fails, it consistently fails until you fix the bug](https://hypothesis.readthedocs.io/en/latest/reproducing.html).
> 
> On the other hand very rare edge cases _might_ be triggered for the first time on an unrelated PR. In my experience this is extremely rare - we average something like one per year on Hypothesis itself - and IMO better than not finding the bug at all. This is exactly why I've provided the "fuzzing settings" though, so that you can run a thousand times the number of test cases, and be confident that if that passes your tests will almost certainly keep passing too.

This also sounds reasonable to me. I think we are more worried about things that flip-flop on the order of per test run, not per PR.

libcst/tests/test_fuzz.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        self.reject_invalid_code(source_code, mode="eval")
+        try:
+            tree = libcst.parse_expression(source_code)
+        except Exception:
+            hypothesis.reject()
+        else:
+            self.assertEqual(source_code, libcst.Module([]).code_for_node(tree))
+
+    @unittest.skip("It's surprisingly tricky to round-trip statements")
```
DragonMinded: I'm curious, what additional stuff is lurking in statement causing us to not round trip properly? Is this something LibCST should be concerned with?

bgw: I think we've got a bug where trailing whitespace (on subsequent lines) is silently dropped instead of raising an exception. The fix will probably involve reading the whitespace state object once parsing is done to check that we're at the EOF.

DragonMinded: That doesn't seem to be too big of a deal. Since this is marked @skip already, we can merge and figure out what the deal is later.

DragonMinded: I can definitely see a few edge cases with parse_statement and parse_expression playing around with this more. You're exactly right about the trailing whitespace being silently dropped. I can tighten up these checks and add more documentation around parse_statement and parse_expression, then re-enable the statement check.

Zac-HD: `"pass\r#\n"` -> `"pass\n"` is representative, FYI.

As more general advice, the best way to fix up an input is by construction by the `.map(...)` method of the strategy, then by rejection sampling using `.filter(...)` on the strategy, and finally by rejection with `assume(...)` or `reject()`.  That's basically a question of efficiency, but it also affects how examples are displayed in verbose mode.

PR #43, opened 2019-08-22
Fix incorrect space validation with Expression operators.

DragonMinded:
## Summary

This was found by Hypothesis, so lets fix it! Turns out we aren't recursively
evaluating whether an expression can be used without spaces against a word
operator. That means that complex expressions such as `not...^A` fail to parse
when we really should allow such expressions.

## Test Plan

Ran `tox` with hypothesis tests failing, ensured that the failing test cases pass now.

### DIFF
--- a/libcst/_nodes/_expression.py
+++ b/libcst/_nodes/_expression.py
@@ -266,6 +266,18 @@ class BaseExpression(_BaseParenthesizedNode, ABC):
 
         return len(self.lpar) > 0 and len(self.rpar) > 0
 
+    def _check_left_right_word_concatenation_safety(
+        self,
+        position: ExpressionPosition,
+        left: "BaseExpression",
+        right: "BaseExpression",
+    ) -> bool:
+        if position == ExpressionPosition.RIGHT:
+            return left._safe_to_use_with_word_operator(ExpressionPosition.RIGHT)
+        if position == ExpressionPosition.LEFT:
+            return right._safe_to_use_with_word_operator(ExpressionPosition.LEFT)
+        return False
+
 
 class BaseAssignTargetExpression(BaseExpression, ABC):
     """
@@ -366,6 +378,9 @@ class Ellipsis(BaseExpression):
             lpar=visit_sequence("lpar", self.lpar, visitor),
             rpar=visit_sequence("rpar", self.rpar, visitor),
         )
+
+    def _safe_to_use_with_word_operator(self, position: ExpressionPosition) -> bool:
+        return True
 
     def _codegen_impl(self, state: CodegenState) -> None:
         with self._parenthesize(state):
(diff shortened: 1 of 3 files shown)


jimmylai:
Codecov complains the base of this diff didn't have test coverage report to compare the delta. @DragonMinded can you rebase on origin/master to make it happy?

jimmylai:
Overall LGTM. Can we reuse a helper to remove the repeated patterns?

This has been repeated 4 times.
```
if position == ExpressionPosition.LEFT:
   return self.something._safe_to_use_with_word_operator(position)
else:  # position == ExpressionPosition.RIGHT
  return self.some_other_thing._safe_to_use_with_word_operator(position)
```
https://github.com/Instagram/LibCST/blob/bf94bad977d53a82754ac04a1490a43b3beea00d/libcst/_nodes/_expression.py#L845-L848

Can we have a helper like 
```
def _check_left_right_word_concatenation_safety(position, left_part, right_part):
    if position == ExpressionPosition.LEFT:
        return left_part._safe_to_use_with_word_operator(position)
    else:  # position == ExpressionPosition.RIGHT:
        return right_part._safe_to_use_with_word_operator(position)
```

jimmylai (approved):

DragonMinded:
I was rebased on origin/master when I put this up. I don't see much value in playing with merge bases to see coverage info.

DragonMinded:
Hmmm, disabling hypothesis makes the coverage look pretty bad. I'm relying on fuzz testing to catch issues here. Not sure how valuable unit tests verifying extreme edge cases like this is, so I think the best thing we can do is work to uncover and fix issues found by Hypothesis and then get it set back up on the critical path.

Zac-HD:
```python
else:
    assert position == ExpressionPosition.RIGHT
    return right_part._safe_to_use_with_word_operator(position)
```

I'd tend to encode this kind of comment as an assertion - it's just as informative to humans, but also tells you when some future change breaks the invariant 😄 

> Not sure how valuable unit tests verifying extreme edge cases like this is, so I think the best thing we can do is work to uncover and fix issues found by Hypothesis and then get it set back up on the critical path.

I'd set it up to run a corpus of known-weird examples - even just checking that it parses would be helpful.

You could even do this by reusing the Hypothesis test methods - [decorate them with explicit examples](https://hypothesis.readthedocs.io/en/latest/reproducing.html#providing-explicit-examples), then have [`phases=(Phase.explicit, Phase.replay)`](https://hypothesis.readthedocs.io/en/latest/settings.html#phases) in the settings to skip the fuzzing stage for now.

DragonMinded:
Ooh, I will look into that, thanks so much! I have wanted to play with Hypothesis after seeing your talk at PyCon this year but you know how it goes. I'm super excited that this gives me the opportunity to do just that :3

PR #59, opened 2019-09-09
New metadata: ScopeProvider

jimmylai:
## Summary
Scope analysis is useful to empower more advanced static analysis and automatic refactoring applications. Some common use cases:
- automatically remove not accessed imports after code refactoring.
- rename a variable and update its all assignments/accesses.

To learn more about Python Scope, see https://www.geeksforgeeks.org/namespaces-and-scope-in-python/

The `ScopeProvider` provides scope object for every node in the tree. Scope object is a container for all variable assignments and accesses. Each module has one single `GlobalScope`. A new scope is created whenever we enter a `LocalScope` (`FunctionScope`/`ClassScope`/`ComprehensionScope`).
Scope object implements `__contains__(self, name:str)`, so we can look up a name by `variable_name in scope`.
Scope object also implements `__getitem__(self, name:str`, so we can get the assignments by `scope[varible_name]`
Each `Assignment` has accesses attribute stores all `Access`.
The variables available in `builtins` are handled differently as `BuiltinAssignmemt ` since they are available in GlobalScope by default and their definition are in builtins module. The other Assignment defined in the current module are of type `Assignment` which has a node (CSTNode) attribute.

`ScopeProvider` depends on ExpressionContextProvider but it's not implemented simply subclass BatchableMetadataProvider. Scope analysis is complex and it requires to create nested scope structures using context managers. A simple BatchableMetadataProvider only provides visit_<CSTNode> function which doesn't provide the needed capability. I ended up define a `ScopeVisitor(CSTVisitor)`. The visitor has a scope object and the metadata `provider` (passed by `ScopeProvider`). ScopeVisitor control the tree visit flow and sometimes manually visit node subtree with different context using the `node.visit(self)` pattern. A trick in `on_leave` sets self.scope to each `node` to make sure each node has a scope object associated.

The implementation wasn't complete and will iterate on it with other PRs. Detailed documentation will be added in a later PR.

## Test Plan
Many unit tests were added with high test coverage.


libcst/metadata/scope_provider.py
```diff
@@ -0,0 +1,309 @@
...
+        return self[name]
+
+    def _record_assignment_as_parent(self, name: str, node: cst.CSTNode) -> None:
+        """Overridden by ClassScope to forward 'nonlocal' assignments from child scopes."""
+        self.record_assignment(name, node)
+
+    def __contains__(self, name: str) -> bool:
+        return len(self[name]) > 0
+
+    def __getitem__(self, name: str) -> Tuple[BaseAssignment, ...]:
```
bgw: This should be an `abc.abstractmethod`.

libcst/metadata/scope_provider.py
```diff
@@ -0,0 +1,309 @@
...
+    def __init__(self, parent: "Scope") -> None:
+        super().__init__()
+        self.parent = parent
+        self.globals = parent.globals
+        self._assignments = defaultdict(list)
+
+    def record_assignment(self, name: str, node: cst.CSTNode) -> None:
+        self._assignments[name].append(Assignment(name=name, scope=self, node=node))
+
+    def _contains_in_parent_scope(self, name: str) -> bool:
```
bgw: Looks like nothing is calling `_contains_in_parent_scope` right now.

I think I implemented this in the AST version because it let me implement `__contains__` more efficiently (no extra tuple construction). If you plan to carry over that small optimization, then leave this here, otherwise remove it.

libcst/metadata/scope_provider.py
```diff
@@ -0,0 +1,309 @@
...
+)
+
+
+@dataclass(frozen=True)
+class Access:
+    node: cst.Name
+    scope: "Scope"
+
+
+class BaseAssignment:
```
bgw: I think this should inherit from `abc.ABC`.

libcst/metadata/scope_provider.py
```diff
@@ -0,0 +1,309 @@
...
+        self.globals = parent.globals
+        self._assignments = defaultdict(list)
+
+    def record_assignment(self, name: str, node: cst.CSTNode) -> None:
+        self._assignments[name].append(Assignment(name=name, scope=self, node=node))
+
+    def _contains_in_parent_scope(self, name: str) -> bool:
+        return name in self.parent
+
+    def _get_from_parent(self, name: str) -> Tuple[BaseAssignment, ...]:
```
bgw: This was called `_get_as_parent` in the AST version because it fetches `name` when you're the parent scope.

(Opinion) While `_get_as_parent` isn't a great name, `_get_from_parent` is even more confusing to me, because it doesn't (usually) get the item from the parent.

libcst/metadata/scope_provider.py
jimmylai: Nice catch. I prefer simple implementation to make the code easier to understand. E.g. I reuse `__getitem__` to implement `__contains__`. Even though it may be slower, we can optimize it later when we met the performance issue. Remove this function for now.

libcst/metadata/scope_provider.py
jimmylai: What about `_getitem_from_self_or_parent`?

libcst/metadata/scope_provider.py
```diff
@@ -0,0 +1,309 @@
...
+class Assignment(BaseAssignment):
+    node: cst.CSTNode
+
+    def __init__(self, name: str, scope: "Scope", node: cst.CSTNode) -> None:
+        self.node = node
+        super().__init__(name, scope)
+
+
+class BuiltinAssignmemt(BaseAssignment):
+    node: None = None
```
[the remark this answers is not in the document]
jimmylai: Since we don't actually have the CSTNode of the builtins' assignment. I intentionally made the node attribute not available in `BuiltinAssignmemt`. Will remove the `node=None`.

libcst/metadata/scope_provider.py
```diff
@@ -0,0 +1,356 @@
...
+    def visit_SetComp(self, node: cst.SetComp) -> Optional[bool]:
+        return self._visit_comp_alike(node)
+
+    def visit_DictComp(self, node: cst.DictComp) -> Optional[bool]:
+        return self._visit_comp_alike(node)
+
+    def visit_GeneratorExp(self, node: cst.GeneratorExp) -> Optional[bool]:
+        return self._visit_comp_alike(node)
+
+    def _visit_comp_alike(
```
bgw: It's clear what this code does, but it's not clear why it does it, and IMO, there's a lot of potential landmines in comprehension scopes.

Can you copy some of the comments in the AST implementation here?

jimmylai: @bgw 
Added docstring for more explanation.

In the original comment regarding nested comprehension, it mentioned the scope of outer iterator node (CompFor) will be the ComprehensionScope while the node.iter will be parent scope. It looked not straightforward to me. I thought it's more intuitive if the child node always extends the parent scope or create a new scope rather than extend grandfather scope. So I made the decision to set the CompFor node to parent scope.

libcst/metadata/scope_provider.py
```diff
@@ -0,0 +1,390 @@
...
+        return False
+
+    def infer_accesses(self) -> None:
+        for access in self.__deferred_accesses:
+            for assignment in access.scope[access.node.value]:
+                assignment.record_access(access)
+        self.__deferred_accesses = []
+
+    def on_leave(self, original_node: cst.CSTNode) -> None:
+        # will this always been called? What if the node wasn't visited? This should be tested.
```
DragonMinded: `on_leave` is guaranteed to be called for every node that `on_visit` is called on. If a node is not visited because it is a child of another node who returned `False` during a visit, its `on_leave` will also not be called.

jimmylai: Yeah, I just had the discussion with @bgw this week. Currently, some of the visit_CSTNode stop the visit by return False and only visited major attributes. Some of the trivial attributes (e.g. lpar, rpar, etc) wasn't visited and thus won't have Scope object associated with in this case. We don't think it'll cause issue in application since we should only need the scope object given a meanful node (Name, FunctionDef, etc).
The ease the concern of this comment, I can just exhaust the visit for all attributes.

PR #60, opened 2019-09-10
Fork Parso Internally

DragonMinded:
## Summary

We need to be able to customize the tokenizer to support 3.6 and below. We've also started wrapping several Parso structures with our own versions, adding indrection. I don't think its fair to try to upstream these changes as they're strictly for our use and would cause additional maintenance burden on Parso without any benefit to them. So, bring the files that we're using into this repo so we can modify them further.

## Test Plan

tox, pyre


libcst/_parser/_parso/_pgen2/_generator.py
```diff
@@ -0,0 +1,361 @@
+# A fork of `parso.pgen2.generator`.
```
bgw: Regarding this file name:

I originally only intended for one underscore in each module's fully qualified name as a way of marking certain modules as private.

When we ended up reworking some of the internal package structure, we ended up with multiple underscores in package names.

I get that this is a matter of opinion, but based on my original idea for underscores, that would mean that this module should be `libcst._parser.parso.pgen2.generator`, not `libcst._parser._parso._pgen2._generator`.

It's up to you what pattern you want to follow.

DragonMinded: Hmm, I wasn't really sure what our methodology was here, since it wasn't consistent to start with. That seems like a relatively easy change to make, I'll put it on the  backlog to fix all of the relevant areas both within parso and elsewhere.

jimmylai:
Found some copyright header inconsistency. For files copied from parso, we should always add the copyright `Copyright 2004-2005 Elemental Security, Inc. All Rights Reserved.`

This files explain the copyright in the header in a clear way.
https://github.com/Instagram/LibCST/blob/aac5103592de649cc1245d2ccb32825df1ceea9b/libcst/_parser/_base_parser.py
Also, we need to exclude copyright check for those files.

For file added by you, you'll need to include the `# Copyright (c) Facebook, Inc. and its affiliates.` header.

libcst/_parser/_parso/_python/_token.py
```diff
@@ -0,0 +1,38 @@
+# A fork of `parso.python.token`.
```
jimmylai: copyright missing.

libcst/_parser/_parso/_utils.py
```diff
@@ -0,0 +1,210 @@
+# A fork of `parso.utils`.
```
jimmylai: copyright missing.

libcst/_parser/_parso/tests/test_fstring.py
```diff
@@ -0,0 +1,39 @@
+# A fork of Parso's tokenize test
```
jimmylai: copyright missing

libcst/_parser/_parso/tests/test_utils.py
```diff
@@ -0,0 +1,47 @@
+# A fork of Parso's tokenize test
```
jimmylai: copy missing.

libcst/_parser/_parso/tests/test_tokenize.py
```diff
@@ -0,0 +1,420 @@
+# A fork of Parso's tokenize test
```
jimmylai: copyright missing.

libcst/_parser/_parso/_pgen2/_grammar_parser.py
```diff
@@ -0,0 +1,166 @@
+# A fork of `parso.pgen2.grammar_parser`.
+# https://github.com/davidhalter/parso/blob/master/parso/pgen2/grammar_parser.py
+#
+# The following changes were made:
+# - Type stubs were directly applied.
+
+# Copyright 2004-2005 Elemental Security, Inc. All Rights Reserved.
```
jimmylai: this is the copyright to be copied to other files.

PR #65, opened 2019-09-15
Fix some typos and add .watchmanconfig file for pyre incremental check

jimmylai:
## Summary
as title.


DragonMinded:
These changes should be separated at least into different commits, but its not a big enough deal to request changes.

PR #66, opened 2019-09-16
[doc] add ScopeProvider document

jimmylai:
## Summary
Add documentation for `ScopeProvide` and explain all related data structure including `Assignment`, `Access` and different type of Scope.
Also moved all metadata provider and its custom data structures to libcst.metadata namespace.

Ran into weird issue of Sphinx autodoc. Some class autodoc cannot be rendered (e.g. Access, GlobalScope).
Didn't see any errors/warnings and don't plan to spent too much time to debug autodoc. 
https://2162-200896124-gh.circle-artifacts.com/0/doc/metadata.html#scope-metadata
After play around with different docstring, found that adding some whitespaces in the beginning of those docstring made them show up. Though it made the formatting odd but that's a work around with reasonable effort I found.
https://2166-200896124-gh.circle-artifacts.com/0/doc/metadata.html#scope-metadata
[update] the bug was really weird. After a couple more updates, now only `ScopeProvider` having this issue. (Access and GlobalScope work fine now)


docs/source/conf.py
```diff
@@ -141,19 +138,21 @@
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
-    (master_doc, 'LibCST.tex', 'LibCST Documentation',
-     'Benjamin Woodruff, Jennifer Taylor', 'manual'),
+    (
+        master_doc,
+        "LibCST.tex",
+        "LibCST Documentation",
+        "Benjamin Woodruff, Jennifer Taylor",
```
DragonMinded: This doesn't match above author line. Should we fix that?

libcst/__init__.py
```diff
@@ -200,6 +200,18 @@
     BasicPositionProvider,
     SyntacticPositionProvider,
 )
+from libcst.metadata.scope_provider import (
+    Access,
+    Assignment,
```
DragonMinded: I'm worried about this being confused with Assign/AnnAssign/AugAssign nodes already present. Same with BaseAssignment below. Should we standardize instead on having people import from `libcst.metadata`? When matchers are added, they will not be included in `libcst/__init__.py`.

libcst/metadata/scope_provider.py
```diff
@@ -41,11 +56,16 @@ def record_access(self, access: Access) -> None:
...
     def accesses(self) -> Tuple[Access, ...]:
+        """Return all accesses of the assignment."""
         # we don't want to publicly expose the mutable version of this
         return tuple(self.__accesses)
 
 
 class Assignment(BaseAssignment):
+    """An assignment records the name, CSTNode and its accesses."""
+
+    #: The node of assignment, it could be a :class:`~libcst.Import`, :class:`ImportFrom`,
```
DragonMinded: Maybe we should link to Python language specification for variable assignment? This is not immediately clear as to why an Assignment handles Import/ImportFrom, FunctionDef and ClassDef.

jimmylai: The Assignment naming here is different than the regular Assignment in Python. Maybe we should have a different naming to make it more clear. E.g. "Definition"

libcst/__init__.py
jimmylai: Moved all the metadata providers and its data structure to namespace libcst.metadata to prevent confusion.

libcst/metadata/scope_provider.py
jimmylai: After discussed with @bgw, Assignment is a better naming here than other candidates. In Python, the concept of declaration and assignment are mixed together. A name could be assigned multiple times and Declaration usually mean the only one declaration. Definition can refer to FunctionDef or ClassDef but it's not directly linked to assignments.
So we decided to keep the name for now.

docs/source/metadata.rst
```diff
@@ -77,16 +81,66 @@ Accessing position metadata through the :class:`~libcst.MetadataDepedent`
...
    :no-undoc-members:
 
-.. autoclass:: libcst.ExpressionContext
\ No newline at end of file
+.. autoclass:: libcst.metadata.ExpressionContext
+
+Scope Metadata
+--------------
+Scope is the block of naming binding. The bind name is not available
+after existing the bind block. Pyhon is
```
bgw: s/Pyhon/Python/

docs/source/metadata.rst
```diff
@@ -77,16 +81,66 @@ Accessing position metadata through the :class:`~libcst.MetadataDepedent`
...
+            for i in range(ITERATIONS):
+                ...
+            return [
+                i for i in range(10)
+            ]
+    Cls().fn()
+
+The scopes of each name assignments and class/function definitions are visualized:
+
+.. image:: _static/img/python_scopes.svg
```
bgw: - Use the png instead of the svg. The svg won't render properly if the user doesn't have the font we need. SVGs can use webfonts, which would solve this, but they're a pain to configure in inkscape, and a PNG is easier.

- Since this image contains the code above, I think including both the text and the graphic below it is redundant. We should just include the image.

docs/source/metadata.rst
```diff
@@ -66,6 +66,10 @@ your provider does not use the visitor pattern for computing metadata for a tree
 ------------------
 Metadata Providers
 ------------------
+:class:`~libcst.metadata.BasicPositionProvider`, :class:`~libcst.metadata.SyntacticPositionProvider`,
+:class:`~libcst.metadata.ExpressionContextProvider` and :class:`~libcst.metadata.ScopeProvider`
+are currently provided. Each metadata provider may has its own custom data structure. Read the following
```
bgw: "Read the following document for more detail." seems like an extraneous sentence because the user is already reading the document.

jimmylai:
> Changes LGTM, but I'll leave it up to @DragonMinded to make the final review.
> 
> If we're going to move metadata to a separate namespace, we should probably figure out a deprecation process for removing metadata providers from the top-level namespace.
> 
> I'm still pretty mystified about the Sphinx issues you ran into...

@DragonMinded already reviewed and approved. All the feedback was addressed. Merging this now. If there were more need to be changed, I can create another PR.

PR #67, opened 2019-09-17
Fix internal underscore convention

DragonMinded:
## Summary

Standardize on the convention that private modules (those we don't expect people to directly import) are prefixed with an underscore. Everything under a directory/module that has an underscore is considered private, unless it is re-exported from a non-underscored module. Most things are exported from libcst directly, but there are a few things in libcst.tool, libcst.codegen and libcst.metadata that are namedspaced as such.

## Test Plan

tox


DragonMinded:
This unifies the approach in _parser versus _nodes, and makes tab completing files in private modules a lot easier (not to mention helping with RSI given the _ placement).

PR #71, opened 2019-09-23
Parent node metadata

jimmylai:
## Summary
Sometimes user may want to access the parent `CSTNode`
for more information or traverse in bottom-up manner.
We provide `ParentNodeProvider` for those use cases.
## Test Plan
Unit tests.


libcst/test_scope.py
```diff
@@ -0,0 +1,21 @@
...
+            enumerate(var)
+
+    def g(self):
+        C.var = 3
+
+
+var = 4
+
+x = C()
+x.var = 5
```
DragonMinded: What is this file?

libcst/metadata/tests/test_parent_node_provider.py
```diff
@@ -0,0 +1,66 @@
...
+                {elt for target in iterator if condition}
+                {elt: target for target in iterator if condition}
+                (elt for target in iterator if condition)
+                """,
+            ),
+        )
+    )
+    def test_parent_node_provier(self, code: str) -> None:
+        wrapper = MetadataWrapper(cst.parse_module(dedent(code)))
+        wrapper.visit(DependentVisitor(test=self))
```
DragonMinded: Nice trick ;)

DragonMinded:
This seems to combine some unrelated changes. I'm not sure what test_scope.py is but it looks like test code left in on accident. Also, what are the METADATA files and how do they relate? Back to you for comments.

libcst/test_scope.py
jimmylai: This wasn't needed. Just removed it.

pip-wheel-metadata/libcst.dist-info/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
DragonMinded: I'm still confused on this file?

pip-wheel-metadata/libcst.dist-info/METADATA
```diff
@@ -0,0 +1,217 @@
+Metadata-Version: 2.1
```
DragonMinded: Same comment here.

pip-wheel-metadata/libcst.dist-info/LICENSE
jimmylai: I'm sorry, accidentally committed those. Just removed them.

PR #73, opened 2019-09-23
Verify Tree Construction Uniqueness

### LINKED ISSUE #72: LibCST didn't generate a `tree`, it's a graph which some node can be children of multiple parent nodes
Surprisingly, LibCST didn't generate a `tree`, it's a graph which some nodes can be the children of multiple parent nodes.

When I worked on ParentNodeProvider https://github.com/Instagram/LibCST/pull/71, found this issue.

Here is a simple example test case to reproduce the issue. https://github.com/Instagram/LibCST/blob/f4b317501cc749307e5bb560563cde770282f959/libcst/metadata/tests/test_parent_node_provider.py#L49-L77
By using a simple visitor to traverse the node and count the number of visited times by id. One of the node was reused three times as the child node of three different parent nodes.
```
SimpleWhitespace(
    value='',
)
``` 
Test result: https://circleci.com/gh/Instagram/LibCST/2283?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Example source code to be parsed:
```
        foo = 'toplevel'
        fn1(foo)
        fn2(foo)
        def fn_def():
            foo = 'shadow'
            fn3(foo)
```

Due to this issue, I wasn't able to implement the desired ParentNodeProvider because a child not can have multiple nodes as parents but we only expect one. We need this to be fixed to unblock ParentNodeProvider. CC @DragonMinded  @bgw

DragonMinded:
## Summary

Add test case for uniqueness of resulting tree when parsing. This should be guaranteed by LibCST, and again guaranteed if using metadata by doing a deep copy. The reason for this second copy is that it is completely valid to ask for metadata on a tree that was constructed via a previous traversal, and we don't want to rely on visitors to guarantee uniqueness itself.

Also add a test case for deep_clone, since this was discovered broken. Fix deep_clone in a subsequent commit, as verified by the newly added test.

This fixes https://github.com/Instagram/LibCST/issues/72

## Test Plan

tox

### DIFF
--- a/libcst/_nodes/base.py
+++ b/libcst/_nodes/base.py
@@ -4,10 +4,12 @@
 # LICENSE file in the root directory of this source tree.
 
 from abc import ABC, abstractmethod
+from copy import deepcopy
 from dataclasses import dataclass, field, fields, replace
 from typing import (
     TYPE_CHECKING,
     Any,
+    Dict,
     List,
     MutableMapping,
     Sequence,
@@ -305,7 +307,23 @@ class CSTNode(ABC):
         >>> tree.deep_equals(tree.deep_clone())
         True
         """
-        return cast(_CSTNodeSelfT, self.visit(_NOOPVisitor()))
+        cloned_fields: Dict[str, object] = {}
+        for field in fields(self):
+            key = field.name
+            if key[0] == "_":
+                continue
+            val = getattr(self, key)
+
+            # We can't use isinstance(val, CSTNode) here due to poor performance
+            # of isinstance checks against ABC direct subclasses. What we're trying
+            # to do here is recursively call this functionality on subclasses, but
+            # if the attribute isn't a CSTNode, fall back to copy.deepcopy.
+            try:
+                cloned_fields[key] = val.deep_clone()
+            except AttributeError:
+                cloned_fields[key] = deepcopy(val)
+
+        return type(self)(**cloned_fields)
 
     def deep_equals(self, other: "CSTNode") -> bool:
         """
--- a/libcst/_nodes/base.py
+++ b/libcst/_nodes/base.py
@@ -4,7 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from abc import ABC, abstractmethod
-from dataclasses import dataclass, fields, replace
+from dataclasses import dataclass, field, fields, replace
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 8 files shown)


libcst/_nodes/base.py
```diff
@@ -347,6 +365,15 @@ def __repr__(self) -> str:
         lines.append(")")
         return "\n".join(lines)
 
+    @classmethod
+    def field(cls, *args: object, **kwargs: object) -> Any:
+        """
+        A helper that allows us to easily use SimpleWhitespace in dataclass constructor
```
jimmylai: This is not only for SimpleWhitespace, other CSTNode also uses it (e.g. TrailingWhitespace).
Can you make the comment more generic?

libcst/_nodes/expression.py
```diff
@@ -2696,9 +2702,9 @@ class Tuple(BaseAssignTargetExpression, BaseDelTargetExpression):
     #: in the tuple.
     elements: Sequence[BaseElement]
 
-    lpar: Sequence[LeftParen] = (LeftParen(),)
+    lpar: Sequence[LeftParen] = field(default_factory=lambda: (LeftParen(),))
```
jimmylai: Why not `lpar  = LeftParen.field()`?

libcst/_parser/tests/test_node_identity.py
```diff
@@ -0,0 +1,44 @@
...
+    def test_tokenize(self, code: str) -> None:
+        class CountVisitor(cst.CSTVisitor):
+            def __init__(self) -> None:
+                self.count = Counter()
+                self.nodes = {}
+
+            def on_visit(self, node: cst.CSTNode) -> bool:
+                self.count[id(node)] += 1
+                if self.count[id(node)] > 1:
+                    raise Exception(
```
jimmylai: Since this is a test, can we use the standard test assert: `assertEqual(self.count[id(node)], 1, f"Node duplication detected between {node} and {self.nodes[id(node)]}")`?

libcst/tests/test_deep_clone.py
```diff
@@ -0,0 +1,51 @@
...
+                self.nodes.add(id(node))
+                return True
+
+        class NodeVerifyVisitor(cst.CSTVisitor):
+            def __init__(self, nodes: Set[int]) -> None:
+                self.nodes = nodes
+
+            def on_visit(self, node: cst.CSTNode) -> bool:
+                if id(node) in self.nodes:
+                    raise Exception(f"Node {node} was not cloned properly!")
```
jimmylai: Same comment as above (assertEqual).

jimmylai (approved):
LGTM other than the comments.

libcst/_nodes/base.py
DragonMinded: O yeah I missed that when I moved this originally. Good catch!

libcst/_nodes/expression.py
DragonMinded: Yeah, this is kinda ugly, but its a special case for I think just this node? It wasn't used often enough to try to make it better.

libcst/_parser/tests/test_node_identity.py
DragonMinded: Good idea.

PR #74, opened 2019-09-24
Bump pyre version to 0.0.30.

DragonMinded:
## Summary

This introduces a few new gotchas (namely attribute access and a bug with
Union[Callable]), but it also removes a whole host of pyre-fixmes and gets us
updated to a release of pyre that came out after May.

## Test Plan

Wait for circle CI. Ran Pyre locally.

### DIFF
--- a/libcst/tests/test_type_enforce.py
+++ b/libcst/tests/test_type_enforce.py
@@ -81,7 +81,7 @@ class TypeEnforcementTest(UnitTest):
             (123, Union[Optional[str], Optional[int]]),
             # Iterables are supported and must match the type covariantly
             ([123], List[int]),
-            ([123], Iterable[int]),
+            ([123], Iterable[int]),  # pyre-ignore This is a type specification
             ([123], Iterable),
             ((123,), Iterable[int]),
             ([123], Sequence[int]),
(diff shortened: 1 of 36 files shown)


libcst/tests/test_type_enforce.py
```diff
@@ -182,7 +182,7 @@ def test_basic_pass(self, value, expected_type) -> None:
             #   List[ -> means its invariant
             #     MyExampleClass -> has a subclass
             # does not allow List[List[MyExampleChildClass]]
-            ([[MyExampleChildClass()]], Iterable[List[MyExampleClass]]),
+            ([[MyExampleChildClass()]], Iterable[List[MyExampleClass]]),  # pyre-ignore
```
jimmylai: Can you also add more detail about what's been ignored here? (Just like what you did for the other cases).

jimmylai:
For some reason, this change failed the docs build job and the error was missing a package which the cause wasn't intuitive to me.
https://circleci.com/gh/Instagram/LibCST/2328?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

libcst/tests/test_type_enforce.py
DragonMinded: Its the same as the above errors :\

PR #75, opened 2019-09-25
Initial export of matchers into LibCST.

DragonMinded:
## Summary

Matchers are now at a point where their API is stable and I'm satisfied with it enough to open-source. This brings 100% of current matcher feature parity into LibCST. I wanted to originally make a separate optional module for this in order for downstream dependencies to not pay an extra download cost. However, its not possible in Python's module model to have a module "libcst" and another module "libcst.matchers". Adding to a completely separate namespace would be hard to discover, so I chose to instead include it in the core. Note that module dependencies are only outward from "libcst.matchers" to "libcst", not the other way around. This is by design and should be kept that way, so that people who want to use LibCST core without matchers do not need to incur extra slowdown or indirection.

Given that an `__init__.py` which was only imports would need to be generated along with any other generated match class file, I chose to generate directly into the `__init__.py` file. This means that "libcst.matchers" is the namespace for all imports. I've generated an `__all__` inside "libcst.matchers" as well so that things work as expected. This is not ideal, but I feel its better than having two separate generated files.

Note that this tickles a fairly serious regression in Pyre, so type checking is slowed down greatly. I expect this to be partially mitigated by an upcoming Pyre open-source release which is why I took the liberty of upgrading us to Pyre first. Hopefully when this releases I will be able to do another upgrade to the newest version easier and fix the regression.

Note also that this does not include documentation for matchers. I wanted to keep this import focused, and will work on documentation and promotion to a supported feature in a subsequent PR. For now, I'm focused on getting this open-sourced in order to unblock downstream users of LibCST.

## Test Plan

This was almost entirely a copy-paste of matchers from inside instagram, with just a bit of changes for typing and import locations. There have been no functional changes, and all tests and code has been preserved as identically as possible while still meeting open-source tox and pyre strictness.

pyre, tox


libcst/codegen/gen_matcher_classes.py
```diff
@@ -0,0 +1,528 @@
...
+
+
+def _get_fields(node: Type[cst.CSTNode]) -> Generator[Field, None, None]:
+    """
+    Given a CSTNode, generate a field name and type string for each.
+    """
+
+    for field in fields(node) or []:
+        if field.name == "_metadata":
+            continue
```
jimmylai: Based on the test report, this never heppend https://codecov.io/gh/Instagram/LibCST/pull/75/diff?src=pr&el=tree#D6-577

libcst/matchers/_matcher_base.py
```diff
@@ -0,0 +1,638 @@
...
+        # Without a cast, pyre thinks that self.func is not a function, even though
+        # it recognizes that it is a _CallableT bound to Callable.
+        self.func: Callable[..., bool] = cast(Callable[..., bool], func)
+
+    def __or__(
+        self, other: _OtherNodeT
+    ) -> "OneOf[Union[MatchIfTrue[_CallableT], _OtherNodeT]]":
+        # Without a cast, pyre thinks that the below OneOf is type OneOf[object]
+        # even though it has the types passed into it.
+        return cast(
```
jimmylai: This is not tested. Can we add a test case for it?
https://codecov.io/gh/Instagram/LibCST/pull/75/diff?src=pr&el=tree#D6-210

libcst/matchers/_matcher_base.py
```diff
@@ -0,0 +1,638 @@
...
+        # Return the opposite evaluation
+        return not _attribute_matches(node, matcher.matcher)
+
+    if isinstance(matcher, MatchIfTrue):
+        # We should only return if the matcher function is true.
+        return matcher.func(node)
+
+    if matcher is None:
+        # Should exactly be None
+        return node is None
```
jimmylai: This looks like a base case worth a test case.
https://codecov.io/gh/Instagram/LibCST/pull/75/diff?src=pr&el=tree#D6-483

libcst/matchers/_matcher_base.py
```diff
@@ -0,0 +1,638 @@
...
+
+    if isinstance(matcher, str):
+        # Should exactly match matcher text
+        return node == matcher
+
+    if isinstance(matcher, bool):
+        # Should exactly match matcher bool
+        return node is matcher
```
jimmylai: This looks like a base case worth a test case.
https://codecov.io/gh/Instagram/LibCST/pull/75/diff?src=pr&el=tree#D6-491

libcst/matchers/_matcher_base.py
```diff
@@ -0,0 +1,638 @@
...
+        )
+
+        if isinstance(matcher, OneOf):
+            # We should compare against each of the sequences in the OneOf
+            for m in matcher.options:
+                if isinstance(m, collections.abc.Sequence):
+                    # Should match the sequence of requested nodes
+                    if _sequence_matches(node, m):
+                        return True
+                elif isinstance(m, MatchIfTrue):
```
jimmylai: Can we test this combination (OneOf + MatchIfTrue)?
https://codecov.io/gh/Instagram/LibCST/pull/75/diff?src=pr&el=tree#D6-507

libcst/matchers/_matcher_base.py
```diff
@@ -0,0 +1,638 @@
...
+                    return matcher.func(node)
+        elif isinstance(matcher, AllOf):
+            # We should compare against each of the sequences in the AllOf
+            for m in matcher.options:
+                if isinstance(m, collections.abc.Sequence):
+                    # Should match the sequence of requested nodes
+                    if not _sequence_matches(node, m):
+                        return False
+                elif isinstance(m, MatchIfTrue):
```
jimmylai: Can we test this combination (AllOf + MatchIfTrue)?
https://codecov.io/gh/Instagram/LibCST/pull/75/diff?src=pr&el=tree#D6-516

libcst/matchers/_matcher_base.py
```diff
@@ -0,0 +1,638 @@
...
+    matcher: Union[
+        BaseMatcherNode,
+        MatchIfTrue[Callable[..., bool]],
+        InverseOf[Union[BaseMatcherNode, MatchIfTrue[Callable[..., bool]]]],
+    ],
+) -> bool:
+    if isinstance(node, MaybeSentinel):
+        # We can't possibly match on a maybe sentinel, so it only matches if
+        # the matcher we have is a InverseOf.
+        return isinstance(matcher, InverseOf)
```
jimmylai: This looks like a base case worth a test case.
https://codecov.io/gh/Instagram/LibCST/pull/75/diff?src=pr&el=tree#D6-601

libcst/codegen/gen_matcher_classes.py
DragonMinded: Its dependent on Python version from my testing.

DragonMinded:
I'm merely exporting this from our internal repository where it is already reviewed. Adding tests where there were none before is outside the scope of this PR.

PR #79, opened 2019-09-26
add helper Scope.get_qualified_names_for

jimmylai:
## Summary
get_qualified_names_for return a list of fully qualified name (QualifiedName) given a CSTNode.
QualifiedName is a dataclass with name and source (QualifiedNameSource) attributes.
It finds the full name of the given node and then use the first part of name to look up assignment in the current scope.
The fully qualified names for different type of assignments:
- Import, ImportFrom: `a.b.c` if c is import from a.b, source=IMPORT
- Assignment: `something`, source=LOCAL
- BuiltinAssignment: `builtins.some_thing`, source=BUILTIN

Use a simple dataclass with field `name` and `source` seems straightforward and it can also make the project root universal qualified name implementation easier. The lint/codemod framework can use the output of FullyQualifiedName metadata provider to generate the universal name in those steps: If the `source` is LOCAL, just append directories from root dir and module name as prefix to the name. If the `source` is IMPORT, then we can use the QualifiedName from the imported module to recursively resolve the universal name if it's indirect import.

## Test Plan
Unit tests.


libcst/metadata/scope_provider.py
```diff
@@ -86,6 +99,84 @@ class BuiltinAssignment(BaseAssignment):
...
+                name_prefixes.append(scope.name)
+            elif isinstance(scope, GlobalScope):
+                break
+            else:
+                raise Exception(f"Unexpected Scope: {scope}")
+            scope = scope.parent
+
+        results.add(
+            QualifiedName(
+                ".".join([*name_prefixes[::-1], *name_parts]), QualifiedNameSource.LOCAL
```
bgw: Minor nit: I find `reversed(name_prefixes)` much easier to read than `name_prefixes[::-1]`.

libcst/metadata/scope_provider.py
```diff
@@ -86,6 +99,84 @@ class BuiltinAssignment(BaseAssignment):
...
+            return (
+                f"{_QualifiedNameUtil.get_full_name_for(node.value)}.{node.attr.value}"
+            )
+        elif isinstance(node, cst.Call):
+            return _QualifiedNameUtil.get_full_name_for(node.func)
+        elif isinstance(node, cst.Subscript):
+            return _QualifiedNameUtil.get_full_name_for(node.value)
+        elif isinstance(node, cst.SimpleString):
+            # In the case of SimpleString of type hints.
+            return literal_eval(node.value)
```
DragonMinded: I am wondering if this should be explicitly disallowed. It might be surprising that people accidentally pass a string instead of a variable the string is assigned to, and find out that it crashes during literal_eval. I guess it is fine as long as we are documenting it though. Thoughts?

libcst/metadata/scope_provider.py
```diff
@@ -86,6 +99,84 @@ class BuiltinAssignment(BaseAssignment):
...
+    def find_qualified_name_for_import_alike(
+        assignment_node: Union[cst.Import, cst.ImportFrom],
+        name_parts: List[str],
+        results: Set[QualifiedName],
+    ) -> None:
+        module = ""
+        if isinstance(assignment_node, cst.ImportFrom):
+            module_attr = assignment_node.module
+            if module_attr:
+                module = _QualifiedNameUtil.get_full_name_for(module_attr)
```
DragonMinded: If a module is relative-only import, it looks like we currently handle this as setting the module to "". I think that's fine, but we should probably have a TODO.

libcst/metadata/scope_provider.py
```diff
@@ -86,6 +99,84 @@ class BuiltinAssignment(BaseAssignment):
...
+        elif isinstance(node, cst.SimpleString):
+            # In the case of SimpleString of type hints.
+            return literal_eval(node.value)
+        return None
+
+    @staticmethod
+    def find_qualified_name_for_import_alike(
+        assignment_node: Union[cst.Import, cst.ImportFrom],
+        name_parts: List[str],
+        results: Set[QualifiedName],
```
DragonMinded: Its a bit weird that this takes in a set that is mutated instead of returning a new set.

libcst/metadata/scope_provider.py
```diff
@@ -86,6 +99,84 @@ class BuiltinAssignment(BaseAssignment):
...
+                        results.add(
+                            QualifiedName(
+                                ".".join([real_name, *name_parts[1:]]),
+                                QualifiedNameSource.IMPORT,
+                            )
+                        )
+
+    @staticmethod
+    def find_qualified_name_for_non_import(
+        assignment: Assignment, name_parts: List[str], results: Set[QualifiedName]
```
DragonMinded: Same comment here..

libcst/metadata/scope_provider.py
```diff
@@ -86,6 +99,84 @@ class BuiltinAssignment(BaseAssignment):
...
+    ) -> None:
+        scope = assignment.scope
+        name_prefixes = []
+        while scope:
+            if isinstance(scope, ClassScope) or isinstance(scope, FunctionScope):
```
bgw: What happens if we have a variable that's defined in the function?

```
def fn():
    class Cls: ...
```

My current understanding is that Cls's qualified name would be `fn.Cls`, which I don't think is correct.

`Cls` should either not have as qualified name, or we should match the behavior of `__qualname__` and generate `fn.<locals>.Cls`.

libcst/metadata/scope_provider.py
```diff
@@ -136,6 +227,34 @@ def record_global_overwrite(self, name: str) -> None:
...
+                        assignment_node, parts, results
+                    )
+                else:
+                    _QualifiedNameUtil.find_qualified_name_for_non_import(
+                        assignment, parts, results
+                    )
+            elif isinstance(assignment, BuiltinAssignment):
+                results.add(
+                    QualifiedName(
+                        f"builtins.{assignment.name}", QualifiedNameSource.IMPORT
```
bgw: It seems really iffy to say that this was sourced from an `IMPORT`. I think we should have a `QualifiedNameSource.BUILTIN`.

libcst/metadata/tests/test_scope_provider.py
```diff
@@ -596,3 +598,81 @@ def test_with_asname(self) -> None:
                 ensure_type(m.body[0], cst.With).items[0].asname, cst.AsName
             ).name,
         )
+
+    def test_get_fully_qualified_names_for(self) -> None:
```
DragonMinded: Can we add a simple test for nested classes and nested functions? I don't see that tested and it looks like from your code that you mean to handle it.

libcst/metadata/scope_provider.py
jimmylai: This is a feature supported in the ast scope API. I guess it's used in some lint use cases for typing.
In the use case of typing, the application may want to know the qualified name of a type object or simple string to verify something.
@bgw do you have comments?

bgw: You could accept the wrapping `libcst.Annotation` instead of a string, but then it's less useful because there are some cases where you can have an annotation outside of a `libcst.Annotation`, e.g.:

```
T = TypeVar("T", bound="List[int]")
```

or even

```
# This a type alias, but it's hard to tell from the CST.
# It might not even be used in this module, but instead exported
# and used as a type alias by other modules.
MyTypeAlias = "List[Union[int, str]]"
```

It's impossible to figure out every place a string might be used as a type annotation without doing a full cross-module type-check first. Even pyre struggles with this.

I think since we don't have a clear idea of what to do here, maybe we should just drop this functionality. It's always easier to add functionality later than it is to remove or fix buggy functionality (without breaking backwards compatibility).

I think it's fair to document that string-wrapped types aren't supported by this API.

libcst/metadata/scope_provider.py
jimmylai: Added QualifiedNameSource.BUILTIN.

libcst/metadata/scope_provider.py
jimmylai: Fixed.

libcst/metadata/scope_provider.py
jimmylai: Fixed.

libcst/metadata/scope_provider.py
jimmylai: Fixed.

libcst/metadata/scope_provider.py
jimmylai: Added TODO.

libcst/metadata/scope_provider.py
jimmylai: I prefer to follow PEP-3155, added `<locals>` to qualified name contains function namespace.

libcst/metadata/tests/test_scope_provider.py
jimmylai: Added test_get_qualified_names_for_nested_cases.

libcst/metadata/scope_provider.py
jimmylai: Dropped the support of SimpleString and add related document/test for now.

docs/source/metadata.rst
```diff
@@ -133,6 +133,19 @@ There were four different type of scope in Python: :class:`~libcst.metadata.Glob
 .. autoclass:: libcst.metadata.ClassScope
 .. autoclass:: libcst.metadata.ComprehensionScope
 
+Qualified Name Metadata
+-----------------------
+Qualified name provides an unambiguous name to locate the definition of variable and it's
+introduced for class and function in `PEP-3155 <https://www.python.org/dev/peps/pep-3155/>`_.
+QualifiedNameProvider provides possible :class:`~libcst.metadata.QualifiedName` given a
+:class:`~libcst.CSTNode`.
+We don't call it `fully qualified name <https://en.wikipedia.org/wiki/Fully_qualified_name>`_
```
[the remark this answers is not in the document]
jimmylai: Nice catch, added a blank line.

PR #80, opened 2019-09-27
Clarify license

DragonMinded:
## Summary

Reproduce copyright headers in files that were forked from repositories other than LibCST in the LICENSE file. Add a copy of the PSF license to license file. Add a link to Apache 2.0 license in the license file. Add an acknowledgements section. Add a future section.

## Test Plan

Pushed branch, looked at the rendered readme and license files.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,27 @@
+All contributions towards LibCST are MIT licensed.
+
+Some Python files have been taken from the standard library and are therefore
+PSF licensed. Modifications on these files are dual licensed (both MIT and
+PSF). These files are:
+
+- libcst/_parser/base_parser.py
+- libcst/_parser/parso/utils.py
+- libcst/_parser/parso/pgen2/generator.py
+- libcst/_parser/parso/pgen2/grammar_parser.py
+- libcst/_parser/parso/python/token.py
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -19,3 +43,58 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
+-------------------------------------------------------------------------------
+
+PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2
+
+1. This LICENSE AGREEMENT is between the Python Software Foundation
```
jimmylai: Why not just add the link to the licence file just like Apache license?

https://opensource.org/licenses/Python-2.0

DragonMinded: We had the choice to do either. The link to Apache license is because Apache 2.0 license is rather long. In this case, since both MIT and PSF are short, its easier on the reader to reproduce the entire license here so as to do one less look-up.

PR #84, opened 2019-10-01
Matchers Documentation

DragonMinded:
## Summary

As promised, this is the documentation for matchers. Included in this is documentation on how to use matchers, as well as a tutorial to help get started and show off various concepts. Also, I was naughty and snuck in two things: 1) A fix to a warning when generating docs. 2) Fixed matchers being incapable of combining @call and @visit/@leave even though it was designed to do so. Those are at least kept in separate commits.

## Test Plan

tox, also look at the generated docs by CI (the pages are far too big to screenshot here).

### DIFF
--- a/docs/source/matchers_tutorial.ipynb
+++ b/docs/source/matchers_tutorial.ipynb
@@ -1,0 +1,428 @@
+{
+ "cells": [
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "=====================\n",
+    "Working with Matchers\n",
+    "=====================\n",
+    "Matchers provide a flexible way of comparing LibCST nodes in order to build ",
+    "more complex transforms. See :doc:`Matchers <matchers>` for the complete ",
+    "documentation.\n",
+    "\n",
+    "Basic Matcher Usage\n",
+    "===================\n",
+    "Let's say you are visiting a LibCST :class:`~libcst.Call` node and you want ",
+    "to know if all arguments provided are the literal ``True`` or ``False``. ",
+    "You look at the documentation and see that ``Call.args`` is a sequence of ",
+    ":class:`~libcst.Arg`, and each ``Arg.value`` is a :class:`~libcst.BaseExpression`. ",
+    "In order to verify that each argument is either ``True`` or ``False`` you ",
+    "would have to first loop over ``node.args``, and then check ",
+    "``isinstance(arg.value, cst.Name)`` for each ``arg`` in the loop before ",
+    "finally checking ``arg.value.value in (\"True\", \"False\")``. \n",
+    "\n",
+    "Here's a short example of that in action:\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "nbsphinx": "hidden"
+   },
+   "outputs": [],
+   "source": [
+    "import sys\n",
+    "sys.path.append(\"../../\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import libcst as cst\n",
+    "\n",
+    "def is_call_with_booleans(node: cst.Call) -> bool:\n",
+    "    for arg in node.args:\n",
+    "        if not isinstance(arg.value, cst.Name):\n",
+    "            # This can't be the literal True/False, so bail early.\n",
+    "            return False\n",
+    "        if cst.ensure_type(arg.value, cst.Name).value not in (\"True\", \"False\"):\n",
+    "            # This is a Name node, but not the literal True/False, so bail.\n",
+    "            return False\n",
+    "    # We got here, so all arguments are literal boolean values.\n",
+    "    return True\n"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "We can see from a few examples that this does work as intended. ",
+    "However, it is an awful lot of boilerplate that was fairly cumbersome to write.\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "call_1 = cst.Call(\n",
+    "    func=cst.Name(\"foo\"),\n",
+    "    args=(\n",
+    "        cst.Arg(cst.Name(\"True\")),\n",
+    "    ),\n",
+    ")\n",
+    "is_call_with_booleans(call_1)\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "call_2 = cst.Call(\n",
+    "    func=cst.Name(\"foo\"),\n",
+    "    args=(\n",
+    "        cst.Arg(cst.Name(\"None\")),\n",
+    "    ),\n",
+    ")\n",
+    "is_call_with_booleans(call_2)\n"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "Let's try to do a bit better with matchers. We can make a better function ",
+    "that takes advantage of matchers to get rid of both the instance check and ",
+    "the ``ensure_type`` call, like so:\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import libcst.matchers as m\n",
+    "\n",
+    "def better_is_call_with_booleans(node: cst.Call) -> bool:\n",
+    "    for arg in node.args:\n",
+    "        if not m.matches(arg.value, m.Name(\"True\") | m.Name(\"False\")):\n",
+    "            # Oops, this isn't a True/False literal!\n",
+    "            return False\n",
+    "    # We got here, so all arguments are literal boolean values.\n",
+    "    return True\n"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "This is a lot shorter and is easier to read as well! We made use of the ",
+    "fact that matchers handles instance checking for us in a safe way. We also ",
+    "made use of the fact that matchers allows us to concisely express multiple ",
+    "match options with the use of Python's or operator. We can also see that ",
+    "it still works on our previous examples:\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "better_is_call_with_booleans(call_1)\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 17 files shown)


DragonMinded:
Generated docs links are https://2589-200896124-gh.circle-artifacts.com/0/doc/matchers.html and https://2589-200896124-gh.circle-artifacts.com/0/doc/matchers_tutorial.html

docs/source/matchers_tutorial.ipynb
```diff
@@ -0,0 +1,377 @@
...
+   },
+   "source": [
+    "=====================\n",
+    "Working with Matchers\n",
+    "=====================\n",
+    "Matchers provide a flexible way of comparing LibCST nodes in order to build more complex transforms. See :doc:`Matchers <matchers>` for the complete documentation.\n",
+    "\n",
+    "Basic Matcher Usage\n",
+    "===================\n",
+    "Lets say you are visiting a LibCST ``Call`` node and you want to know if all arguments provided are the literal ``True`` or ``False``. You look at the documentation and see that ``Call.args`` is a sequence of ``Arg``, and each ``Arg.value`` is a ``BaseExpression``. In order to verify that each argument is either ``True`` or ``False`` you would have to first loop over ``node.args``, and then check ``isinstance(arg.value, cst.Name)`` for each ``arg`` in the loop before finally checking ``arg.value.value in (\"True\", \"False\")``. \n",
```
jimmylai: Can you link Call to :class:`~libcst.Call`? That's easier for user who don't familiar with CSTNode.

jimmylai: Same comment for Arg and BaseExpression.

docs/source/matchers_tutorial.ipynb
```diff
@@ -0,0 +1,377 @@
...
+    "    )\n"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "We've turned our original function into a single call to ``matches``. As an added benefit, the match node can be read from left to right in a way that makes sense in english: \"match any call with zero or more arguments that are the literal ``True`` or ``False``.\". As we can see, it works as intended: "
```
jimmylai: Do we need the dot inside the double quotes? It looks odd in the generated doc. 
image

docs/source/matchers_tutorial.ipynb
```diff
@@ -0,0 +1,377 @@
...
+    "better_is_call_with_booleans(call_2)\n"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "We still have one more trick up our sleeve though. Matchers don't just allow us to specify which attributes we want to match on exactly. It also allows us to specify rules for matching sequences of nodes, like the list of ``Arg`` nodes that appears in ``Call``. Lets make use of that, turning our original ``is_call_with_booleans`` function into a call to ``matches``:"
```
jimmylai: link `matches` to its document.

docs/source/matchers_tutorial.ipynb
```diff
@@ -0,0 +1,377 @@
...
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "Matcher Decorators\n",
+    "==================\n",
+    "You can already do a lot with just ``matches``. It lets you define the shape of nodes you want to match and LibCST takes care of the rest. However, you still need to include a lot of boilerplate into your visitors in order to identify which nodes you care about. Matcher decorators help reduce that boilerplate.\n",
```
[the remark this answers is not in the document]
jimmylai: link `visitors` to visitor page incase user don't learn it yet.
https://2589-200896124-gh.circle-artifacts.com/0/doc/visitors.html#visitors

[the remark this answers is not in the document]
jimmylai: linkt `Matcher decorators` to decorator page https://2589-200896124-gh.circle-artifacts.com/0/doc/matchers.html#decorators

docs/source/matchers_tutorial.ipynb
```diff
@@ -0,0 +1,377 @@
...
+    "print(new_module.code)\n"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "Using matcher decorators we successfully removed all of the boilerplate around state tracking! The only thing that ``leave_Name`` needs to concern itself with is the actual business logic of the transform. However, it still needs to check to see if the value of the node should be inverted. This is because the ``Call.func`` is a ``Name`` in this case. Lets use another matcher decorator to make that problem go away:"
```
jimmylai: This use case has the need of attribute visitor/transformer.
If we can have something like this, it's easier and more straightforward.
```
@m.call_if_inside(m.Call(args=(
        m.ZeroOrMore(m.Arg(m.Name("True") | m.Name("False"))),
    )).args)  # we ask to only look into args attribute
def leave_Name(self, original_node: cst.Name, updated_node: cst.Name) -> cst.Name:
    ...
```

When we see more use cases of attribute matcher come up, we can consider adding the support.

jimmylai: Link Call.func and Name to its own documents.

docs/source/matchers_tutorial.ipynb
```diff
@@ -0,0 +1,377 @@
+{
```
jimmylai: This tutorial has a couple of `lets` (third-person singular form)  should be more correct if change to `let's` (let us).

docs/source/matchers_tutorial.ipynb
DragonMinded: Will change.

docs/source/matchers_tutorial.ipynb
DragonMinded: I think that's a copy mistake, I can fix it.

docs/source/matchers_tutorial.ipynb
DragonMinded: Good idea.

docs/source/matchers_tutorial.ipynb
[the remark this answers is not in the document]
DragonMinded: Good ideas.

docs/source/matchers_tutorial.ipynb
DragonMinded: I'm not sure how far I want to go with matcher decorators, and right now we don't have support for checking that you're inside an attribute (in the non-decorator case they do not make sense). This was an intentional omission because there's no syntax that works in the general case.

DragonMinded: Good idea.

docs/source/matchers_tutorial.ipynb
DragonMinded: Will change.

jimmylai:
Found this issue from the code committed earlier. The `:class:BaseMatcherNode` didn't link to its document properly.
image

DragonMinded:
Updated artifacts with fixes for above comments: https://2633-200896124-gh.circle-artifacts.com/0/doc/matchers.html and https://2633-200896124-gh.circle-artifacts.com/0/doc/matchers_tutorial.html

jimmylai:
The term "logical or operator" may cause confusion to users. Python's logical or operator should be `or` rather than `|`, `|` is called `Bitwise Or`. https://docs.python.org/3/library/operator.html#mapping-operators-to-functions

https://2589-200896124-gh.circle-artifacts.com/0/doc/matchers.html#libcst.matchers.OneOf
image

DragonMinded:
Fixed logical or/bitwise or documentation mistake.

libcst/matchers/_matcher_base.py
```diff
@@ -123,10 +153,40 @@ def __repr__(self) -> str:
...
     """
-    Matcher node that matches all of its options. Useful when you want to match
-    against a concrete matcher and a :class:`MatchIfTrue` at the same time. You
-    can also construct a :class:`AllOf` matcher by using Python's and operator
-    (``&``) with generated matcher classes.
+    Matcher that matches all of its options. Useful when you want to match
+    against a concrete matcher and a :class:`MatchIfTrue` at the same time. Also
+    useful when you want to match against a concrete matcher and a
+    :func:`DoesNotMatch` at the same time. You can also construct a
+    :class:`AllOf` matcher by using Python's boolean and operator with concrete
```
jimmylai:
```suggestion
    :class:`AllOf` matcher by using Python's bitwise and operator with concrete
```

libcst/matchers/_matcher_base.py
```diff
@@ -342,16 +484,41 @@ def ZeroOrOne(
...
+    This will match any function call that has two or three arguments, and
+    the first and last arguments are strings::
+
+        m.Call(args=(m.Arg(m.SimpleString()), m.ZeroOrOne(), m.Arg(m.SimpleString())))
+
     """
     return cast(AtMostN[Union[_MatcherT, DoNotCareSentinel]], AtMostN(matcher, n=1))
 
 
 def DoesNotMatch(obj: _OtherNodeT) -> _OtherNodeT:
```
jimmylai: Question, this is analogy to `matches()`, why not call it `do_not_matches()`

libcst/matchers/_matcher_base.py
```diff
@@ -229,14 +316,18 @@ def __repr__(self) -> str:
...
-def MatchRegex(
-    regex: Union[str, Pattern[AnyStr]]
-) -> MatchIfTrue[Callable[[str], bool]]:
+def MatchRegex(regex: Union[str, Pattern[str]]) -> MatchIfTrue[Callable[[str], bool]]:
     """
-    Used as a convenience wrwapper to :class:`MatchIfTrue` which allows for
-    matching against a regex. This should be used against attributes which
-    are strings, to allow matching regular expressions. An example would be
-    ``m.Name(m.MatchRegex(r'[A-Za-z]+'))``.
+    Used as a convenience wrapper to :class:`MatchIfTrue` which allows for
```
jimmylai: It worth mentioned that MatchRegex uses Python builtin `re` module (so special sequences is available e.g. \d \w) and link to re doc https://docs.python.org/3/library/re.html. 
Also mentions it takes either a regex string or a `Pattern`.

libcst/matchers/_matcher_base.py
```diff
@@ -77,18 +80,45 @@ def __invert__(self: _BaseMatcherNodeSelfT) -> "_BaseMatcherNodeSelfT":
 def DoNotCare() -> DoNotCareSentinel:
     """
     Used when you want to match exactly one node, but you do not care what node it is.
-    Useful inside sequences such as function def and argument matchers. You do not
-    need to use this for attributes since :func:`DoNotCare` is already the default.
+    Useful inside sequences such as a :class:`libcst.matchers.Call`'s args attribte.
```
jimmylai: The libcst.matchers.Call link doesn't work. https://2645-200896124-gh.circle-artifacts.com/0/doc/matchers.html
(When it works, it's clickable and shows only the simple class name Call.)
image

DragonMinded: I can change that back, but this is intentional since we don't have a page rendering out concrete classes. They're identical to the LibCST classes so I don't think adding them gives us any benefit.

libcst/matchers/_matcher_base.py
DragonMinded: Can do.

libcst/matchers/_matcher_base.py
DragonMinded: Its not the opposite of `matches()`. Its a predicate used inside matchers. That's why it isn't named opposite of `matches()`.

libcst/matchers/_matcher_base.py
DragonMinded: Will fix.

PR #87, opened 2019-10-01
Add deep_remove and RemoveFromParent() to LibCST.

DragonMinded:
## Summary

We've gotten a few comments that RemovalSentinel is a bit difficult to use. Its not that intuitive to return RemovalSentinel.REMOVE. We already have helper methods on `CSTNode` which return new values (`deep_clone`, `with_changes`) so we have precedent here for adding another. Add a `deep_remove` onto CSTNode that can be called similar to the other modifying calls, and identically to `deep_replace` which was newly added. Add a `RemoveFromParent()` helper method that constructs a `RemovalSentinel.REMOVE` and refer to that as the default in documentation. Hopefully this helps the ergonomics of removing nodes a bit.

## Test Plan

tox, look at generated docs.

### DIFF
--- a/docs/source/visitors.rst
+++ b/docs/source/visitors.rst
@@ -5,6 +5,7 @@ Visitors
 
 .. autoclass:: libcst.CSTVisitor
 .. autoclass:: libcst.CSTTransformer
+.. autofunction:: libcst.RemoveFromParent
 .. autoclass:: libcst.RemovalSentinel
 
 Visit and Leave Helper Functions
--- a/libcst/__init__.py
+++ b/libcst/__init__.py
@@ -185,7 +185,7 @@ from libcst._nodes.whitespace import (
 )
 from libcst._parser.entrypoints import parse_expression, parse_module, parse_statement
 from libcst._parser.types.config import PartialParserConfig
-from libcst._removal_sentinel import RemovalSentinel
+from libcst._removal_sentinel import RemovalSentinel, RemoveFromParent
 from libcst._visitors import CSTNodeT, CSTTransformer, CSTVisitor, CSTVisitorT
 from libcst.metadata.base_provider import (
     BaseMetadataProvider,
@@ -208,6 +208,7 @@ __all__ = [
     "MetadataException",
     "ParserSyntaxError",
     "PartialParserConfig",
+    "RemoveFromParent",
     "RemovalSentinel",
     "ensure_type",
     "visit_batched",
--- a/libcst/_nodes/base.py
+++ b/libcst/_nodes/base.py
@@ -55,7 +55,9 @@ class _ChildrenCollectionVisitor(CSTVisitor):
 
 
 class _ChildrenReplacementTransformer(CSTTransformer):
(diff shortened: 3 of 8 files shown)


DragonMinded:
Ugh, the base of this commit somehow made it into my other PR that just landed when I interactive rebased to address some comments, so include https://github.com/Instagram/LibCST/commit/271566bdbf32a4bf52b45649a2308ae47c14e4c2 as part of your review.

jimmylai:
What are the example CSTNodes cannot be removed?

One concern about the remove() API is that it's not specific to the node and may cause confusion.
When calling deep_clone() and with_changes() on a node, we expect something specific to the node node (clone or apply some changes). remove is different in this case. User can do things like those and they behalves the same which may not meet the user's expectation:

```
    return updated_node.remove()
```

```
    return updated_node.value.remove()
```
Will user try to do the second remove to remove the value from the node? How can we prevent this confusion?

One solution I can think of is to make the remove as a staticmethod. Not sure if it's a better solution.
```
def leave_Arg(self, original_node: cst.Arg, updated_node: cst.Arg) -> Union[cst.Arg, cst.RemovalSentinel]:
                return cst.Arg.remove()
```

jimmylai:
Another thought: what about user just call remove to mark the node to be removed and we remove all nodes at the end of traversal?
That means user can just call the remove() but don't need to return .remove() in the leave function.
For example, if users want to remove all annotation in a FunctionDef, they can do this:

```
def visit_FunctionDef(self, original_node, updated_node):
    if updated_node.returns:
        updated_node.returns.remove()
    for param in updated_node.params.params:
        if param.annotation:
            param.annotation.remove()
    return updated_node
```

DragonMinded:
Examples of nodes that can't be removed are things like operators. They are required for things like binary operations and augmented assignments. It makes no sense to allow their removal, because what you really want is to remove the whole parent node. The parent node is often an expression or also a non-removal node in the context its in, so that also doesn't make sense. We calculate if a node is ever removable and include that information in the return value of the various `leave_<Node>` overrides that we codegen if you are curious to see exact ones.

I don't think we can support `deep_remove()` or nested remove calls using this syntax and I think your point is good that this will look enticing to use in a way similar to `updated_node.returns.remove()` so I am going to ditch this in favor of a shorter constructor helper. I am trying to explore ways to make removing a node more ergonomic (I still forget what the sentinel value is), and I think I will try again with a top level helper function like `RemoveFromParent()` which just returns `RemovalSentinel.REMOVE`. That makes a lot of sense from an action perspective and crucially cannot be applied to nested nodes.

I also am going to add a `deep_remove()` similar to `deep_replace()` in the other PR I have up, to fill in that hole.

DragonMinded:
Okay, this has been revamped with the described changes.

DragonMinded:
Updated the test that I added to make it more clear what's going on with `deep_remove()` in case somebody uses that for an example.

DragonMinded:
I can't figure out HOW to update the PR title. I will google it...

PR #91, opened 2019-10-02
Improve Scope analysis and add a tutorial doc

jimmylai:
## Summary
Changes:
1. add `Scope.get_assignments_for()` to make accessing Assignments easier while traversing CST tree.
With existing API, user needs to parse the name string from CSTNode and then use `scope[name]`.
This new API is similar to `get_qualified_names_for()` but returns assignments instead when given a CSTNode.

2. implement Scope.__iter__ for accessing all assignments in current scope.

3. improve document of Scope: after vs before
image

4. add scope tutorial to demonstrate some advanced use cases.
  - Automatically Remove Unused Import

https://2776-200896124-gh.circle-artifacts.com/0/doc/scope_tutorial.html

## Test Plan
unit tests.

### DIFF
--- a/docs/source/scope_tutorial.ipynb
+++ b/docs/source/scope_tutorial.ipynb
@@ -1,0 +1,191 @@
+{
+ "cells": [
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "==============\n",
+    "Scope Analysis\n",
+    "==============\n",
+    "Scope analysis keeps track of assignments and accesses which could be useful for code automatic refactoring. If you're not familiar with Scope analysis, see :doc:`Metadata <metadata>` for more detail about Scope metadata. This tutorial demonstrates some use cases of Scope analysis.\n",
+    "\n",
+    "Automatically Remove Unused Import\n",
+    "==================================\n",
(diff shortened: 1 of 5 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -284,6 +292,14 @@ def f(self) -> "c":
                 )
         return results
 
+    def get_assignments_for(self, node: cst.CSTNode) -> Collection[BaseAssignment]:
+        """ Get all assignments in current scope given a :class:`~libcst.CSTNode`. """
+        full_name = _QualifiedNameUtil.get_full_name_for(node)
+        if full_name is None:
+            return set()
+        parts = full_name.split(".")
+        return self[parts[0]]
```
DragonMinded: I'm not sure on the implications of claiming that "a.b" assignments are equivalent to "a" assignments. Do we have this documented anywhere?

docs/source/scope_tutorial.ipynb
```diff
@@ -0,0 +1,191 @@
...
+    "    print(scope)"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "We're interested in those unused names and we traverse all import assignments in each scope to build a lookup table ``import_names_to_remove`` to record the mapping of import node and their unused names. There exist other types of assignment like :class:`~libcst.FunctionDef` or :class:`~libcst.ClassDef` but we don't care them here."
```
DragonMinded: Missing about in "care them here".

libcst/metadata/scope_provider.py
jimmylai: We decided to use Assignments class to replace this function and we'll clarify the current scope analysis only handles local variable name.

jimmylai:
Comments are addressed and added to PR #94.
We decided to replace  get_assignments_for with Assignments/Accesses.
Also improved the tutorial.

PR #92, opened 2019-10-04
Upgrade Pyre to 0.32.

DragonMinded:
## Summary

This fixes a TODO, removes several ignore statements and most importantly, addresses a massive speed regression triggered by importing matchers.

## Test Plan

pyre!


libcst/_metadata_dependent.py
```diff
@@ -2,9 +2,10 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-
+#
```
DragonMinded: This works around a bug in isort.

DragonMinded:
Yes, that should fix that, it is much faster again.

PR #93, opened 2019-10-04
rewrite ExpressionContextProvider to handle nested List/Tuple and Attribute

jimmylai:
## Summary
Nested List and Tuple are not analyzed correctly previously.
E.g. when nested List or Tuple used as assignment targets, all simple Name in List or Tuple should have STORE context.
In `((a, b), c) = ((1, 2), 3)`, name a/b/c has STORE context. Same for `[[a, b], c] = [[1, 2], 3]`

In LibCST, Attribute.attr is a Name but it's a str in ast.
In ast expression context, Name has LOAD context when it is a local variable. To honor that, we only set context to left most Name in attribute access.
E.g. In `a.b.c = 1`, only Name("a") has STORE context, b/c has None as context.

New added Assignments and Accesses APIs.
https://github.com/Instagram/LibCST/pull/93/commits/a0557123f309c925ecf0a86507e8af1376421817

## Test Plan
Unit  tests.

### DIFF
--- a/libcst/metadata/expression_context_provider.py
+++ b/libcst/metadata/expression_context_provider.py
@@ -6,7 +6,7 @@
 # pyre-strict
 
 from enum import Enum, auto
-from typing import Optional
+from typing import Optional, Sequence
 
 import libcst as cst
 from libcst.metadata.base_provider import BatchableMetadataProvider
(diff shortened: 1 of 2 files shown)


libcst/metadata/expression_context_provider.py
```diff
@@ -53,82 +53,114 @@ class ExpressionContext(Enum):
...
 
-    def visit_Assign(self, node: cst.Assign) -> None:
+    def visit_Assign(self, node: cst.Assign) -> bool:
         for target in node.targets:
-            self._check_type_and_set_metadata(target.target, ExpressionContext.STORE)
+            target.visit(
+                ExpressionContextVisitor(self.provider, ExpressionContext.STORE)
+            )
+        node.value.visit(self)
+        return False
```
DragonMinded: I still really like the trick of stopping execution of the current visitor and swapping in a different one for the subtree. Its really elegant.

libcst/metadata/scope_provider.py
```diff
@@ -98,6 +100,65 @@ class BuiltinAssignment(BaseAssignment):
     pass
 
 
+def _get_name_for(node: Union[str, cst.CSTNode]) -> Optional[str]:
```
bgw: Could we rename `_QualifiedNameUtil` to `_NameUtil`, and put `get_name_for` as a static method inside `_NameUtil`?

I think it'd be easier to understand if the implementations of `get_full_name_for` and `get_name_for` were adjacent.

libcst/metadata/scope_provider.py
```diff
@@ -98,6 +100,65 @@ class BuiltinAssignment(BaseAssignment):
...
+        name = _get_name_for(node)
+        return set(self._assignments[name]) if name in self._assignments else set()
+
+    def __contains__(self, node: Union[str, cst.CSTNode]) -> bool:
+        """Check if a name str or :class:`~libcst.CSTNode` has any assignment by ``node in scope.assignments``"""
+        return len(self[node]) > 0
+
+
+class Accesses:
+    """A container to provide all assignments in a scope."""
```
bgw: s/assignments/accesses/

jimmylai:
To make PR review easier, I removed the changes of Assignments and Accesses. Will create another PR for it.

PR #97, opened 2019-10-09
Matchers 💜 metadata

DragonMinded:
## Summary

This is an RFC for adding metadata capabilities to matchers. This is done by adding two things to matcher base. First, it adds a 'metadata' field to all generated concrete matcher classes which allows for you to specify a mapping of metadata provider to desired metadata. All matcher predicate operations work here, so you can specify a `OneOf()`, `AllOf()` or `DoesNotMatch()` for a metadata value, as well as doing the same thing to the mapping itself. Flexibility is provided so that the developer can use whatever is considered more convenient and readable for the situation. Second, it adds a `MatchMetadata` special matcher which matches any node type as long as the metadata contained within matches. This is effectively a `DoNotCare()` wildcard matcher, except it matches on metadata. It takes a single initializer 'metadata' field which operates identically to the above specifications.

Given that we need to be able to look up metadata now, I've added an optional kwonly-arg `metadata_resolver` to `matches` which I default to False. I allow you to pass either an instance of `MetadataWrapper` or `MetadataDependent` to it with appropriate lookup functions under the hood. This allows `matches` to continue to be used both within and outside of transforms, much like metadata. For convenience, I've added a `matches` method to the matcher decoratable visitors so that you don't have to pass `metadata_resolver=self` to all matcher calls. I have also made sure to hook up `self` to all internal calls to `matches` so that matcher decorators automatically get metadata support. For transforms/visitors, the same requirement for explicitly using metadata still exists: You must declare your `METADATA_DEPENDENCIES` to include those that you match on. So, consistency is maintained between explicit metadata lookup and matchers that use metadata.

A few things came up that I considered and ultimately discarded in this implementation. First, I considered updating `DoNotCare()` to allow specifying metadata but rejected this as it overloads the meaning of `DoNotCare()` and I find it confusing. I find a more explicit `MatchMetadata()` for a wildcard that only pins down metadata to be a lot more readable. I also considered adding a helper `matches` method to `CSTNode` which works similar to the `matches` helper on decorators, but instead takes in a match class and an optional metadata resolver. I rejected this because I want to keep the libcst core free of matchers in the case that somebody does not want to use them. I also considered other, easier to type metadata specification interfaces such as specifying the mapping of metadata provider to value as a list of two-item tuples or wrapper classes. This would allow me to make both the provider and the value covariant and would allow Pyre to catch the case where you specify a class as a metadata provider which isn't actually a provider. I found that using a mapping just as the name implies, to provide a map of provider to value, was far more intuitive. So, I rejected the slightly more type-safe variant and instead went with a more convenient and readable API.

I did not include documentation updates in this PR because I am not sure we will be satisfied with the API. Once we lock that down and we are okay with the implementation I will add additional documentation discussing how to make matchers and metadata best friends.

This also complicates types further in libcst/matchers/__init__.py and I find that they are past the point of readability. I have started to refactor the codegen section in this PR to make it such that we can do type-related reductions applied across all of our codegen. However, until we are sure we like this current proposal I'm not going to further refine the codegen. Ultimately I would like to vastly simplify types so they are more readable.

## Test Plan

Added a decent amount of tests, pyre comes out clean. I needed to add one hack (`Type[Any]`) because there is no `Mapping` type in Python that is covariant for the key type, but I think I will circle around and use Protocols to fix this hole. I also need to add more tests to verify that `@call_if_inside` and `@call_if_not_inside` work, but I have high confidence that they are working.


libcst/matchers/_matcher_base.py
```diff
@@ -76,6 +84,11 @@ def __and__(
     def __invert__(self: _BaseMatcherNodeSelfT) -> "_BaseMatcherNodeSelfT":
         return cast(_BaseMatcherNodeSelfT, InverseOf(self))
 
+    def __hash__(self) -> int:
+        # Allow matchers whose attributes are sequences or mappings to be
+        # used as a dictionary key.
+        return hash(repr(self))
```
carljm: Where does this diff add using matchers as dictionary keys?

This hash definition is kinda smelly and violates a number of the guidelines in  https://docs.python.org/3/reference/datamodel.html#object.__hash__ (e.g. you should never define `__hash__` on a class that does not also define `__eq__`, you should not define `__hash__` for mutable objects). It's hard to evaluate the actual risk without seeing why this is needed.

libcst/matchers/_matcher_base.py
```diff
@@ -344,6 +357,46 @@ def _match_func(value: object) -> bool:
...
+    # key, which is what we need here. So, lets loosen the type.
+    def __init__(self, metadata: Mapping[Type[Any], object]) -> None:
+        self.metadata: Mapping[meta.ProviderT, object] = metadata
+
+    def __or__(self, other: _OtherNodeT) -> "OneOf[Union[MatchMetadata, _OtherNodeT]]":
+        # Without the cast, pyre thinks that this returns OneOf[object]
+        return cast(OneOf[Union[MatchMetadata, _OtherNodeT]], OneOf(self, other))
+
+    def __and__(self, other: _OtherNodeT) -> "AllOf[Union[MatchMetadata, _OtherNodeT]]":
+        # Without the cast, pyre thinks that this returns OneOf[object]
```
carljm: I presume this is a copy-paste foul and should read `AllOf[object]`

libcst/matchers/tests/test_matchers_with_metadata.py
```diff
@@ -0,0 +1,473 @@
...
+import libcst as cst
+import libcst.matchers as m
+import libcst.metadata as meta
+from libcst.matchers import matches
+from libcst.testing.utils import UnitTest
+
+
+class MatchersMetadataTest(UnitTest):
+
+    ONCALL_SHORTNAME = "instagram_server_framework"
```
carljm: Remove?

libcst/codegen/gen_matcher_classes.py
```diff
@@ -237,7 +241,9 @@ def leave_Subscript(
     ) -> cst.Subscript:
         if updated_node.value.deep_equals(cst.Name("Union")):
             # Take the original node, remove do not care so we have concrete types.
-            concrete_only_expr = _remove_do_not_care(original_node)
+            concrete_only_expr = _remove_match_metadata(
+                _remove_do_not_care(original_node)
```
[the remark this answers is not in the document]
DragonMinded: That might help some of the speed issues with generating this file. I haven't profiled it but it is definitely slow and combining transforms would probably help.

libcst/codegen/transforms.py
```diff
@@ -0,0 +1,18 @@
...
+    ) -> cst.BaseExpression:
+        if updated_node.value.deep_equals(cst.Name("Union")):
+            slc = updated_node.slice
+            if isinstance(slc, cst.Slice):
+                raise Exception("Unexpected Slice in Union!")
+            elif isinstance(slc, cst.Index):
+                return slc.value
+            elif len(slc) == 1:
+                # This is a Union[SimpleType] which is equivalent to
+                # just SimpleType
```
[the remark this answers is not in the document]
DragonMinded: You're right about this, and I should update the comment. In the spot it was copied from, it could be that it was previously modified in the tree and only one `ExtSlice` was left, since I wasn't rendering out between passes. This is one particular spot in the grammar where either an `Index` or a single `ExtSlice` can render to the same thing. "Fixing" that would make it hard to generate code by adding `ExtSlice` one by one though, and we will never parse to a single `ExtSlice` so I think its fine to just update the comment here.

libcst/matchers/_matcher_base.py
DragonMinded: Matchers are used as keys in matcher decorators. Without this, you can't have a matcher decorator that includes metadata matching. Possibly a dubious design decision previously on my part, but one that is present at the time of implementing this nonetheless. I briefly considered adding an `__eq__` and all of these are *technically* immutable in the fact that they're not designed to be messed with once constructed, but I guess your point is valid. I can do a bit of extra work to expose read-only properties or convert some of these types to frozen dataclasses (doesn't matter either way) and then make sure there's an `__eq__` as well. For concrete types, they're all frozen dataclasses so they're "guaranteed" to be immutable, but the point still stands. Matchers *should* be immutable because they match on LibCST nodes, which *are* immutable. I can shore that up and close the gap.

libcst/matchers/_matcher_base.py
DragonMinded: Yerp, will fix when I iterate on this.

libcst/matchers/tests/test_matchers_with_metadata.py
DragonMinded: Haha whoops. I'll scan the repo and remove those.

DragonMinded:
I'm kinda sold on moving to `MetadataMatcher` with one concern: I think that having a magic `metadata` attribute on each concrete class absolutely hints at the fact that its okay to match on nodes AND metadata at once, since metadata is conceptually a property of the node. We have the need for a standalone `MatchMetadata` because we want to be able to match on wildcard nodes as long as metadata matches, but we also want an ergonomic way to match on a concrete node *and* metadata at the same time. Expecting people to use AllOf or the `&` constructor to match up metadata and concrete nodes is probably non-obvious and non-ergonomic, so I would propose still having two pieces. I *think* this is close to what you were proposing, or maybe I didn't understand entirely.

So, given the precident of `MatchIfTrue`, `DoesNotMatch` and `MatchRegex`, I would propose keeping the name `MatchMetadata` for the loose matcher, and then change the type of `metadata` attribute to `Union[OneOf[MatchMetadata], AllOf[MatchMetadata], MatchMetadata]` so it can be stacked inside a concrete matcher. This aligns with the fact that specifying multiple attributes on a concrete node is conceptually identical to an `AllOf` with individual concrete nodes all specifying a single attribute. Then, change `MatchMetadata` to instead take a key/value as its two constructor parameters instead of a mapping.

We would still need to keep the `__hash__` change here as I realze that we type sequences as taking `Sequence`, and `list` is a valid sequence. So, right now, if you construct a matcher and use it in a decorator, it works if you use a tuple but blows up if you use a list, which is unacceptable and non-obvious.

Thoughts?

libcst/codegen/transforms.py
```diff
@@ -0,0 +1,18 @@
+import libcst as cst
+
+
+class SimplifyUnionsTransformer(cst.CSTTransformer):
+    def leave_Subscript(
+        self, original_node: cst.Subscript, updated_node: cst.Subscript
+    ) -> cst.BaseExpression:
+        if updated_node.value.deep_equals(cst.Name("Union")):
```
jimmylai: Let's use matcher?
```suggestion
        if m.matches(updated_node, m. Subscript(value=m.Name("Union")):
```

jimmylai:
I think it's useful to have the `metadata=` attribute on matcherNode to specify metadata conditions since it's straightforward but the Mapping with type Any is concerning. We could probably combine those ideas and make the type of `metadata` as union of special matchers, so user can do Name(metadata=m.AllOf(MetadataMatcher(ExpressionContextProvider, ExpressionContext.STORE)))

The wildcard matcher MatchMetadata is also useful because user may just want to match the any node has specific metadata value.

This PR is pretty large. Some part of this PR can be separated to make the review easier, e.g. the codegen to simplify Union[SimpleString] as SimpleString.
I'm interested in seeing the test coverage report to see how coverage of new code.

libcst/codegen/transforms.py
DragonMinded: We should not use matchers in codegen because it couples more of LibCST into codegen of LibCST itself. I intentionally use the least amount of it as I can get away with.

carljm:
Oh yeah, so my comments weren't totally clear. I think keeping `metadata` attribute on matcher nodes totally makes sense, for the reasons you outlined. I was envisioning exactly what you suggested, have the type of `metadata` attribute be `Union[OneOf[MatchMetadata], AllOf[MatchMetadata], MatchMetadata]`.

I also think it's reasonable to have `MatchMetadata` work either as "attached to a node's `metadata` attribute, match metadata on that node" or "used standalone, match metadata on any node." I think that's pretty intuitive and better than the more complex options.

So I think we're on the same page!

libcst/codegen/transforms.py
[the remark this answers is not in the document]
carljm: On a tangent, it makes me wonder why indexing with a tuple is modeled as ExtSlice thing rather than just an index that is a tuple? At runtime it's just indexing with a tuple:

```
>>> class Foo:
...     def __getitem__(self, idx):
...             print(idx)
...
>>> f = Foo()
>>> f[1, 2, 3]
(1, 2, 3)
```

But I would guess there are good reasons for this that I'm not thinking of.

libcst/matchers/_matcher_base.py
carljm: Yeah, I'm not gonna be a stickler on runtime immutability if the intended uses are immutable, but if it's not hard to make them frozen dataclasses that would certainly be ideal.

I do think it would be better to define `__eq__` and to use an explicit set of fields for both equality and hashing rather than the rather implicit and potentially fragile route of relying on `repr`.

DragonMinded:
Cool, thanks for clarifying! I think myself, @carljm and @jimmylai are on the same page for the general direction this is taking, so I have some more work to do!

libcst/matchers/_matcher_base.py
carljm: I guess I'm still slightly confused why this is related to metadata, and why it needs to be added on `BaseMatcherNode` in this diff but wasn't needed before, even though BaseMatcherNode would have been used in matcher decorators before this too?

libcst/codegen/transforms.py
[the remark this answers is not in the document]
DragonMinded: I legitimately do not know either. I bet the answer is buried in one of the PEPs somewhere. Its weird that there would be this specific exemption for `ExtSlice` when it was added for numerical analysis (and used for typing) and nothing else in Python supports it.

libcst/matchers/_matcher_base.py
DragonMinded: Will do, that answers my internal question to myself on whether I should add `__eq__` or not.

DragonMinded: You're right that its not related to metadata as much as the issue was uncovered by the current implementation of metadata using a mapping. The problem exists in current code that we can't use match decorators in combination with matchers constructed with lists. I can split this out in the final PR.

DragonMinded: I think with that, I should implement `__eq__` hack on the `BaseMatcherNode` which compares in a similar manner, since I am satisfied with the `__repr__` that dataclasses generates. Essentially, this is all to get around having to codegen a `__eq__` and `__hash__` *per* generated class.

libcst/codegen/transforms.py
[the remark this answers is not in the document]
carljm: Hmm, this seems to be a LibCST specific thing. In Python's AST, `f[1, 2, 3]` parses to a Subscript node whose `.slice` attribute is an `Index` whose `.value` is `Tuple`:

```
>>> t = ast.parse("f[1, 2, 3]")
>>> t.body[0].value
<_ast.Subscript object at 0x7efeb70b08d0>
>>> t.body[0].value.slice
<_ast.Index object at 0x7efeb70b0978>
>>> t.body[0].value.slice.value
<_ast.Tuple object at 0x7efeb70b09b0>
```

Whereas in LibCST, it parses to a list of `ExtSlice`.

In Python's AST, I think the extended form of slice is only for when you have slice ranges, like `1:3`; it's not related to the use of commas/tuples in a slice.

Maybe there is a good reason for this difference, but it strikes me as possibly a bug/mistake, and maybe one that would clean up some code like this if we fixed it (and if it were gonna be fixed, sooner would be better than later.)

@bgw do you have any additional context here? If it is a bug, I can file a separate issue (and maybe work on a PR too).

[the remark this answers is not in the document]
DragonMinded: Hmm, it strikes me that this would end up making it even harder to work with `Subscript`, as you would need to be able to recognize `Slice`, `Index`, `ExtSlice` *and* `Tuple`. I think what you're proposing is technically correct, but it would mean that dealing with `Subscript` becomes even more irregular than it already is which is something we desperately tried to avoid in LibCST.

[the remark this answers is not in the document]
DragonMinded: The grammar specifies the following (simplified to remove other trailer types):

```
trailer: '[' subscriptlist ']'
subscriptlist: subscript (',' subscript)* [',']
subscript: test | [test] ':' [test] [sliceop]
sliceop: ':' [test]
```

Which implies that the fact the AST parses a `subscriptlist` into a tuple is an implementation detail for CPython. It doesn't actually specify `ExtSlice` anywhere. I think the mistake here is in allowing a `Slice`, `Index` or `Sequence[Extslice]` instead of having it ALWAYS be an `ExtSlice` which represents `subscriptlist`.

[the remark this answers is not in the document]
DragonMinded: This would also simplify `Subscript` in a similar manner, because you're always guaranteed a sequence, and would get rid of the need to check for sequence versus node types.

libcst/matchers/_matcher_base.py
carljm: Would it be fairly easy and more explicit / less hacky to dynamically get the list of fields from dataclass internal metadata for a generic `__hash__` rather than relying on `repr`? And then convert lists to tuples at that point before hashing?

What makes me slightly nervous is that in general it's eminently possible for two distinct classes/instances to end up with the same `repr`. Not sure if that could ever happen within LibCST. It's just definitely more robust to actually hash the fields' values instead of their reprs.

libcst/codegen/transforms.py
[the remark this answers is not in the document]
carljm: Interesting. I guess that special-casing of comma-separated lists in the grammar is necessary to allow range-slices within tuple elements, which otherwise would not be allowed. That must be the thing that was added to support some numerical use cases, since I think using a regular tuple as an index has been allowed for a long time (the presence of `test` in the grammar means any expression can be an index).

Given that it's definitely more reasonable to have this as a special form, I guess.

FWIW though my proposal (and matching AST) would have nested the `Tuple` inside the `Index`, where you can already have any expression at all. So it wouldn't increase the irregularity. It would decrease it, because the `.slice` of a Subscript would never be a list of anything, it would always either be an `Index` or a `Slice`.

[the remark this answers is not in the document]
carljm: Ok, the AST does parse `f[1, 2:3]` into an `ExtSlice`, because grammatically tuples can't contain a Slice. So the question here is just whether the case of `f[1, 2, 3]` should be made consistent with `f[(1, 2, 3)]` or consistent with `f[1, 2, 3:4]`. AST chooses the former, LibCST currently chooses the latter. I don't think either one has an advantage in overall regularity, and I think it makes slightly more sense for the unusual node type to be triggered by the presence of the otherwise-illegal range-slice, not by the presence of commas. But not sure it's worth changing at this point.

libcst/matchers/_matcher_base.py
carljm: Offline conversation resolved that likely the best way to handle this is to make matchers default equality and hashing based on identity, not value, the same way LibCST nodes do it.

libcst/codegen/transforms.py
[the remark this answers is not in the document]
DragonMinded: It comes down to a question of possibility of deprecation. It should be easy to modify LibCST to always parse to a `Sequence[ExtSlice]` going forward since the types allow it and existing code will not be surprised by its existing. Then we can mark a deprecation date where we no longer support construction with a raw `Index` or `Slice` and close the typing gap. Going towards tuple instead would be more difficult.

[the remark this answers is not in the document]
carljm: I summarized our conversation here in https://github.com/Instagram/LibCST/issues/101

libcst/matchers/_matcher_base.py
```diff
@@ -789,26 +869,62 @@ def _attribute_matches(  # noqa: C901
...
+    node: libcst.CSTNode,
+    metadata: Union[
+        MatchMetadata,
+        AllOf[MatchMetadata],
+        OneOf[MatchMetadata],
+        InverseOf[MatchMetadata],
+    ],
+    metadata_lookup: Callable[[meta.ProviderT, libcst.CSTNode], object],
+) -> bool:
+    # Now, evaluate the matcher node itself.
```
carljm: This comment seems out of place? Inherited from an older version of the code?

DragonMinded: I think so. Let me rebase and fix.

libcst/matchers/_matcher_base.py
```diff
@@ -789,26 +869,61 @@ def _attribute_matches(  # noqa: C901
...
+            _metadata_matches(node, metadata, metadata_lookup)
+            for metadata in metadata.options
+        )
+    elif isinstance(metadata, AllOf):
+        return all(
+            _metadata_matches(node, metadata, metadata_lookup)
+            for metadata in metadata.options
+        )
+    elif isinstance(metadata, InverseOf):
+        return not _metadata_matches(node, metadata.matcher, metadata_lookup)
```
carljm: According to Codecov, we don't have test coverage of any of these clauses, only the last one? If that's true, it seems worth fixing; this is pretty core to the functionality.

DragonMinded:
Docs are included as part of this PR. There's only one special matcher being added in this case, and its added to the matchers documentation including examples.

libcst/matchers/_matcher_base.py
DragonMinded: I'll check it out, but I thought I had tests for this.

PR #99, opened 2019-10-10
Add missing license header to matchers generator file.

DragonMinded:
## Summary

As it says on the tin!

## Test Plan

tox


DragonMinded:
Will do. I think as we mature more of LibCST and associated libraries, this will happen less and less since I won't be incubating anything inside Instagram before releasing.

PR #100, opened 2019-10-10
Make `BaseMetadataProvider[T]` covariant over `T`

bgw:
## Summary

Because we'd consider `BaseMetadataProvider[int]` to be a subtype of
`BaseMetadataProvider[object]`, it should be covariant over its
typevar, rather than invariant.

This isn't entirely correct because we have a mutable data structure
(`_computed`) that depends on the typevar, and pyre points this out
(though with a really confusing error message). However, it's not
correct to say that `BaseMetadataProvider` is invariant either, so I
think this is the lesser evil.

I don't think it's practical to redesign this API to avoid the variance
issue, so I'm ignoring the new type error that results from this change.

I think this may resolve some of the issues we've seen internally with
D17820032.

## Test Plan

```
pyre check
```

### DIFF
--- a/libcst/metadata/base_provider.py
+++ b/libcst/metadata/base_provider.py
@@ -32,12 +32,14 @@ if TYPE_CHECKING:
     from libcst.metadata.wrapper import MetadataWrapper
 
 
-ProviderT = Type["BaseMetadataProvider[Any]"]
-_T = TypeVar("_T")
+ProviderT = Type["BaseMetadataProvider[object]"]
+# BaseMetadataProvider[int] would be a subtype of BaseMetadataProvider[object], so the
+# typevar is covariant.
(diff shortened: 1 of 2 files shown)


libcst/metadata/base_provider.py
```diff
@@ -75,7 +82,11 @@ def _gen_impl(self, module: "Module") -> None:
         """
         ...
 
-    def set_metadata(self, node: "CSTNode", value: _T) -> None:
+    # pyre-ignore[46]: The covariant `value` isn't type-safe because we write it to
+    # pyre: `self._computed`, however we assume that only one subclass in the MRO chain
+    # pyre: will every call `set_metadata`, so it's okay for our purposes. There's no
```
jimmylai: The `every` seems redundant?
```suggestion
    # pyre: will call `set_metadata`, so it's okay for our purposes. There's no
```

PR #102, opened 2019-10-11
Immutability and Hash/Equality for Matchers

DragonMinded:
## Summary

During the discussion on the RFC for matchers + metadata we ran into issues with matchers as they're currently defined. This PR shores up potential immutability issues with special matchers and defines an explicit `__hash__` and `__eq__` for matchers (both concrete and special). Given the fact that dataclasses become fairly useless if you also override their `__init__`, I opted to manually provide immutability via properties on special matchers.

## Test Plan

Wrote a new test that failed before these fixes. Pyre and tox for normal stuff.

### DIFF
--- a/libcst/codegen/gen_matcher_classes.py
+++ b/libcst/codegen/gen_matcher_classes.py
@@ -511,7 +511,7 @@ for node in _get_nodes():
 
     generated_code.append("")
     generated_code.append("")
-    generated_code.append("@dataclass(frozen=True)")
+    generated_code.append("@dataclass(frozen=True, eq=False, unsafe_hash=False)")
     generated_code.append(f'class {node.name}({", ".join(classes)}):')
     all_exports.add(node.name)
 
--- a/libcst/matchers/_matcher_base.py
+++ b/libcst/matchers/_matcher_base.py
@@ -133,7 +133,16 @@ class OneOf(Generic[_MatcherT], BaseMatcherNode):
 
         if len(actual_options) < 2:
             raise Exception("Must provide at least two options to OneOf!")
-        self.options: Sequence[_MatcherT] = actual_options
+        self._options: Sequence[_MatcherT] = tuple(actual_options)
+
+    @property
+    def options(self) -> Sequence[_MatcherT]:
+        """
+        The normalized list of options that we can choose from to satisfy a
+        :class:`OneOf` matcher. If any of these matchers are true, the
+        :class:`OneOf` matcher will also be considered a match.
+        """
+        return self._options
 
     def __or__(self, other: _OtherNodeT) -> "OneOf[Union[_MatcherT, _OtherNodeT]]":
         # Without a cast, pyre thinks that the below OneOf is type OneOf[object]
@@ -145,10 +154,10 @@ class OneOf(Generic[_MatcherT], BaseMatcherNode):
 
     def __invert__(self) -> "AllOf[_MatcherT]":
         # Invert using De Morgan's Law so we don't have to complicate types.
-        return cast(AllOf[_MatcherT], AllOf(*[DoesNotMatch(m) for m in self.options]))
+        return cast(AllOf[_MatcherT], AllOf(*[DoesNotMatch(m) for m in self._options]))
 
     def __repr__(self) -> str:
-        return f"OneOf({', '.join([repr(o) for o in self.options])})"
+        return f"OneOf({', '.join([repr(o) for o in self._options])})"
 
 
 class AllOf(Generic[_MatcherT], BaseMatcherNode):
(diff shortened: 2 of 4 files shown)


libcst/matchers/_matcher_base.py
```diff
@@ -76,6 +76,18 @@ def __and__(
...
+        Matchers are only treated as equal by identity. This matches the behavior of
+        CPython's AST nodes and LibCST nodes. Comparing matchers by value is useless
+        as they exist only as a way to programatically describe the shape of a tree.
+        """
+        return self is other
+
+    def __hash__(self) -> int:
+        # Equality of nodes is based on identity, so the hash should be too.
+        return id(self)
+
```
carljm: I think this is the default behavior of `__eq__` and `__hash__` (inherited from `object`) for any Python class that doesn't define them explicitly. So I think it's sufficient for this diff to just turn off the addition of those methods by `@dataclass`; we don't need to define methods that implement the default behavior.

libcst/matchers/_matcher_base.py
```diff
@@ -133,7 +145,16 @@ def __init__(self, *options: Union[_MatcherT, "OneOf[_MatcherT]"]) -> None:
 
         if len(actual_options) < 2:
             raise Exception("Must provide at least two options to OneOf!")
-        self.options: Sequence[_MatcherT] = actual_options
+        self._options: Sequence[_MatcherT] = tuple(actual_options)
+
+    @property
+    def options(self) -> Sequence[_MatcherT]:
```
carljm: This is fine with me, but tbh it's not really needed for this diff; if we hash only by identity, not by value, it no longer matters whether the instance attributes are mutable; the only thing that must be immutable is the identity, which is by definition immutable.

libcst/matchers/_matcher_base.py
DragonMinded: Oh, good call. I'll try that.

libcst/matchers/_matcher_base.py
DragonMinded: I still like the idea of providing an Exception when somebody tries to update a constructed matcher. We already do that for the generated classes, so I'll elect to keep this for special matchers to make sure there's no gap.

libcst/matchers/_matcher_base.py
DragonMinded: Yup, worked fine. Glad I wrote that test ;)

libcst/codegen/gen_matcher_classes.py
```diff
@@ -511,7 +511,7 @@ def _get_fields(node: Type[cst.CSTNode]) -> Generator[Field, None, None]:
 
     generated_code.append("")
     generated_code.append("")
-    generated_code.append("@dataclass(frozen=True)")
+    generated_code.append("@dataclass(frozen=True, eq=False, unsafe_hash=False)")
```
jimmylai: The default value of `unsafe_hash` is False, we can just skip it.

libcst/matchers/tests/test_decorators.py
```diff
@@ -968,3 +968,36 @@ def bar() -> None:
...
         self.assertEqual(visitor.visits, ['"foo"', '"bar"', '"foobar"'])
         self.assertEqual(visitor.leaves, ['"foo"', '"bar"'])
+
+    def test_init_with_unhashable_types(self) -> None:
+        # Set up a simple visitor with a call_if_inside decorator.
+        class TestVisitor(MatcherDecoratableTransformer):
+            def __init__(self) -> None:
+                super().__init__()
+                self.visits: List[str] = []
+                self.leaves: List[str] = []
```
jimmylai: Unused variable.
```suggestion
```

libcst/codegen/gen_matcher_classes.py
DragonMinded: No, it is true when `frozen=True`, hence why there was a hash generated for these classes that cause the problem this diff is addressing.

libcst/matchers/tests/test_decorators.py
DragonMinded: Fixed.

libcst/codegen/gen_matcher_classes.py
jimmylai: >>> If eq and frozen are both true, by default dataclass() will generate a __hash__() method for you. If eq is true and frozen is false, __hash__() will be set to None, marking it unhashable (which it is, since it is mutable). If eq is false, __hash__() will be left untouched meaning the __hash__() method of the superclass will be used (if the superclass is object, this means it will fall back to id-based hashing).
https://docs.python.org/3/library/dataclasses.html

Based on CPython document, when `frozen=True and eq=False`, the __hash__ will left untouched and fall back to id-based when superclass is object, which seems apply to our case?

jimmylai: I verified it and the document is correct that when frozen and eq are True, setting unsafe_hash=False or not won't change the behavior of the hash function and it's the identity.

DragonMinded: Without these changes, we can't hash list and the provided unit test fails. Perhaps one of these can be dropped, but it seems foolish to rely on defaults when we want to explicitly dictate behavior. Especially given how unintelligible dataclass docs are with respect to what values are defaulted given other values.

PR #105, opened 2019-10-11
Generalize codegen cleanup step and apply it to all codegen

DragonMinded:
##  Summary

I noticed that the typed visitors codegen was creating messy types such as `Union[SingleType]`. We have a clean-up for `Union[SingleType]` snuck into `gen_matcher_classes`. So, lets remove that snuck-in clean-up from `gen_matcher_classes` and apply it globally to all codegen. This makes its purpose a lot more obvious, while helping decouple necessary codegen from prettifying/simplifying transforms. It also cleans up typed visitors, so that's a bonus.

## Test Plan

tox

### DIFF
--- a/libcst/codegen/gen_matcher_classes.py
+++ b/libcst/codegen/gen_matcher_classes.py
@@ -105,16 +105,6 @@ class CleanseFullTypeNames(cst.CSTTransformer):
         return updated_node.with_changes(comma=cst.MaybeSentinel.DEFAULT)
 
 
-class DoubleQuoteStrings(cst.CSTTransformer):
-    def leave_SimpleString(
-        self, original_node: cst.SimpleString, updated_node: cst.SimpleString
-    ) -> cst.SimpleString:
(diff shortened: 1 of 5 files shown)


libcst/codegen/gen_matcher_classes.py
```diff
@@ -128,17 +128,6 @@ def leave_ExtSlice(
                     return cst.RemoveFromParent()
         return updated_node
 
-    def leave_Subscript(
```
jimmylai: LGTM, other than `leave_Subscript`, I found some other good stuff in libcst/codegen/gen_matcher_classes.py can be moved to libcst/codegen/transforms.py for reusability. E.g. convert ForwardRef as SimpleString.

DragonMinded: I will move `DoubleQuoteStrings` as well but the ForwardRef handling is entirely related to matchers because of how we codegen using LibCST.

PR #106, opened 2019-10-11
Use import libcst as cst and explicit keyword arg for consistency and readability

jimmylai:
## Summary
As discussed offline, use tuple for Sequence may cause confusion when nested with many other parentheses.

This PR was generated using those two transformers.
```
module = cst.parse_module(src)

class CallTransformer(cst.CSTTransformer):
    def leave_Call(self, o, u):
        if m.matches(
            u,
            m.Call(
                func=m.Attribute(
                    value=m.Name("cst") | m.Name("m"), attr=m.Name("Call")
                ),
                args=[m.AtLeastN(n=2)],
            ),
        ):
            return u.with_changes(
                args=[
                    cst.Arg(keyword=cst.Name("func"), value=u.args[0].value),
                    cst.Arg(keyword=cst.Name("args"), value=u.args[1].value),
                ]
            )
        return u

class CSTImportTransformer(cst.CSTTransformer):
    def leave_Attribute(self, o, u):
        if m.matches(u, m.Attribute(value=m.Name("libcst"))):
            return u.with_changes(value=cst.Name("cst"))
        return u
```

### DIFF
--- a/libcst/matchers/tests/test_matchers.py
+++ b/libcst/matchers/tests/test_matchers.py
@@ -4,7 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 # pyre-strict
-import libcst
+import libcst as cst
 import libcst.matchers as m
 from libcst.matchers import matches
 from libcst.testing.utils import UnitTest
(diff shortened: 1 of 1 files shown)


jimmylai:
> b

Since I ready have change ready, just remove the tuple to list change on this PR. Will update the doc to use list over tuple in another PR.

PR #111, opened 2019-10-15
Throw on bad usage of @visit/@leave

DragonMinded:
## Summary

It is currently possible to add a `@visit` or `@leave` decorator to a `visit_<Node>` or `leave_<Node>` method on a visitor. This does exactly what you would expect: it adds a visitor that calls this function when the matcher in the decorator is true. However, its probably surprising to people that we don't deactivate the original visitor (as defined by the fact that it is a `visit_<Node>` or `leave_<Node>` function). So the end result is that we call the function twice. Once for every visit to `<Node>` and once again every time the matcher is true. While technically correct, this behavior is useless since `visit_<Node>` will *always* call a strict superset of nodes as compared to `@visit(m.<Node>)`.

## Test Plan

tox, changed a test to use the wrong function name and saw an exception.


DragonMinded:
I took the liberty of doing some refactoring and fixing a TODO in order to lay the groundwork for this diff. It makes the actual change simpler and cleans up some other code, so it seems like a win-win.

libcst/matchers/_visitors.py
```diff
@@ -58,11 +65,11 @@ def _get_possible_match_classes(matcher: BaseMatcherNode) -> List[Type[cst.CSTNo
         return [getattr(cst, matcher.__class__.__name__)]
 
 
-def _get_possible_annotated_classes(annotation: Type[object]) -> List[Type[object]]:
```
DragonMinded: This needed a change since it is technically a lie. This function takes either a `Type[object]` or a `Union[Type[Object], ...]`. This was exposed by adding the above `TYPED_FUNCTION_RETURN_MAPPING`.

libcst/matchers/_visitors.py
```diff
@@ -58,11 +65,11 @@ def _get_possible_match_classes(matcher: BaseMatcherNode) -> List[Type[cst.CSTNo
...
+def _get_possible_annotated_classes(annotation: object) -> List[Type[object]]:
     if getattr(annotation, "__origin__", None) is Union:
         return getattr(annotation, "__args__", [])
     else:
-        return [annotation]
+        return [cast(Type[object], annotation)]
```
DragonMinded: We could add runtime assertions that there isn't a nested Union[Union[...]] or anything else weird, but this seems like overkill for a best-effort runtime validation.

carljm:
Lint/pyre not happy?

libcst/codegen/gather.py
```diff
@@ -0,0 +1,141 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
jimmylai: github is not smart enough to mark `libcst/codegen/gather.py` and `libcst/codegen/gen_type_mapping.py` as moved changes, that can make code review much easier.

libcst/matchers/_visitors.py
```diff
@@ -296,13 +310,20 @@ def _gather_constructed_leave_funcs(
 
     for funcname in dir(obj):
         try:
-            func = cast(Callable[[cst.CSTNode], None], getattr(obj, funcname))
+            possible_func = getattr(obj, funcname)
+            if not ismethod(possible_func):
+                continue
+            func = cast(Callable[[cst.CSTNode], None], possible_func)
         except Exception:
```
jimmylai: After adding the `not ismethod` check, do we still need this `except Exception:`?

libcst/matchers/_visitors.py
```diff
@@ -75,20 +82,9 @@ def _get_valid_leave_annotations_for_classes(
...
-            **{x: getattr(cst, x) for x in dir(cst)},
-            "cst": cst,
-            "libcst": cst,
-        }
-        type_hints = get_type_hints(meth, namespace, namespace)
-        if "return" in type_hints:
-            retval.update(_get_possible_annotated_classes(type_hints["return"]))
+        retval.update(
+            _get_possible_annotated_classes(TYPED_FUNCTION_RETURN_MAPPING[cls])
+        )
```
jimmylai: Will this change the result of retval? Or it's just a refactoring.

libcst/matchers/_visitors.py
DragonMinded: Yes, I think so. The exception isn't for the type, it is for if we accidentally do work on getattr due to properties.

libcst/matchers/_visitors.py
DragonMinded: It is a pure refactor! It changes absolutely nothing (I verified with all tests that it is identical).

Issue #101, opened 2019-10-10
Improve regularity of Subscript slice


carljm:
As in several places in Python's syntax, commas in subscripts are a bit overloaded. Do they simply form a tuple, or are they a separator in a distinct special form, as they are in e.g. function calls?

In the runtime, Python tries really hard to pretend that they simply form a tuple: `foo[1, 2, 3]` and `foo[(1, 2, 3)]` are indistinguishable at runtime (in both cases the `__getitem__` of `foo`'s class receives the tuple `(1, 2, 3)`); even `foo[1, 2, 3:4]` just sends the tuple `(1, 2, slice(3, 4, None))` to `__getitem__`.

But there's a catch: the literal syntax for slices (e.g. `3:4`) is not valid anywhere except directly inside a subscript. It is not valid in a tuple. So while `foo[1, 2, 3:4]` is valid syntax, `foo[(1, 2, 3:4)]` is not, giving the lie to the idea that these forms should be fully equivalent. (What that syntax _means_ is a different question entirely; it's not used in the standard library, but it is used for e.g. slicing multi-dimensional numpy arrays.)

This leaves some gray area in terms of how a syntax tree should represent these forms. It seems that `foo[1, 2, 3]` should be a simple tuple index. And in fact, Python's AST does represent it that way! But `foo[1, 2, 3:4]` cannot be represented that way without weakening the tuple node to allow it to contain a literal slice, which in general it cannot. So the AST invents a special ExtSlice node which is effectively much like a tuple that can contain slices, and that occurs in the AST only when a subscript contains multiple comma-separated values, at least one of which is a slice.

LibCST handles it similarly, but slightly differently. Today in LibCST the `slice` attribute of a `Subscript` node can be one of three types:

1) An `Index` node containing  any arbitrary expression (e.g. the `1` in `foo[1]`). This is like the AST.
2) A `Slice` node representing a slice such as `1:` or `2:4` or `2:8:2`. This is also like the AST.
3) A `Sequence` of `ExtSlice` nodes, each of which has an optional trailing comma and which itself has a value that is either an `Index` or a `Slice`. This is used for cases like `foo[1, 2, 3:4]`, but also for simple `foo[1, 2, 3]`. Unlike the AST, LibCST uses `ExtSlice` anytime the subscript has commas at the top level, not only when one of the elements is a slice.

[...]

PR #113, opened 2019-10-16
Rename ExtSlice

DragonMinded:
## Summary

This is part one of a plan to deprecate `ExtSlice` and fix the irregularity of `Subscript.slice` which will address #101. In this diff I simply rename `ExtSlice` to `SubscriptElement`, providing a backwards-compatible shim for existing surfaces using or visiting over `ExtSlice`. This is somewhat complicated by the fact that we need to not just allow construction of nodes/matchers using `ExtSlice`, but we also need to be able to call `visit_ExtSlice` and `leave_ExtSlice` (as well as attribute visitors) on old visitors even though the new node is named `SubscriptElement`. The construction/instance check/matching side of things will work with only the `ExtSlice = SubscriptElement` alias since internally we refer to everything as `SubscriptElement`, but for string-based function lookup, we need to get a little more clever and make the default  visit_SubscriptElement` delegate onward to `visit_ExtSlice` so that either form works.

## Test Plan

tox, write a test to exercise using `ExtSlice` in construction, visiting and matchers.


libcst/_parser/conversions/expression.py
```diff
@@ -694,15 +694,15 @@ def convert_subscriptlist(
     config: ParserConfig, children: typing.Sequence[typing.Any]
 ) -> typing.Any:
     if len(children) > 1:
-        # This is a list of ExtSlice, so construct as such by grouping every
+        # This is a list of SubscriptElement, so construct as such by grouping every
         # subscript with an optional comma and adding to a list.
         extslices = []
```
carljm: nit: rename variable to `elements`?

DragonMinded: Good catch!

DragonMinded:
Yeah, renaming is problematic precisely because of the visitor pattern. `@visit`/`@leave` fixes that as you see here, but they're inherently slower and slightly less ergonomic for general cases. But, its possible!

libcst/_typed_visitor.py
```diff
@@ -4247,6 +4229,41 @@ def visit_Subscript_whitespace_after_value(self, node: "Subscript") -> None:
...
+    def leave_ExtSlice_slice(self, node: "ExtSlice") -> None:
+        pass
+
+    @mark_no_op
+    def visit_ExtSlice_comma(self, node: "ExtSlice") -> None:
+        pass
+
+    @mark_no_op
+    def leave_ExtSlice_comma(self, node: "ExtSlice") -> None:
+        pass
```
jimmylai: Can we call `warnings.warn(..., DeprecationWarning)` in those deprecated visit and leave function to provide warning to users?

jimmylai: After rethink about it, for showing the warning, we'll need some work on at_visit/at_leave which doesn't seem worth it.

PR #115, opened 2019-10-17
Remove parsing Subscript slice directly into Index/Slice

DragonMinded:
## Summary

This makes sure we always wrap elements in a SubscriptElement, even when there
is only one element. This makes things more regular while still being backwards
compatible with existing creation. The meat of this is in two halves, which can't
be split due to not wanting to break the build between commits. The first half
is just the changes to the parser and updates to tests. This includes a test to
be sure we can still render code that uses old construction types. The second half
is changes to codegen which made assumptions about `Subscript` and demonstrates
the need to make this change in the first place. This includes a fix to
`CSTNode.with_deep_changes` type to make it more correct and also more usable in
transforms without additional type assertions.

## Test Plan

Existing tests, added another deprecation test.

### DIFF
--- a/libcst/_nodes/base.py
+++ b/libcst/_nodes/base.py
@@ -402,7 +402,7 @@ class CSTNode(ABC):
 
     def with_deep_changes(
         self: _CSTNodeSelfT, old_node: "CSTNode", **changes: Any
-    ) -> Union[_CSTNodeSelfT, "CSTNode"]:
+    ) -> _CSTNodeSelfT:
         """
         A convenience method for applying :attr:`with_changes` to a child node. Use
         this to avoid chains of :attr:`with_changes` or combinations of
--- a/libcst/_nodes/tests/test_subscript.py
+++ b/libcst/_nodes/tests/test_subscript.py
@@ -17,7 +17,10 @@ class SubscriptTest(CSTNodeTest):
         (
             # Simple subscript expression
             (
-                cst.Subscript(cst.Name("foo"), cst.Index(cst.Integer("5"))),
+                cst.Subscript(
+                    cst.Name("foo"),
+                    (cst.SubscriptElement(cst.Index(cst.Integer("5"))),),
+                ),
                 "foo[5]",
                 True,
             ),
@@ -25,10 +28,14 @@ class SubscriptTest(CSTNodeTest):
             (
                 cst.Subscript(
                     cst.Name("foo"),
-                    cst.Slice(
-                        lower=cst.Integer("1"),
-                        upper=cst.Integer("2"),
-                        step=cst.Integer("3"),
+                    (
+                        cst.SubscriptElement(
+                            cst.Slice(
+                                lower=cst.Integer("1"),
+                                upper=cst.Integer("2"),
+                                step=cst.Integer("3"),
+                            )
+                        ),
                     ),
                 ),
                 "foo[1:2:3]",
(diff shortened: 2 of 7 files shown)


libcst/_nodes/base.py
```diff
@@ -401,8 +401,8 @@ def deep_remove(
         )
 
     def with_deep_changes(
-        self: _CSTNodeSelfT, old_node: "CSTNode", **changes: Any
-    ) -> Union[_CSTNodeSelfT, "CSTNode"]:
+        self: _CSTNodeSelfT, old_node: CSTNodeT, **changes: Any
+    ) -> Union[_CSTNodeSelfT, CSTNodeT]:
```
jimmylai: The type annotations here are confusing to me.
`_CSTNodeSelfT` and `CSTNodeT` seems refer to same thing, why do we need two names instead of one.
Also, while you're on typing here, can we also fix this inconsistency:
https://github.com/Instagram/LibCST/blob/a4124c91016a3430a6ce8bee0be0059989df8859/libcst/_nodes/base.py#L375-L378

libcst/codegen/gen_matcher_classes.py
```diff
@@ -213,26 +217,36 @@ def leave_Subscript(
...
                     # Special case for Sequence[Union] so that we make more collapsed
                     # types.
                     if possibleunion.value.deep_equals(cst.Name("Union")):
-                        return updated_node.with_changes(
-                            slice=nodeslice.with_changes(
-                                value=possibleunion.with_changes(
-                                    slice=[*possibleunion.slice, _get_do_not_care()]
-                                )
-                            )
+                        return updated_node.with_deep_changes(
```
jimmylai: Good use case of `with_deep_changes`!

libcst/_nodes/tests/test_subscript.py
```diff
@@ -17,18 +17,25 @@ class SubscriptTest(CSTNodeTest):
...
-                        upper=cst.Integer("2"),
-                        step=cst.Integer("3"),
+                    (
+                        cst.SubscriptElement(
+                            cst.Slice(
+                                lower=cst.Integer("1"),
+                                upper=cst.Integer("2"),
+                                step=cst.Integer("3"),
+                            )
+                        ),
```
jimmylai: Just curious, do you make those changes with codemod?

DragonMinded: No :( Since the codemod portion of libcst isn't open-source yet, it was easier to do them by hand.

libcst/_nodes/base.py
DragonMinded: There is no typing inconsistency with `deep_replace`. We care about the type of `self`, and the type we are going to replace `old_node` with, because those are the two possible types that we might get back from the function. `old_node` just needs to be any `CSTNode`, but it doesn't factor into the return. Either we replace a child, and the return is the type of `self`, or `old_node` == `self`, so our return is whatever type `new_node` is, hence the union.

I think this current fix is actually broken and I unnecessarily widened the type. `with_changes` guarantees the same node as output, so even if `old_node` == `self`, the output is still `_CSTNodeSelfT`. I can make a fix to that here and change back `old_node` to be of type `CSTNode` as it was before, and fix the return to not be a union but instead just be `_CSTNodeSelfT`

jimmylai: Yeah, return `_CSTNodeSelfT ` makes more sense to me.

PR #117, opened 2019-10-17
[metadata] make QualifiedNameProvider typing more specific

jimmylai:
## Summary
Optional[Collection[QualifiedName]] -> Collection[QualifiedName]
## Test Plan
pyre


bgw:
Seems like an improvement, LGTM.

PR #125, opened 2019-10-22
Add a best practices section with various recommended practices

DragonMinded:
## Summary

There are going to be many more where this comes from, but these are the three cases that have come up most frequently, so I started with these. Hopefully this helps give additional direction to people using LibCST.

## Test Plan

`tox -e docs`

### DIFF
--- a/docs/source/best_practices.rst
+++ b/docs/source/best_practices.rst
@@ -1,0 +1,192 @@
+==============
+Best Practices
+==============
+
+While there are plenty of ways to interact with LibCST, we recommend some patterns
+over others. Various best practices are laid out here along with their justifications.
+
+Avoid ``isinstance`` when traversing
+------------------------------------
+
+Excessive use of ``isinstance`` implies that you should rewrite your check as a
+matcher or unroll it into a set of visitor methods. Often, you should make use of
+:func:`~libcst.ensure_type` to make your type checker aware of a node's type.
+
+.. autofunction:: libcst.ensure_type
+
+Often it is far easier to use :ref:`libcst-matchers` over explicit instance checks
+in a transform. Matching against some pattern and then extracting a value from a
+node's child is often easier and far more readable. Unfortunately this clashes with
+various type-checkers which do not understand that :func:`~libcst.matchers.matches`
+guarantees a particular set of children. Instead of instance checks, you should use
+:func:`~libcst.ensure_type` which can be inlined and nested.
+
+For example, if you have written the following::
+
+    def get_identifier_name(node: cst.CSTNode) -> Optional[str]:
+        if m.matches(node, m.Name()):
+            assert isinstance(node, cst.Name)
+            return node.value
+        return None
+
+You could instead write something like::
+
+    def get_identifier_name(node: cst.CSTNode) -> Optional[str]:
+        return (
+            cst.ensure_type(node, cst.Name).value
+            if m.matches(node, m.Name())
+            else None
+        )
+
+If you find yourself attempting to manually traverse a tree using ``isinstance``,
+you can often rewrite your code using visitor methods instead. Nested instance checks
+can often be unrolled into visitors methods along with matcher decorators. This may
+entail adding additional state to your visitor, but the resulting code is far more
+likely to work after changes to LibCST itself. For example, if you have written the
+following::
+
+    class CountBazFoobarArgs(cst.CSTVisitor):
+        """
+        Given a set of function names, count how many arguments to those function
+        calls are the identifiers "baz" or "foobar".
+        """
+
+        def __init__(self, functions: Set[str]) -> None:
+            super().__init__()
+            self.functions: Set[str] = functions
+            self.arg_count: int = 0
+
+        def visit_Call(self, node: cst.Call) -> None:
+            # See if the call itself is one of our functions we care about
+            if isinstance(node.func, cst.Name) and node.func.value in self.functions:
+                # Loop through each argument
+                for arg in node.args:
+                    # See if the argument is an identifier matching what we want to count
+                    if isinstance(arg.value, cst.Name) and arg.value.value in {"baz", "foobar"}:
+                        self.arg_count += 1
+                        
+You could instead write something like::
+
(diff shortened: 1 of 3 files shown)


docs/source/best_practices.rst
```diff
@@ -0,0 +1,187 @@
...
+            # Pop the latest call off the stack
+            if m.matches(node.func, m.Name()):
+                self.call_stack.pop()
+
+        @m.visit(m.Arg(m.Name("baz") | m.Name("foobar")))
+        def _count_args(self, node: cst.Arg) -> None:
+            # See if the most shallow call is one we're interested in, so we can
+            # count the args we care about only in calls we care about.
+            if self.call_stack[-1] in self.functions:
+                self.arg_count += 1
```
jimmylai: This example collected call stack and that made it complex.
Looks like we can do it in a much simpler way with a matcher check?
```
def visit_Call(self, node: cst.Call) -> None:
    if m.matches(node, m.Call(func=Name(), value=m.OneOf(*[m.Name(s) for s in self. functions]))):
        self.arg_count += [m.matches(arg.value, m.OneOf(m.Name("baz"), m.Name("foobar"))) for arg in node.args]
```

DragonMinded: The point of this example is showing how to unroll a set of instance checks into more than one visitor method. Your proposed solution does get that point across.

jimmylai: I understand the point but the new code doesn't seem simpler as it claims to be.

In old code, all logic were inside visit_Call, which is easy to follow. In new code, the reader needs to figure out the orders and consequences of those functions (visit_Call, leave_Call and _count_arg), which seems harder to read.
I hope the example can present the `simpler` better.

docs/source/best_practices.rst
```diff
@@ -0,0 +1,187 @@
...
+
+While there is more code than the previous example, it is arguably easier to understand
+and maintain each piece of the code. Note that LibCST is traversing the tree completely
+in both cases, so while the first appears to be faster, it is actually doing the same
+amount of work as the second.
+
+Prefer ``updated_node`` when modifying trees
+--------------------------------------------
+
+When you are using :ref:`libcst-visitors` to modify a LibCST tree, only return
```
jimmylai: Can we make it more explicit using `Transformer`? Only Transformer can return modifications in `leave_<Node>`.

docs/source/best_practices.rst
```diff
@@ -0,0 +1,187 @@
...
+
+Running the recommended transform will return us a new code snippet that looks like this::
+
+    renamed_some_func(1, 2, renamed_other_func(3))
+
+However, running the version which modifies ``original_node`` will instead return::
+
+    renamed_some_func(1, 2, other_func(3))
+
+That's because the ``updated_node`` tree contains the modification to ``other_func``.
```
jimmylai: Good example explains the caveat.

docs/source/best_practices.rst
```diff
@@ -0,0 +1,187 @@
...
+By returning modifications to ``original_node`` instead of ``updated_node``, we accidentally
+discarded all the work done deeper in the tree.
+
+Provide a ``config`` when generating code from templates
+--------------------------------------------------------
+
+When generating complex trees it is often far easier to pass a f-string to
```
jimmylai: Why it needs to be a f-string?

docs/source/best_practices.rst
```diff
@@ -0,0 +1,187 @@
...
+            *module.body,
+            cst.parse_statement(
+                "print('Hello, world!')",
+                config=module.config_for_parsing,
+            ),
+        ),
+    )
+    new_code_string = new_module.code
+
+Leaving out the ``config`` parameter means that LibCST will assume some defaults
```
jimmylai: Is it possible to have a simple example to demonstrate the caveat when user don't pass config?

docs/source/best_practices.rst
DragonMinded: I completely disagree. Your proposed example uses a lot of Python features to make the least amount of code, but the density of required reading is high. I have trouble following it and I only can tell what it does because it proposes replacing other code with the same effect. It is also broken (you can't add a list to an int). It also continues to do manual traversal of the tree, which the recommended practice says to avoid. It also doesn't help demonstrate how somebody might spread logic across multiple visitors using state, which is precisely one motivation for moving to visitors instead of matchers/deep_replace. Basically, it is not only not a good example for the practice discussed on the page, but I would not view it is a good practice in general.

docs/source/best_practices.rst
DragonMinded: Good idea.

docs/source/best_practices.rst
DragonMinded: Because it primes the idea of using the various parsers to parse templated code.

docs/source/best_practices.rst
DragonMinded: Not really. Any simpler examples don't have the caveat of parsing subtly incorrectly.

docs/source/best_practices.rst
jimmylai: I'm not saying we should use my example. Just to see if there is a better example can demonstrate it's a lot simpler to use visitor functions.

jimmylai: To make it clear, my second comment wasn't proposing to use the example in my first comment. I tried to propose simplifying your example by comparing your before and after codes to demonstrate the after codes wasn't simpler. Sorry about the confusion if it wasn't clear enough.
For `visit_<CSTNode>()` function with consequences and orders tracked by class attributes (e.g. the stack variable or just a boolean in some other use cases), it's usually less readable because user need to figure out the orders and the consequences. I think the matcher decorator pattern like `call_if_inside` + `visit_<CSTNode>` is a better way to handle it. It keep the order of logic and consequences in the same place which is a lot more readable.

```
@call_if_inside(m.Call(func=m.OneOf(*[m.Name(s) for s in self.functions])))
def visit_Arg(node: cst.Arg):
    if m.matches(node.value, m.OneOf(m.Name("baz"), m.Name("foobar"))):
        self.count += 1
```

PR #126, opened 2019-10-22
Make the codegen enter/leave tracking more generic

bgw:
## Summary

I need to do some additional work on visit/leave to make codegen
re-entrant, so this makes it more generic.

This should have an additional small positive effect of creating less
throwaway objects when we're doing codegen without position calculation.

## Test Plan

pyre/unit tests/lint

### DIFF
--- a/libcst/_nodes/internal.py
+++ b/libcst/_nodes/internal.py
@@ -11,7 +11,6 @@ from typing import TYPE_CHECKING, Iterable, Iterator, List, Optional, Sequence, 
 
 from libcst._add_slots import add_slots
 from libcst._maybe_sentinel import MaybeSentinel
-from libcst._position import CodeRange
 from libcst._removal_sentinel import RemovalSentinel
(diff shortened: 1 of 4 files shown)


libcst/_nodes/internal.py
```diff
@@ -48,7 +47,10 @@ def add_indent_tokens(self) -> None:
     def add_token(self, value: str) -> None:
         self.tokens.append(value)
 
-    def record_position(self, node: "CSTNode", position: CodeRange) -> None:
+    def before_visit(self, node: "CSTNode") -> None:
+        pass
+
+    def after_leave(self, node: "CSTNode") -> None:
```
DragonMinded: I'm worried that people will confuse these with the various `visit_X` and `leave_X` functions. Should we name them something like `before_codgen` and `after_codegen` instead?

jimmylai:
LGTM

PR #128, opened 2019-10-23
Fix deep_clone for sequence types

DragonMinded:
## Summary

It turns out that when we had a sequence type, like Module's body, we were incorrectly
leaning on copy.deepcopy for the rest of the tree. This works, but its likely to blow
the stack when doing a deep clone of a very very large file. Fix that oversight here.

## Test Plan

`tox`

### DIFF
--- a/libcst/_nodes/base.py
+++ b/libcst/_nodes/base.py
@@ -110,6 +110,19 @@ def _indent(value: str) -> str:
     return "\n".join(f"    {l}" for l in value.split("\n"))
 
 
+def _clone(val: object) -> object:
+    # We can't use isinstance(val, CSTNode) here due to poor performance
+    # of isinstance checks against ABC direct subclasses. What we're trying
+    # to do here is recursively call this functionality on subclasses, but
+    # if the attribute isn't a CSTNode, fall back to copy.deepcopy.
+    try:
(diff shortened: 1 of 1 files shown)


libcst/_nodes/base.py
```diff
@@ -349,14 +362,12 @@ def deep_clone(self: _CSTNodeSelfT) -> _CSTNodeSelfT:
...
-            try:
-                cloned_fields[key] = val.deep_clone()
-            except AttributeError:
-                cloned_fields[key] = deepcopy(val)
+            if isinstance(val, (str, bytes)):
+                cloned_fields[key] = _clone(val)
+            elif isinstance(val, Sequence):
+                cloned_fields[key] = tuple(_clone(v) for v in val)
+            else:
+                cloned_fields[key] = _clone(val)
```
jimmylai: Isn't those isinstance check slow?

Can we do some micro benchmark to see how fast it is like using `%timeit` to clone a large tree?

DragonMinded: Experimentally, on a massive file on internal code in IG, this sped up cloning by a significant amount. However, I can take a similar approach where I attempt to iterate and if it fails I will fall back in the `except` case.

PR #129, opened 2019-10-23
[scope] add node to LocalScope

jimmylai:
## Summary
add FunctionDef and ClassDef to FunctionScope and ClassScope
In a recent use case, the user wanted to get the node defines the scope and she had to do `scope.parent[scope.name]`. That returns a list of possible Assignments contain FucntionDef/ClassDef nodes.
If we just store the node in scope object, it's simple and accurate as `scope.node`
CC @bgw 
## Test Plan
Unit test.

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -526,7 +526,12 @@ class FunctionScope(LocalScope):
     When a function is defined, it creates a FunctionScope.
     """
 
-    pass
+    #: The :class:`~libcst.FunctionDef` node defines the current scope.
+    node: cst.FunctionDef
+
+    def __init__(self, parent: Scope, name: str, node: cst.FunctionDef) -> None:
+        super().__init__(parent, name)
+        self.node = node
 
 
 # even though we don't override the constructor.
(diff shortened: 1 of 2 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -682,7 +703,7 @@ def visit_FunctionDef(self, node: cst.FunctionDef) -> Optional[bool]:
         return False
 
     def visit_Lambda(self, node: cst.Lambda) -> Optional[bool]:
-        with self._new_scope(FunctionScope):
+        with self._new_scope(FunctionScope, name=None, node=node):
```
DragonMinded: We don't need `name=None` here.

libcst/metadata/scope_provider.py
```diff
@@ -584,10 +596,17 @@ def __init__(self, provider: "ScopeProvider") -> None:
...
-        self, kind: Type[LocalScope], name: Optional[str] = None
+        self,
+        kind: Type[LocalScope],
+        name: Optional[str] = None,
+        node: Optional[cst.CSTNode] = None,
     ) -> Iterator[None]:
         parent_scope = self.scope
-        self.scope = kind(parent_scope, name)
+        if node and kind in (FunctionScope, ClassScope):
+            # pyre-ignore: pyre don't know FunctionScope and ClassScope take node arg.
```
DragonMinded: It seems this pyre error is alluding to the fact that these now have different constructors. I think its better to have the node arg across all constructors and do nothing with it in those that don't care.

jimmylai: Add the node to all LocalScope.

PR #132, opened 2019-10-28
Add a reentrant (incremental) codegen API

bgw:
## Summary

**Context:** This is an experimental performance optimization that we're
hoping to use for our internal linter at Instagram. I added some
documentation, but it's unsupported, and isn't very user-friendly.

This adds `ExperimentalReentrantCodegenProvider`, which tracks the
codegen's internal state (indentation level, character offsets,
encoding, etc.) and for each statement, it stores a `CodegenPartial`
object.

The `CodegenPartial` object has enough information about the previous
codegen pass to run the codegen on part of a tree and patch the result
back into the original module's string.

In cases where we need to generate a bunch of small independent patches
for the same file (and we can't just generate a new tree with each patch
applied), this *should* be a faster alternative.

I don't have any performance numbers because I still need to test this
end-to-end with our internal codebase, but I'd be shocked if it was
slower than what we're doing.

This could theoretically live outside of LibCST, but it depends on a
whole bunch of LibCST internals, so there's some value in making sure
that this is in sync with the rest of LibCST.

## Test Plan

- Unit tests
- Pyre

Screenshot_2019-10-28 Experimental APIs — LibCST documentation

### DIFF
--- a/libcst/metadata/reentrant_codegen.py
+++ b/libcst/metadata/reentrant_codegen.py
@@ -56,7 +56,7 @@ class CodegenPartial:
         """
         return self.get_original_module_code().encode(self._prev_codegen_state.encoding)
 
-    def get_original_node_code(self) -> str:
+    def get_original_statement_code(self) -> str:
         """
         Equivalent to :meth:`libcst.Module.code_for_node` on the current statement,
         except that it uses the cached result from our previous code generation pass,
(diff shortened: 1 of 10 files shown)


libcst/metadata/reentrant_codegen.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def get_original_node_code(self) -> str:
+        """
+        Equivalent to :meth:`libcst.Module.code_for_node` on the current statement,
+        except that it uses the cached result from our previous code generation pass,
+        so it's faster.
+        """
+        return self._prev_codegen_state.get_code()[self.start_offset : self.end_offset]
+
+    def get_modified_node_code(self, node: BaseStatement) -> str:
```
DragonMinded: Should this be named `get_modified_statement_code` since it doesn't work on any node type?

bgw:
Sorry for all the force pushes, I realized that I was running an outdated version of black, causing my local machine to disagree with CircleCI, and had to wipe my `.tox` to fix it.

DragonMinded:
Yeah, I ran into the same issues when I upgraded black to support 3.8 (and I'm sure we'll get more now that there's a new official black release).

PR #137, opened 2019-10-30
Implement `findall` in matcher API.

DragonMinded:
## Summary

Findall does what you expect given a matcher and a tree: It returns all nodes
that exist in that tree which match the given matcher. For convenience, findall
works with regular LibCST trees as well as MetadataWrappers. It is also provided
as a helper method on the matcher transforms.

Note that this does not return an iterator. I based this decision on the fact that we
don't currently have a way to use visitors under the hood to create an iterator. If I
were to implement this myself, I would effectively be reinventing traversal of trees.
I don't think this is that big of a deal. If we need an efficient way to find the first
occurance of a match in a tree, I can implement `find` which bails early and document
it as such. For now, I don't want to overcomplicate the API, so I will wait until this
is needed to implement it.

## Test Plan

tox

### DIFF
--- a/libcst/matchers/_matcher_base.py
+++ b/libcst/matchers/_matcher_base.py
@@ -994,7 +994,7 @@ def _matches(
 def _construct_metadata_fetcher_null() -> Callable[
     [meta.ProviderT, libcst.CSTNode], object
 ]:
-    def _fetch() -> object:
+    def _fetch(*args: object, **kwargs: object) -> object:
         return _METADATA_MISSING_SENTINEL
 
     return _fetch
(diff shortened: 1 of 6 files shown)


libcst/matchers/_matcher_base.py
```diff
@@ -994,7 +994,7 @@ def _matches(
 def _construct_metadata_fetcher_null() -> Callable[
     [meta.ProviderT, libcst.CSTNode], object
 ]:
-    def _fetch() -> object:
+    def _fetch(*args: object, **kwargs: object) -> object:
```
DragonMinded: I apparently had no test for this and it failed when I tested it in `findall`. A regression for this is included in this PR.

libcst/matchers/tests/test_findall.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+        self.assertNodeSequenceEqual(booleans, [cst.Name("a"), cst.Name("b")])
+
+        # Test that failing to provide metadata leads to no match
+        booleans = findall(
+            wrapper.module,
+            m.MatchMetadata(
+                meta.ExpressionContextProvider, meta.ExpressionContext.STORE
+            ),
+        )
```
DragonMinded: This chunk right here tests the no metadata provider fix commented on above.

libcst/matchers/_visitors.py
```diff
@@ -658,3 +682,24 @@ def matches(
         function.
         """
         return matches(node, matcher, metadata_resolver=self)
+
+    def findall(
+        self,
+        tree: Union[cst.MaybeSentinel, cst.RemovalSentinel, cst.CSTNode],
```
jimmylai: Why support MaybeSentinel and RemovalSentinel? What's the use cases?

DragonMinded: As it says in the `findall` documentation: "Note that the node can also be a RemovalSentinel or a MaybeSentinel in order to use findall directly on transform results and node attributes."

libcst/matchers/tests/test_findall.py
```diff
@@ -0,0 +1,164 @@
...
+from typing import Optional, Sequence
+
+import libcst as cst
+import libcst.matchers as m
+import libcst.metadata as meta
+from libcst.matchers import findall
+from libcst.testing.utils import UnitTest
+
+
+class MatchersFindAllTest(UnitTest):
```
jimmylai: The test cases here test the deep_equals instead of identity. Can we also have some test cases to test the identity of nodes to make sure we get the nodes from the original tree?

DragonMinded: We already have tests for various pieces of the system, such as transform/visit. We shouldn't retest every part of the system for every new PR otherwise our tests will be unwieldy. We need to trust that a visitor that gathers indeed keeps nodes by identity as that's how its defined.

I don't think that adding even more testing here benefits us.

jimmylai: I was thinking a test case like this.

```
a = 1
b = a + 2
```
When findall with ExpressionContext == STORE, we expect to get the first `Name("a")` instead of second `Name("a")`.

DragonMinded: That's how `ExpressionContext` works along with traversal, and we have tests for each piece of the system.

PR #139, opened 2019-11-01
Implement a `MatchMetadataIfTrue` matcher

DragonMinded:
## Summary

This is especially helpful for checking qualified names of nodes against one item
or a list of items that you wish to match against. I chose to create a new matcher
instead of widening the type of `MatchMetadata` to take in either a value or
a callable. I was originally going to do the former, but having a MatchIfTrue
and a MatchMetadataIfTrue felt more orthogonal and became easier to explain than
a single MatchIfTrue that could take two types of values.

## Test Plan

tox

### DIFF
--- a/libcst/matchers/_matcher_base.py
+++ b/libcst/matchers/_matcher_base.py
@@ -321,10 +321,10 @@ class MatchIfTrue(Generic[_CallableT]):
     def __init__(self, func: _CallableT) -> None:
         # Without a cast, pyre thinks that self.func is not a function, even though
         # it recognizes that it is a _CallableT bound to Callable.
-        self._func: Callable[..., bool] = cast(Callable[..., bool], func)
+        self._func: Callable[[object], bool] = cast(Callable[[object], bool], func)
 
     @property
-    def func(self) -> Callable[..., bool]:
+    def func(self) -> Callable[[object], bool]:
         """
         The function that we will call with a LibCST node in order to determine
         if we match. If the function returns ``True`` then we consider ourselves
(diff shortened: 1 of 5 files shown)


DragonMinded:
Tests covering missing functionality added, and fixed up invert. This is ready for review.

jimmylai:
`MatchMetadata` needs to use with `metadata=` while `MatchMetadataIfTrue` can be used to match any node directly.
I'm worrying about the differences between `MatchIfTrue` and `MatchMetadataIfTrue` may be confusing to users.

DragonMinded:
`MatchMetadata` doesn't need to be used with `metadata=` at all. It can be used in any spot that `MatchIfTrue` can be used, same with `MatchMetadataIfTrue`. There are no restrictions for any of the nodes.

libcst/matchers/_matcher_base.py
```diff
@@ -446,7 +472,93 @@ def __invert__(self) -> "MatchMetadata":
...
+        m.Name(
+            value="foo",
+            metadata=m.MatchMetadataIfTrue(
+                meta.SyntacticPositionProvider,
+                lambda position: position.start.column == 0,
+            )
+        )
+
+    This can be used in place of any concrete matcher as long as it is not the
+    root matcher. Calling :func:`matches` directly on a :class:`MatchMetadataIfTrue`
```
jimmylai: What happens when it's used as a root matcher?

DragonMinded: It does still work in the base case, but I've intentionally narrowed the types on `matches` so that it is a mypy/pyre error. For @visit and @leave nodes, it will error out with the runtime type checking since you can't infer types.

PR #141, opened 2019-11-08
[metadata] add for loop support to ExpressionContextProvider

jimmylai:
## Summary
add for loop support to ExpressionContextProvider.
CC @mvismonte 

## Test Plan
New unit test.


DragonMinded:
LGTM, I assume you'll fix the tests before land.

PR #146, opened 2019-11-13
Fix not widening matcher sequence types inside unions

DragonMinded:
## Summary

Basically, I screwed up when I did the codegen refactor and didn't widen types for sequences properly. This re-allows sequence wildcards to be used as entries in a sequence, and fixes the `MatchIfTrue` stanza for sequence types.

## Test Plan

Pyre, tox, compare to internal codemods that require these widened types.


DragonMinded:
Once this gets reviewed, I can make another point release to fix the type issues.

PR #148, opened 2019-11-13
Implement an ``extract`` function

DragonMinded:
## Summary

Create an ``extract`` function which will extract child nodes/sequences/metadata given a successful match. This PR is a bit of a mess since I'm effectively using the matcher engine as an extract engine. So, I had to widen all the recursive return types and combine things at appropriate points in order to make this work.

## Test Plan

Pyre, a host of new tests to verify things, existing tests for the rest of matchers. Hopefully code coverage bot lets me know if/where I missed anything.


docs/source/matchers.rst
```diff
@@ -31,6 +31,7 @@ selectively control when LibCST calls visitor functions.
 
 .. autofunction:: libcst.matchers.matches
 .. autofunction:: libcst.matchers.findall
+.. autofunction:: libcst.matchers.extract
```
joshkehn: <3 RST

libcst/matchers/_matcher_base.py
```diff
@@ -293,6 +293,87 @@ def __repr__(self) -> str:
...
+        """
+        return self._name
+
+    def __or__(self, other: _OtherNodeT) -> "OneOf[Union[_MatcherT, _OtherNodeT]]":
+        # Without a cast, pyre thinks that the below OneOf is type OneOf[object]
+        # even though it has the types passed into it.
+        return cast(OneOf[Union[_MatcherT, _OtherNodeT]], OneOf(self, other))
+
+    def __and__(self, other: _OtherNodeT) -> "AllOf[Union[_MatcherT, _OtherNodeT]]":
+        # This doesn't mase sense. If we have multiple SaveMatchedNode captures
```
joshkehn:
```suggestion
        # This doesn't make sense. If we have multiple SaveMatchedNode captures
```

libcst/matchers/tests/test_extract.py
```diff
@@ -0,0 +1,411 @@
...
+            cst.ensure_type(
+                cst.ensure_type(expression, cst.Tuple).elements[1].value, cst.Call,
+            )
+            .args[2]
+            .value
+        )
+        self.assertEqual(nodes, {"arg": extracted_node})
+
+    def test_extract_sequence(self) -> None:
+        expression = cst.parse_expression("a + b[c], d(e, f * g, h.i.j)")
```
joshkehn: Probably my newness makes this hard to read, but if I'm understanding correctly the input/match/assert here is the same thing? Would it be reasonable to wrap a helper that allows specifying the input and output as an expression for a particular matcher? I suppose variable names would get in the way.

DragonMinded: Its pretty samey and I'll admit to copy-pastaing the tests a bit. I'm on the fence about improving their quality past the point of validating the featureset though.

PR #150, opened 2019-11-14
Remove length restrictions for AllOf/OneOf

DragonMinded:
## Summary

These checks are well-intentioned, but unfortunately make construction of a OneOf/AllOf from a sequence expansion a pain in the ass. So, drop them. The code already does the right thing, this was just providing helpful messages.

## Test Plan

All existing tests. Eyes.


DragonMinded:
Yep!

PR #154, opened 2019-11-19
Implement an `extractall` function

DragonMinded:
## Summary

This is equivalent to calling `findall` on a node, and then calling `extract` with the same
matcher on each of the returned nodes from `findall`.

## Test Plan

tox, added tests. This is a refactor of `findall` so I expect most edge cases to be covered already. I only tested the basic cases as a result.

### DIFF
--- a/libcst/matchers/tests/test_findall.py
+++ b/libcst/matchers/tests/test_findall.py
@@ -11,7 +11,7 @@ from typing import Optional, Sequence
 import libcst as cst
 import libcst.matchers as m
 import libcst.metadata as meta
-from libcst.matchers import findall
+from libcst.matchers import extractall, findall
 from libcst.testing.utils import UnitTest
 
 
@@ -162,3 +162,16 @@ class MatchersFindAllTest(UnitTest):
         visitor = TestTransformer()
         wrapper.visit(visitor)
         self.assertNodeSequenceEqual(visitor.results, [cst.Name("a"), cst.Name("b")])
+
+
+class MatchersExtractAllTest(UnitTest):
+    def test_extractall_simple(self) -> None:
+        expression = cst.parse_expression("a + b[c], d(e, f * g, h.i.j)")
+        matches = extractall(expression, m.Arg(m.SaveMatchedNode(~m.Name(), "expr")))
+        extracted_args = cst.ensure_type(
+            cst.ensure_type(expression, cst.Tuple).elements[1].value, cst.Call,
+        ).args
+        self.assertEqual(
+            matches,
+            [{"expr": extracted_args[1].value}, {"expr": extracted_args[2].value}],
+        )
--- a/docs/source/matchers.rst
+++ b/docs/source/matchers.rst
@@ -32,6 +32,7 @@ selectively control when LibCST calls visitor functions.
 .. autofunction:: libcst.matchers.matches
 .. autofunction:: libcst.matchers.findall
 .. autofunction:: libcst.matchers.extract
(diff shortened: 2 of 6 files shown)


libcst/matchers/tests/test_findall.py
```diff
@@ -162,3 +162,34 @@ def foo(bar: int) -> bool:
...
+            raise AssertionError(
+                f"\n{seq1!r}\nis not deeply equal to \n{seq2!r}{suffix}"
+            )
+
+        for node1, node2 in zip(seq1, seq2):
+            if not node1.deep_equals(node2):
+                raise AssertionError(
+                    f"\n{seq1!r}\nis not deeply equal to \n{seq2!r}{suffix}"
+                )
+
```
jimmylai: Can we have some tests to cover this part?

jimmylai:
Trying to check the generated doc  but for some reason, the uploaded artifacts tab didn't show up in circleci job even though the artifacts seemed uploaded successfully. https://circleci.com/gh/Instagram/LibCST/3991#tests/containers/0

libcst/matchers/tests/test_findall.py
DragonMinded: You want tests that cover the tests?

jimmylai: I know it's test helper, just feel like we probably don't want to add code that don't run but this can be used to guard future test cases.
This code block is pretty much the same as MatchersFindAllTest, can we just reuse it? E.g. keep those matcher test helpers in a base class.

https://github.com/Instagram/LibCST/blob/675f8b3e8297b362f4de9b1ba6354e35c0fb049f/libcst/matchers/tests/test_findall.py#L25-L35

DragonMinded: Its unit test code though. Its not used in production, and if a new test is written that fails, it hits this code. I see no value in testing tests nor trying to commonize this code. Its a waste of engineering effort for no benefit.

DragonMinded: Jimmy, this is dead code. I just checked and its not even called. Why are you focusing on test coverage to the detriment of an actual review?

jimmylai: In general, duplicated code adds tech debt and maintenance burden for future. I didn't specific focus on test coverage. Code quality is as important as implemented features to me and I review both of them.

DragonMinded: You're completely missing the point. I checked the code and it was entirely unused and I deleted it, it looks like a copy-paste mistake. You're focusing on the minutia of "We should have tested code" instead of the bigger picture. This isn't valuable feedback at all.

jimmylai:
image

image

image


The type annotations looks complex and may not be useful to users. We should try to simplify it.

Otherwise LGTM.

DragonMinded:
In general the sphynx generated type hints are way overcomplicated since they unroll everything :(

It is a lot simpler in code: https://github.com/Instagram/LibCST/pull/154/files#diff-4f19c4e1c6cef2912261ee532dff5807R1582-R1593

I think maybe we can get rid of the InverseOf part, since that is hidden from the type system.

DragonMinded:
I got rid of the InverseOf part and it halved the length of the type hints so I think that is a pretty good improvement. I still want to eventually tackle printing the exported name instead of the full name (so `libcst.matchers.MatchIfTrue` instead of `libcst.matchers._matcher_base.MatchIfTrue` but this is still an improvement.

PR #155, opened 2019-11-19
Make matcher type hint exception pickleable

DragonMinded:
## Summary

I guess even on newer Python versions, pickling exceptions still doesn't work right. Fix that here.

## Test Plan

Wrote a new test to verify pickleability!

### DIFF
--- a/libcst/matchers/_visitors.py
+++ b/libcst/matchers/_visitors.py
@@ -13,6 +13,7 @@ from typing import (
     Optional,
     Sequence,
     Set,
+    Tuple,
     Type,
     TypeVar,
     Union,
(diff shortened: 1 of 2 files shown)


libcst/matchers/_visitors.py
```diff
@@ -53,12 +54,23 @@
...
+def _match_decorator_unpickler(kwargs: Any) -> "MatchDecoratorMismatch":
+    return MatchDecoratorMismatch(**kwargs)
+
+
 class MatchDecoratorMismatch(Exception):
-    # pyre-ignore We don't care about the type of func, just that its callable.
-    def __init__(self, func: Callable[..., Any], message: str) -> None:
-        super().__init__(
-            # pyre-ignore Pyre doesn't believe functions have __qualname__
-            f"Invalid function signature for {func.__qualname__}: {message}"
```
jimmylai: The original code gets `__qualname__` in `MatchDecoratorMismatch` and it looks simpler.
Why move `__qualname__` to callsite? Can we just check the existence of `__qualname__` and fall back when it's not available?

DragonMinded: Because you can't pickle a `Callable`.

jimmylai: What's the use case of pickling MatchDecoratorMismatch? Is it for the output of codemod framework process?

DragonMinded: Yes. Any exception that can be thrown during a transform will need to be pickled and displayed to the user when codemodding.

PR #158, opened 2019-11-22
[metadata] add cache field to metadata wrapper

jimmylai:
## Summary
add cache field to metadata wrapper.

## Test Plan
unit tests.


libcst/metadata/tests/test_metadata_wrapper.py
```diff
@@ -41,3 +43,34 @@ def test_hash_by_identity(self) -> None:
...
+                self.cache = cache
+
+            def visit_Pass(self, node: cst.Pass) -> Optional[bool]:
+                self.set_metadata(node, self.cache)
+
+        cached_data = object()
+        # pyre-fixme[6]: Expected `Mapping[Type[BaseMetadataProvider[object]],
+        #  object]` for 2nd param but got `Dict[Type[SimpleCacheMetadataProvider],
+        #  object]`.
+        mw = MetadataWrapper(m, cache={SimpleCacheMetadataProvider: cached_data})
```
jimmylai: Not sure how to correctly type this. CC @bgw

DragonMinded: This looks like a covariance issue. Dict/Mapping are invariant in keys. Try a `Typevar` with `covariant=True` and see if that doesn't fix this: https://mypy.readthedocs.io/en/latest/generics.html#variance-of-generic-types

jimmylai: The generic type used by BaseMetadataProvider is already defined as covariant. It's not clear to me how to actually fix this.
```
_ProvidedMetadataT = TypeVar("_ProvidedMetadataT", covariant=True)

class BaseMetadataProvider(MetadataDependent, Generic[_ProvidedMetadataT]):
    ...

class SimpleCacheMetadataProvider(BatchableMetadataProvider[object]):
    ...
```

libcst/metadata/base_provider.py
```diff
@@ -44,9 +44,17 @@ class BaseMetadataProvider(MetadataDependent, Generic[_ProvidedMetadataT]):
     # explanation.
     _computed: MutableMapping["CSTNode", _ProvidedMetadataT]
 
-    def __init__(self) -> None:
+    is_cache_required: bool = False
+
+    def __init__(self, cache: object = None) -> None:
```
bgw: It's probably good enough for an MVP, but I'm not particularly happy about `cache` being typed as `object`. Is there a way we could make this a typevar? Could we infer `is_cache_required` from the generic type parameter?

jimmylai: typing_inspect.get_parameters

libcst/metadata/base_provider.py
```diff
@@ -44,9 +44,17 @@ class BaseMetadataProvider(MetadataDependent, Generic[_ProvidedMetadataT]):
...
         super().__init__()
         self._computed = {}
+        if self.is_cache_required and not cache:
```
DragonMinded: This should explicitly check against `cache is None`, otherwise we will have no way of passing a cache such as a dictionary that is empty.

libcst/metadata/base_provider.py
DragonMinded: I would not use `get_parameters` to make caching depending on the existence of an init param. Instead, either we should use type generics as Ben suggested, or provide a cache-required base class and a no-cache-required base class, and check for `isinstance` of either to determine cache requirements.

libcst/metadata/base_provider.py
```diff
@@ -44,9 +44,17 @@ class BaseMetadataProvider(MetadataDependent, Generic[_ProvidedMetadataT]):
...
+            # The metadata provider implementation is responsible to store and use cache.
+            raise Exception(
+                f"Cache is required for initializing {self.__class__.__name__}."
```
DragonMinded: This exception can explain *how* to use cache, such as "Cache is required..... Please use <method name here> to calculate cache and pass in using the `cache` parameter." Something to the effect of helping the user to the solution instead of just saying there's a problem.

libcst/metadata/tests/test_metadata_wrapper.py
DragonMinded: Perhaps we can't use a Mapping here? What if we introduce a `MetadataCache` class that holds all of the per-file provider-to-data mappings? We need some other layer that can calculate cache for a list of files, and then grab the cache per-file for transform/metadata instantiation anyway. So if we sidestep the Mapping covariance issue with our own wrapper it might also help build the next layer?

jimmylai: We discussed about building a full repo manager to handle generating the per-file provider-to-data given a list. It will come in the next few PRs.

DragonMinded: Ah, then I don't think its worth solving this issue for now, pyre-ignore seems fine.

jimmylai: We discussed about building a full repo manager to handle generating the per-file provider-to-data given a list. It will come in the next few PRs.

libcst/metadata/base_provider.py
jimmylai: The second type generic requires changing the signature of existing metadata provider which will break our custom metadata provider in distillery.

libcst/metadata/base_provider.py
jimmylai: Make sense. We don't have the helper to get cache yet, so I didn't add it in the comment. I can add it now as a placeholder.

DragonMinded:
Why are the majority of the changed files just formatting changes?

jimmylai:
> LGTM aside from unrelated formatting changes.

Tht formatting changes was done by: tox -e autofix
The trailing commas were removed if it's a single argument to a call which was from new version of Black.
We should've formatting the entire codebase when we upgrade Black.

PR #159, opened 2019-11-23
Type inference metadata

jimmylai:
## Summary
Type inference metadata surface type annotation data (str) provided by pyre query on Name, Attribute and Call nodes.

Sometimes type annotation is not available and the available type annotation may not be useful. We'll continue to work with Pyre team to increase the coverage and accuracy.

This is the first step, later on, we may be able to use Scope Analysis to simplify the type and it'll be useful if we can provide the Assignment/Scope info.

## Test Plan
Unit tests.

### DIFF
--- a/libcst/metadata/tests/test_type_inference_provider.py
+++ b/libcst/metadata/tests/test_type_inference_provider.py
@@ -1,273 +1,66 @@
-from textwrap import dedent
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# pyre-strict
+
+import json
+from pathlib import Path
 
 import libcst as cst
 from libcst import MetadataWrapper
 from libcst.metadata.type_inference_provider import TypeInferenceProvider
-from libcst.testing.utils import UnitTest
+from libcst.testing.utils import UnitTest, data_provider
+from libcst.tests.test_pyre_integration import TEST_SUITE_PATH, PyreData
 
 
 class TypeInferenceProviderTest(UnitTest):
-    def test_basic_class_types(self) -> None:
+    @data_provider(
+        ((TEST_SUITE_PATH / "simple_class.py", TEST_SUITE_PATH / "simple_class.json"),)
+    )
+    def test_simple_class_types(self, source_path: Path, data_path: Path) -> None:
+        data: PyreData = json.loads(data_path.read_text())
         wrapper = MetadataWrapper(
-            cst.parse_module(
-                dedent(
-                    """\
-                    from typing import Sequence
-
-
-                    class Item:
-                        def __init__(self, n: int):
-                            self.number = n
-
-
-                    class ItemCollector:
-                        def get_items(self, n: int) -> Sequence[Item]:
-                            return [Item() for i in range(n)]
-
-
-                    collector = ItemCollector()
-                    items = collector.get_items()
-                    for item in items:
-                        item.number
-                    """
-                )
-            ),
+            cst.parse_module(source_path.read_text()),
             # pyre-fixme[6]: Expected `Mapping[Type[BaseMetadataProvider[object]],
-            #  object]` for 2nd param but got `Dict[Type[TypeInferenceProvider],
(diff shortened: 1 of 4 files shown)


libcst/metadata/type_inference_provider.py
```diff
@@ -0,0 +1,38 @@
...
+                "annotation"
+            ]
+        self.lookup = lookup
+
+    def _parse_metadata(self, node: cst.CSTNode) -> None:
+        range = self.get_metadata(PositionProvider, node)
+        key = (range.start.line, range.start.column, range.end.line, range.end.column)
+        if key in self.lookup:
+            self.set_metadata(node, self.lookup[key])
+            self.lookup.pop(key)
```
bgw: This is a weird use of pop. If you don't care about the return value, the more idiomatic pattern for dicts seems to be

```
del self.lookup[key]
```

(discarding the return value of pop on a list is common, but I think that's just because there's not as good of an alternative)

libcst/metadata/type_inference_provider.py
```diff
@@ -0,0 +1,38 @@
...
+    is_cache_required = True
+
+    def __init__(self, cache: object):
+        super().__init__(cache)
+        lookup = {}
+        for item in cache:
+            location = item["location"]
+            start = location["start"]
+            end = location["stop"]
+            lookup[(start["line"], start["column"], end["line"], end["column"])] = item[
```
bgw: It'd be interesting to run this over IGSRV and log if we ever get duplicate ranges. That was one concern I had with this approach. We probably won't, but if it did happen, we might want to start considering node types in addition to positional ranges.

libcst/metadata/tests/test_type_inference_provider.py
```diff
@@ -0,0 +1,259 @@
...
+                    ).body.body[0],
+                    cst.SimpleStatementLine,
+                ).body[0],
+                cst.Assign,
+            )
+            .targets[0]
+            .target,
+            cst.Attribute,
+        )
+        self.assertEqual(types[self_number_attr], "int")
```
bgw: This is a rather large example, but then you only inspect two types on it.

I'd either test more things, or strip down the size of the example.

libcst/metadata/type_inference_provider.py
jimmylai: I can just combine the two statements together.

libcst/metadata/type_inference_provider.py
jimmylai: Make sense, I can run some test and get some statistic numbers.

libcst/metadata/tests/test_type_inference_provider.py
jimmylai: Some more asserts are added after your review.

libcst/metadata/type_inference_provider.py
mvismonte: Did you ever run this on IGSRV?

libcst/metadata/type_inference_provider.py
```diff
@@ -0,0 +1,62 @@
...
+    annotation: str
+
+
+class TypeInferenceProvider(BatchableMetadataProvider[str]):
+    METADATA_DEPENDENCIES = (PositionProvider,)
+    is_cache_required = True
+
+    def __init__(self, cache: List[InferredType]) -> None:
+        super().__init__(cache)
+        lookup: Dict[Tuple[int, int, int, int], str] = {}
```
mvismonte: What's the reason why  we're not using a datatype here for the key instead?  It might makes things more readable in the long term

libcst/metadata/type_inference_provider.py
jimmylai: We want to finish end to end integration first and then run on IGSRV to collect more examples in test suite. Back to when we just started the integration work, Rey run on IGSRV files to find mismatch positions and report to Pyre team (Shannon). Shannon already fixed many reported issues. So TypeInference provider assumes that the position always matches, if it doesn't match, then no inferred type is returned.
Once we landed this in IGSRV, I'll run on more IGSRV files to collect issues to report and build test suites.

libcst/metadata/type_inference_provider.py
jimmylai: Change to use existing data class `CodeRange`.

PR #162, opened 2019-12-05
Fix mark_no_op decorator type preservation on mypy

### LINKED ISSUE #161: mypy CSTTypedTransformerFunctions issue
## context
ran into a peculiar bug where the `CSTTypedTransformerFunctions` types seem to not work with `mypy`. `leave_{Node}` calls fail mypy type check with `Signature of "leave_{Node}" is incompatible with supertype "CSTTypedTransformerFunctions"`.

the type definition that mypy says is incompatible doesnt *look* incompatible to me, though: https://github.com/Instagram/LibCST/blob/c9b10fe427c12c494b894fb3957afd75047591e7/libcst/_typed_visitor.py#L5785-L5787

i'm not sure yet if this is an issue with mypy or libcst, but wanted to share here for replication. (mypy's error messages aren't super descriptive so am still trying to figure out _where_ mypy is getting the incompatible defintion from.)

#### `./requirements.txt`
```
libcst==0.2.4
mypy==0.750
```

#### `./transformer.py`
```py
import libcst

class Transformer(libcst.CSTTransformer):

    def leave_Module(self, original_node: libcst.Module, updated_node: libcst.Module) -> libcst.Module:
        return updated_node

    def leave_Call(self, original_node: libcst.Call, updated_node: libcst.Call) -> libcst.Call:
        return updated_node
```

#### `mypy errors`
```
$ mypy transformer.py 
transformer.py:5: error: Signature of "leave_Module" incompatible with supertype "CSTTypedTransformerFunctions"
transformer.py:8: error: Signature of "leave_Call" incompatible with supertype "CSTTypedTransformerFunctions"
Found 2 errors in 1 file (checked 1 source file)
```

zhammer:
## Summary
fix #161 by following decorator type preservation example code in mypy docs: https://mypy.readthedocs.io/en/latest/generics.html#declaring-decorators

## Test Plan
will consult with libcst folks on type tests

### DIFF
--- a/libcst/_typed_visitor_base.py
+++ b/libcst/_typed_visitor_base.py
@@ -10,8 +10,7 @@ from typing import TYPE_CHECKING, Any, Callable, TypeVar, cast
 if TYPE_CHECKING:
     from libcst._typed_visitor import CSTTypedBaseFunctions  # noqa: F401
 
-T = TypeVar("T")
-FuncType = Callable[["CSTTypedBaseFunctions", T], None]
+FuncType = Callable[["CSTTypedBaseFunctions", Any], None]
 F = TypeVar("F", bound=FuncType)
 
 
--- a/libcst/_typed_visitor_base.py
+++ b/libcst/_typed_visitor_base.py
@@ -10,8 +10,7 @@ from typing import TYPE_CHECKING, Any, Callable, TypeVar, cast
 if TYPE_CHECKING:
     from libcst._typed_visitor import CSTTypedBaseFunctions  # noqa: F401
 
-FuncType = Callable
-F = TypeVar("F", bound=FuncType)
+F = TypeVar("F", bound=Callable)
 
 
 def mark_no_op(f: F) -> F:
--- a/libcst/_typed_visitor_base.py
+++ b/libcst/_typed_visitor_base.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, Any, Callable, TypeVar, cast
 if TYPE_CHECKING:
     from libcst._typed_visitor import CSTTypedBaseFunctions  # noqa: F401
 
-FuncType = Callable[["CSTTypedBaseFunctions", Any], None]
+FuncType = Callable[["CSTTypedBaseFunctions", ...], None]
 F = TypeVar("F", bound=FuncType)
 
 
--- a/libcst/_typed_visitor_base.py
+++ b/libcst/_typed_visitor_base.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, Any, Callable, TypeVar, cast
 if TYPE_CHECKING:
     from libcst._typed_visitor import CSTTypedBaseFunctions  # noqa: F401
 
-FuncType = Callable[["CSTTypedBaseFunctions", ...], None]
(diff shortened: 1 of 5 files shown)


zhammer:
hmm.. going to look into how pyre handles decorator typing. https://github.com/facebook/pyre-check/commit/6743d5ffbdd77ae8755f10c751b2adf640394748

zhammer:
yeah, not really sure how to fix this in a way that satisfies mypy and pyre. also would be odd to do a mypy fix with type checking CI is only done for pyre as it could break again in the future.

main downside of not fixing is folks using mypy can't use static types.

libcst/_typed_visitor_base.py
```diff
@@ -10,12 +10,11 @@
 if TYPE_CHECKING:
     from libcst._typed_visitor import CSTTypedBaseFunctions  # noqa: F401
 
-T = TypeVar("T")
+FuncType = Callable
```
DragonMinded: Hmm, yeah. I know pyre is moving away from generics without a constraint, so this probably won't work going forward with pyre anyway.

DragonMinded: Does mypy require the extra level of indirection? If not, can you change the below line to just `F = TypeVar("F", bound=Callable)` directly and make this work on both?

zhammer: Ha didn’t mean to resolve that. This is a good idea, I’ll try it tomorrow.

libcst/_typed_visitor_base.py
```diff
@@ -10,12 +10,10 @@
 if TYPE_CHECKING:
     from libcst._typed_visitor import CSTTypedBaseFunctions  # noqa: F401
 
-T = TypeVar("T")
+F = TypeVar("F", bound=Callable)
 
 
-def mark_no_op(
-    f: Callable[["CSTTypedBaseFunctions", T], None]
```
zhammer: @DragonMinded - this change *does* remove the typing constraint that `mark_no_op` can only decorate this signature of callable, right?

I imagine it's more important to preserve func signature than type the decorated function.

zhammer:
hmmm.. new pyre failures, will look at this as well. also i'm having issues installing / running pyre locally which is why this has been a lot of commits to this branch

libcst/_typed_visitor_base.py
zhammer: added in 05c86d8, thanks @DragonMinded :)

zhammer:
comment from e5ac6b3 which includes a few updates to visitor function signatures in the test suite.

> It seems that @mark_no_op was not correctly preserving type signatures
> in pyre previous to this change. Preserving type signatures in the
> decorator caused some pyre errors in tests where visitors are defined
> that don't share the same arg names as their parent.

zhammer:
yay!

libcst/_typed_visitor_base.py
DragonMinded: Honestly, there's no safety concern in accidentally decorating a class or a function that we don't care about. I would rather have type safety of decorated functions, especially since `mark_no_op` is an implementation detail of LibCST core. This is fine.

PR #164, opened 2019-12-11
Intial Codemod Framework Open-Source

DragonMinded:
## Summary

This is the initial work needed to open-source the codemod framework. There is still a lot missing that I will be adding in a future PR: No ability to list/set up a repo, no config file format yet, lots of convenience features missing, etc. This code is mostly a direct export from Instagram's internal repo, with changes made to imports only. Hopefully splitting this way helps reviews.

## Test Plan

tox, pyre.

PR #165, opened 2019-12-12
Codemod Utilities and Example Codemods

DragonMinded:
## Summary

This PR brings the missing feature-set that I alluded to in PR #164. This adds a config file format to override defaults, a utility for initializing the config file, a utility for listing out known codemods given an existing config file, and a series of universally-applicable codemods that were previously developed inside Instagram. In the spirit of continuing to use `libcst.tool` for command-line operations related to LibCST, all of these utilities were added to `libcst.tool` and it was refactored a bit to allow for this.

Note that this is still missing documentation, which I am working on in yet another PR. However, it is complete and therefore worth releasing.

## Test Plan

pyre, tox, run `libcst.tool` locally and verify that its behavior is correct. I also used `libcst.tool codemod` to fix pyre-strict issues in the codemod directory, since Instagram internally uses `pyre-unsafe` and is strict by default.

### DIFF
--- a/libcst/codemod/__init__.py
+++ b/libcst/codemod/__init__.py
@@ -1,3 +1,8 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+#
 # pyre-strict
 from libcst.codemod._cli import (
     ParallelTransformResult,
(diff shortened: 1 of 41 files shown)


DragonMinded:
Checking on coverage, the uncovered files are all CLI utilities that are inherently impossible/pointless to test. The uncovered bits of libcst.codemod are now correctly covered by the additional tests and codemods with the exception of some uncovered files for the fstring codemod. I'm putting up tests to fix that now.

PR #172, opened 2019-12-17
add test suite for pyre regression test

jimmylai:
## Summary
add test suite for pyre regression test.
Explicitly not to add `pyre-strict` to test files to make sure pyre can infer some of the types smartly.
Missing or wrong inferred typing issues are commented.
CC @shannonzhu

Run the command to generate pyre output in test suite directory tests/pyre: `python libcst/tests/test_pyre_integration.py`
Tried to run the script in CI to automatically verify the generated file is up to date but pyre query depends on watchman and it's not straightforward to install watchman in CI environment.
This attempt filed: https://app.circleci.com/jobs/github/Instagram/LibCST/4633/parallel-runs/0/steps/0-104

Pyre 0.0.36 didn't work well. Upgrade to 0.0.38 resolved a couple type coverage and accuracy issues.

Issue: we should try avoiding the test py file was formatted unexpectedly and caused the pyre data inconsistency. Solution: add `# fmt: off` to top of file.

## Test Plan

Unit tests.

### DIFF
--- a/libcst/tests/pyre/simple_class.json
+++ b/libcst/tests/pyre/simple_class.json
@@ -1,0 +1,438 @@
+{
+  "types": [
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 11,
+          "column": 26
+        },
+        "stop": {
+          "line": 11,
+          "column": 29
+        }
+      },
+      "annotation": "typing.Type[int]"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 22,
+          "column": 4
+        },
+        "stop": {
+          "line": 22,
+          "column": 8
+        }
+      },
+      "annotation": "libcst.tests.pyre.simple_class.Item"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 16,
+          "column": 44
+        },
+        "stop": {
+          "line": 16,
+          "column": 48
+        }
+      },
+      "annotation": "typing.Type[libcst.tests.pyre.simple_class.Item]"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 22,
+          "column": 12
+        },
+        "stop": {
+          "line": 22,
+          "column": 17
+        }
+      },
+      "annotation": "typing.Sequence[libcst.tests.pyre.simple_class.Item]"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 15,
+          "column": 6
+        },
+        "stop": {
+          "line": 15,
+          "column": 19
+        }
+      },
+      "annotation": "typing.Type[libcst.tests.pyre.simple_class.ItemCollector]"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 16,
+          "column": 18
+        },
+        "stop": {
+          "line": 16,
+          "column": 22
+        }
+      },
+      "annotation": "libcst.tests.pyre.simple_class.ItemCollector"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 17,
+          "column": 16
+        },
+        "stop": {
+          "line": 17,
+          "column": 22
+        }
+      },
+      "annotation": "libcst.tests.pyre.simple_class.Item"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 21,
+          "column": 8
+        },
+        "stop": {
+          "line": 21,
+          "column": 17
+        }
+      },
+      "annotation": "libcst.tests.pyre.simple_class.ItemCollector"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 12,
+          "column": 22
+        },
+        "stop": {
+          "line": 12,
+          "column": 23
+        }
+      },
+      "annotation": "int"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 16,
+          "column": 27
+        },
+        "stop": {
+          "line": 16,
+          "column": 30
+        }
+      },
+      "annotation": "typing.Type[int]"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 16,
+          "column": 35
+        },
+        "stop": {
+          "line": 16,
+          "column": 43
+        }
+      },
+      "annotation": "typing.Callable(typing.GenericMeta.__getitem__)[[typing.Type[Variable[typing._T_co](covariant)]], typing.Type[typing.Sequence[Variable[typing._T_co](covariant)]]]"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 20,
+          "column": 0
+        },
+        "stop": {
+          "line": 20,
+          "column": 9
+        }
+      },
+      "annotation": "libcst.tests.pyre.simple_class.ItemCollector"
+    },
+    {
+      "location": {
+        "path": "libcst/tests/pyre/simple_class.py",
+        "start": {
+          "line": 17,
+          "column": 32
+        },
+        "stop": {
+          "line": 17,
+          "column": 37
+        }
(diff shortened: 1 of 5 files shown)


libcst/tests/pyre/simple_class.json
```diff
@@ -0,0 +1,242 @@
...
+        "start": {
+          "line": 11,
+          "column": 8
+        },
+        "stop": {
+          "line": 11,
+          "column": 16
+        }
+      },
+      "annotation": "typing.Callable(libcst.tests.pyre.simple_class.Item.__init__)[[Named(self, unknown), Named(n, unknown)], unknown]"
```
jimmylai: The type of self, n, and return of `__init__` are marked as unknown.
@shannonzhu

libcst/tests/pyre/simple_class.json
```diff
@@ -0,0 +1,242 @@
...
+        "start": {
+          "line": 16,
+          "column": 8
+        },
+        "stop": {
+          "line": 16,
+          "column": 17
+        }
+      },
+      "annotation": "typing.Callable(libcst.tests.pyre.simple_class.ItemCollector.get_items)[[Named(self, unknown), Named(n, unknown)], unknown]"
```
jimmylai: Same unknown typing issue here.

libcst/tests/pyre/simple_class.json
```diff
@@ -0,0 +1,242 @@
...
+        "start": {
+          "line": 16,
+          "column": 27
+        },
+        "stop": {
+          "line": 16,
+          "column": 30
+        }
+      },
+      "annotation": "typing.Type[]"
```
jimmylai: Missing type here, it should be `"typing.Type[int]"`.

libcst/tests/pyre/simple_class.py
```diff
@@ -0,0 +1,23 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# fmt: off
+from typing import Sequence
+
+
+class Item:
+    def __init__(self, n: int) -> None:
+        self.number = n
```
jimmylai: We expect pyre to infer `self.number` has type int but it's not.

libcst/tests/pyre/simple_class.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class ItemCollector:
+    def get_items(self, n: int) -> Sequence[Item]:
+        return [Item(i) for i in range(n)]
```
jimmylai: The type of `range` is missing.

libcst/tests/pyre/simple_class.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+collector = ItemCollector()
+items = collector.get_items(3)
```
jimmylai: The type of `collector.get_items` (a callable) is missing.

libcst/tests/test_pyre_integration.py
```diff
@@ -0,0 +1,67 @@
...
+    annotation: str
+
+
+def _sort_by_position(data: InferredType) -> Tuple[int, int, int, int]:
+    start = data["location"]["start"]
+    stop = data["location"]["stop"]
+    return (start["line"], start["column"], stop["line"], stop["column"])
+
+
+if __name__ == "__main__":
```
mvismonte: Where is the actual test being run?  This looks like it's generating data based off the python files in `libcst/tests/pyre`

jimmylai: @mvismonte you read it right, the actually unit test was just added now.
It's `PyreIntegrationTest`. Currently, Pyre don't return typing on a couple cases and I've report them to Shannon.

libcst/tests/test_pyre_integration.py
```diff
@@ -0,0 +1,172 @@
...
+        start = pos.start
+        end = pos.end
+        self.attributes.append(node)
+        tup = (start.line, start.column, end.line, end.column)
+        # remove this if condition when the type issues are fixed.
+        if not any(
+            node.deep_equals(name) and tup == _tup
+            for (name, _tup) in {
+                (cst.Name("n"), (11, 23, 11, 24)),
+                (cst.Name("None"), (11, 34, 11, 38)),
```
jimmylai: @shannonzhu Those two name are missing types from pyre.

libcst/tests/test_pyre_integration.py
```diff
@@ -0,0 +1,172 @@
...
+        self.attributes.append(node)
+        tup = (start.line, start.column, end.line, end.column)
+
+        # remove this if condition when the type issues are fixed.
+        if not any(
+            node.deep_equals(name)
+            for name in {
+                cst.Attribute(cst.Name("self"), attr=cst.Name("number")),
+                cst.Attribute(cst.Name("collector"), attr=cst.Name("get_items")),
+                cst.Attribute(cst.Name("item"), attr=cst.Name("number")),
```
jimmylai: @shannonzhu Those attributes are missing types from Pyre.

libcst/tests/test_pyre_integration.py
jimmylai: Running test_pyre_integration.py is a codegen. (We also have codegen for other LibCST code).
We run it manually and commit the generated code. In this case, the generated code is `simple_class.json` which is the output of Pyre query with some trans.

libcst/tests/pyre/simple_class.py
shannonzhu: The reason pyre doesn't infer `int` for `self.number` here is because this attribute was not given an annotation (this would be an error in strict: "Missing attribute annotation [4]: Attribute `number` of class `Item` has type `int` but no type is specified.") 

It's true that at the moment of assignment we are sure that `self.number` is an integer. But would it be more confusing to special case this, when any later reference to `self.number` will be unknown? Maybe consistency is better here.

Note: Pyre does infer that if you assign an attribute like:
- a literal, ie.`self.x = 1`
- an annotated parameter with the same name, ie. `self.number = number`

Then Pyre will assume that the annotation will always be of the type at assignment unless otherwise specified. In these cases, attempting to assign the attribute to something of a different type later on will yield a type error, the same way as it would for an attribute declared like `self.number: int = n`. 
This is where we draw the line, because when the value field gets more complex some or parts of the type could be unknown, and it becomes less safe or consistent for pyre to assume an explicit annotation in cases where it happens to know the type.

Let me know if this makes sense! Will move on to addressing the other issues brought up for now, but happy to revisit this one. :)

libcst/tests/pyre/simple_class.py
shannonzhu: Range is just a class!
https://github.com/python/typeshed/blob/master/stdlib/2and3/builtins.pyi#L1075

Pyre does seem to reveal the type of range correctly in lookup:

      "4:33-4:38/typing.Type[range]";
      "4:33-4:41/range";

libcst/tests/pyre/simple_class.py
shannonzhu: Hm, when I tested this I was able to get the type of get_items as part of the JSON output:
{"location":{"start":{"line":52,"column":8},"stop":{"line":52,"column":27}},"annotation":"typing.Callable(test_prop.ItemCollector.get_items)[[Named(n, int)], typing.Sequence[test_prop.Item]]"},

my test file looks like this:

```
import typing
from typing import (
    TYPE_CHECKING,
    Any,
    AsyncGenerator,
    AsyncIterable,
    AsyncIterator,
    Awaitable,
    BinaryIO,
    Callable,
    ClassVar,
    Collection,
    DefaultDict,
    Dict,
    Final,
    Generator,
    Generic,
    Iterable,
    Iterator,
    List,
    Mapping,
    NamedTuple,
    NewType,
    Optional,
    Protocol,
    Sequence,
    Set,
    SupportsInt,
    Tuple,
    Type,
    TypeAlias,
    TypeVar,
    Union,
    cast,
    overload,
)
from mypy_extensions import TypedDict
from typing_extensions import Literal

# pyre-strict

class Item:
    def __init__(self, n: int) -> None:
        self.number = n

class ItemCollector:
    def get_items(self, n: int) -> Sequence[Item]:
        return [Item(i) for i in range(n)]

collector = ItemCollector()
reveal_type(collector.get_items)
items = collector.get_items(3)

```

Let me know if I'm reproing something differently, happy to take another look!

libcst/tests/pyre/simple_class.py
jimmylai: This test file didn't enable pyre strict mode. So, Pyre doesn't complain self.number is missing type. The explanation of missing type annotation make sense. We can probably assume all code will be in pyre strict mode and I can update the test case as pyre-strict.

libcst/tests/pyre/simple_class.py
jimmylai: Correct, my comment was outdated, the type shows up after change to use 0.0.38.

libcst/tests/pyre/simple_class.py
jimmylai: Correct, my comment was outdated, the type shows up after change to use 0.0.38.

PR #179, opened 2019-12-26
Full repo manager

jimmylai:
## Summary
Add Full repo manager.
Also add Sphinx documents.
https://4831-200896124-gh.circle-artifacts.com/0/doc/metadata.html
image

## Test Plan
Integration test:
image

Also used the test suite data and mock the IPC calls for unit tests.

### DIFF
--- a/libcst/metadata/full_repo_manager.py
+++ b/libcst/metadata/full_repo_manager.py
@@ -1,0 +1,105 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# pyre-strict
+import json
+import subprocess
+from pathlib import Path
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    Dict,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
+)
+
+from mypy_extensions import TypedDict
+
+from libcst._parser.entrypoints import parse_module
+from libcst.metadata.type_inference_provider import (
+    InferredType,
+    PyreData,
+    TypeInferenceProvider,
+)
+from libcst.metadata.wrapper import MetadataWrapper
+
+
+if TYPE_CHECKING:
+    from libcst.metadata.base_provider import ProviderT
+
+
+class FullRepoManager:
+    def __init__(
+        self,
+        repo_root_dir: str,
+        paths: List[str],
+        providers: List["ProviderT"],
+        timeout: int = 5,
+    ) -> None:
+        self.root_path: Path = Path(repo_root_dir)
+        self._cache: Dict["ProviderT", Mapping[str, object]] = {}
+        self._timeout = timeout
+        for provider in providers:
+            handler = self._get_cache_handler(provider)
+            if handler:
+                self._cache[provider] = handler(paths)
+
+    def _get_cache_handler(self, provider: "ProviderT") -> Optional[Callable]:
+        maps: Dict["ProviderT", Callable[[List[str]], Mapping[str, object]]] = {
+            TypeInferenceProvider: self._handle_pyre_cache
+        }
+        return maps.get(provider)
+
+    def _handle_pyre_cache(self, paths: List[str]) -> Mapping[str, object]:
+        params = ",".join(f"path='{self.root_path/path}'" for path in paths)
+        cmd = f'''pyre query "types({params})"'''
+        try:
+            stdout, stderr, return_code = run_command(cmd, timeout=self._timeout)
+        except subprocess.TimeoutExpired as exc:
+            raise exc
+
+        if return_code != 0:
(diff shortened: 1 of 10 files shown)


libcst/metadata/full_repo_manager.py
```diff
@@ -0,0 +1,107 @@
...
+        providers: List["ProviderT"],
+        timeout: int = 5,
+    ) -> None:
+        self.root_path: Path = Path(repo_root_dir)
+        self._cache: Dict["ProviderT", Mapping[str, object]] = {}
+        self._timeout = timeout
+        for provider in providers:
+            handler = self._get_cache_handler(provider)
+            if handler:
+                self._cache[provider] = handler(paths)
```
mvismonte: Correct me if I'm wrong, but this looks like it's making shell calls in the constructor.  IMO, it's best to do expensive processing in another method.  Constructing objects should be light

libcst/metadata/full_repo_manager.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def _get_cache_handler(self, provider: "ProviderT") -> Optional[Callable]:
+        maps: Dict["ProviderT", Callable[[List[str]], Mapping[str, object]]] = {
+            TypeInferenceProvider: self._handle_pyre_cache
+        }
+        return maps.get(provider)
```
mvismonte: Should we make this map an argument to the constructor instead?  I notice that in your tests that this method is patched.  If we make the maps an argument, then we can avoid patching and pass in our own handler method.

mvismonte (changes requested):
Looks good overall.  Just a few questions about the API design

libcst/metadata/full_repo_manager.py
jimmylai: This method is not patched and it's run in the tests.
The patched method is `_handle_pyre_cache` which it requires actually query pyre daemon (which is an integration test). We don't want to build an integration test as a unit test to avoid flaky or slow test.

libcst/metadata/full_repo_manager.py
jimmylai: I understand this concern, we need to run the batched pyre query before user calling.
If we want to make the constructor cheap, we can add one resolve step and cache the result.
```
m = FullRepoManager(root, paths, providers)
m.get_metadata_wrapper_for_path(path)  # it implicitly resolve the cache and store for reuse.
```

jimmylai: The implicit resolve step is added as `_resolve_cache()`

libcst/metadata/full_repo_manager.py
```diff
@@ -0,0 +1,154 @@
...
+        metadata provider like :class:`~libcst.metadata.TypeInferenceProvider`.
+
+        :param paths: a collection of paths to access full repository data.
+        :param providers: a collection of metadata provider classes require accessing full repository
+            data, currently supports :class:`~libcst.metadata.TypeInferenceProvider`.
+        :param timeout: number of seconds. Raises `TimeoutExpired <https://docs.python.org/3/library/subprocess.html#subprocess.TimeoutExpired>`_
+            when timeout.
+        """
+        self.root_path: Path = Path(repo_root_dir)
+        self._cache: Dict["ProviderT", Mapping[str, object]] = {}
```
mvismonte: I  guess there's no way of specifying that the type of the value of the map is whatever is provided by `ProviderT`?

libcst/metadata/full_repo_manager.py
mvismonte: Thanks for moving the expensive calls out!  This definitely feels better.  I think I wasn't as clear on the other part before.  I'm wondering if we should pass in the map you define on line 65 into the constructor as well.  This way, we can pass in a lambda function as the handler function for `TypeInferenceProvider` instead of mocking out `_handle_pyre_cache`.  In the long term, this feels to me so that we don't bloat this class with handler function implementations, but you've thought about this API design a bit more than I have.

mvismonte (approved):
Comments inline, but this looks good overall

libcst/metadata/full_repo_manager.py
jimmylai: The current implementation keeps `TypeInferenceProvider` shell out logic in FullRepoManager. The other way is to keep the logic in `TypeInferenceProvider` to keep related logic together.

libcst/metadata/full_repo_manager.py
jimmylai: We didn't think of  an easy way to do that. One idea we had to annotate the type of cache for each provider is to add a second generic type variable in provider class definition for the cache type. This approach requires inspect the second generic type at runtime which make it complex.

jimmylai:
Addressed all comments and moved type inference related logic to `TypeInferenceProvider`.
From now on each metadata provider requires cache needs to implement `gen_cache()` and `FullRepoManager` calls `gen_cache` to batch compute required cache.

PR #189, opened 2020-01-08
Support codemods in Subdirectories of configured modules.

DragonMinded:
## Summary

The way I coded this originally assumed that all codemods would exist in a single directory. This is already not the case inside Instagram, where we have multiple nested subdirectories of codemods. So, implement the ability to iterate over subdirectories. While I'm at it, deduplicate codemods that were re-exported from other modules such as tests by keeping only one version of a runnable codemod around.

## Test Plan

tox, pyre, added some subdirectories and verified that codemods were listed, verified tests were not listed, verified duplicates were not listed.

### DIFF
--- a/libcst/tool.py
+++ b/libcst/tool.py
@@ -19,7 +19,7 @@ import os.path
 import sys
 import textwrap
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Dict, List, Sequence
+from typing import Any, Callable, Dict, List, Sequence, Tuple, Type
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


libcst/tool.py
```diff
@@ -636,6 +636,38 @@ def _initialize_impl(proc_name: str, command_args: List[str]) -> int:
     return 0
 
 
+def _recursive_find(base_dir: str, base_module: str) -> List[Tuple[str, object]]:
```
jimmylai: Can we use `glob.glob` or `Path.rglob` to find all files in a simpler way?

DragonMinded: Possibly, but we'd still need to mangle the directory names to module names, and we don't get much from glob or rglob.

PR #190, opened 2020-01-08
Add docs for CodemodTest

DragonMinded:
## Summary

Add docs for `CodemodTest`, so that people know how to test their codemods. Unfortunately it looks like sphinx cannot both include inherited members AND specify an explicit list of members, so in order to get documentation to show up right for this class I had to explicitly opt out all of `Unittest`. What a pain. However, it works, and the documentation rendering properly is more important than the code looking nice.

I also fixed the warnings that crept into the docs process during this PR.

## Test Plan

tox -e docs, verified no errors, verified docs rendered right.

codemodtest

### DIFF
--- a/docs/source/metadata.rst
+++ b/docs/source/metadata.rst
@@ -187,12 +187,12 @@ We provide :class:`~libcst.metadata.ParentNodeProvider` for those use cases.
 
 Type Inference Metadata
 -----------------------
-`Type inference <https://en.wikipedia.org/wiki/Type_inference>`_ is to automatically infer
+`Type inference <https://en.wikipedia.org/wiki/Type_inference>`__ is to automatically infer
 data types of expression for deeper understanding source code.
 In Python, type checkers like `Mypy <https://github.com/python/mypy>`_ or
-`Pyre <https://pyre-check.org/>`_ analyze `type annotations <https://docs.python.org/3/library/typing.html>`_
+`Pyre <https://pyre-check.org/>`__ analyze `type annotations <https://docs.python.org/3/library/typing.html>`__
 and infer types for expressions.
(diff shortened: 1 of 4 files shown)


docs/source/metadata.rst
```diff
@@ -187,12 +187,12 @@ We provide :class:`~libcst.metadata.ParentNodeProvider` for those use cases.
...
 -----------------------
-`Type inference <https://en.wikipedia.org/wiki/Type_inference>`_ is to automatically infer
+`Type inference <https://en.wikipedia.org/wiki/Type_inference>`__ is to automatically infer
 data types of expression for deeper understanding source code.
 In Python, type checkers like `Mypy <https://github.com/python/mypy>`_ or
-`Pyre <https://pyre-check.org/>`_ analyze `type annotations <https://docs.python.org/3/library/typing.html>`_
+`Pyre <https://pyre-check.org/>`__ analyze `type annotations <https://docs.python.org/3/library/typing.html>`__
 and infer types for expressions.
-:class:`~libcst.metadata.TypeInferenceProvider` is provided by `Pyre Query API <https://pyre-check.org/docs/querying-pyre.html>`_
+:class:`~libcst.metadata.TypeInferenceProvider` is provided by `Pyre Query API <https://pyre-check.org/docs/querying-pyre.html>`__
```
DragonMinded: I converted these to anonymous links to silence duplicate anchor errors.

libcst/metadata/position_provider.py
```diff
@@ -127,9 +127,7 @@ class PositionProvider(BaseMetadataProvider[CodeRange]):
     significant.
 
     The positions provided by this provider should eventually match the positions used
-    by Pyre_ for equivalent nodes.
-
-    .. _Pyre: https://github.com/facebook/pyre-check
+    by `Pyre <https://github.com/facebook/pyre-check>`__ for equivalent nodes.
```
DragonMinded: Same comment here, silencing duplicate errors.

PR #205, opened 2020-01-10
[helper] QualifiedNameProvider.has_name

jimmylai:
## Summary
Those patterns are used very often currently and this helper makes it simpler to users.
`any(qn.name == target_name for qn in self.get_metadata(QualifiedNameProvider, node, set()))`
`any(qn == target_qualified_name for qn in self.get_metadata(QualifiedNameProvider, node, set()))`
Some user may do it with for loop + if condition which makes it more complex.

With the new helper: `QualifiedNameProvider.has_name(visitor, node, "target")`

## Test Plan
Unit test.

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -27,6 +27,7 @@ from typing import (
 
 import libcst as cst
 from libcst._add_slots import add_slots
+from libcst._metadata_dependent import MetadataDependent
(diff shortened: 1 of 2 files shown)


jimmylai:
Rename to `has_name`.
Most of the use cases need a `has_name`, few use cases need a `contains_any`. Provided the `has_name` helper for now.

Issue #122, opened 2019-10-21
3.8 Support


cooperlees:
I have a test runner that uses Pyre as an option for type-checking, [ptr](https://github.com/facebookincubator/ptr). I put up a diff to move CI to start using official 3.8 for test running on travis and this results in the expected ValueError:
```
raise ValueError(
ValueError: LibCST can only parse code using one of the following versions of Python's grammar: 3.5, 3.6, 3.7. More versions may be supported by future releases.
```
- CI Job: https://travis-ci.com/facebookincubator/ptr/jobs/247724096

What is there an expected timeframe for 3.8 support? I will have the tool not run `pyre` for now in >=3.8. What's the thoughts on possibly adding a section to the README explaining the 3.8 support plans with 3.8 being an official version now?

DragonMinded:
This is on my list of TODOs underneath some deprecation tasks and continued work on `matchers` API. I don't have any real timeline for supporting this but I can certainly bump priority since there is demand for it.

Support for 3.8 comes in two flavors, however: First is the ability to execute LibCST on 3.8 code. This *should* be available, but you're running into a snag with LibCST's automatic detection of versions. If a python version isn't supplied to the parser, it assumes the version that you're running and since we can't parse 3.8 yet, it fails. Fixing that is possible if you want to hardcode the parser version (which we do in some parser tests). The second flavor is arguably the only useful version, which is the ability to parse 3.8 code, regardless of the currently running python. That's what I assume this feature request is for, and what I can't give an official timeline on yet.

Can you provide your own timeline and such so I can tell how urgently this needs to get addressed?

cooperlees:
That's fine. Long as it's planned :) I have disabled it in ptr for >= 3.8 for now. My PR is linked above to do this. I'll keep an eye and when this is closed I'll undo the disable :)

DragonMinded:
Not sure why mentioning this in a comment auto-closed on merge, but this is not finished.

bgw:
> Not sure why mentioning this in a comment auto-closed on merge

:laughing: 

![Screenshot from 2019-10-28 17-20-48](https://user-images.githubusercontent.com/180404/67727636-6b814780-f9a7-11e9-9f7c-e0b249dcc4f1.png)

Certain keywords, like "closes" and "fixes" can be used to auto-close issues, however github isn't context sensitive. https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

dseevr:
[...]

PR #207, opened 2020-01-13
Add posonly parameter support to Parameters node

DragonMinded:
## Summary

This is part one of a two-part change to support the posonly param indicator "/". After this and the next PR, we will fully suppotr Python 3.8 syntax. This contributes to #122 

## Test Plan

pyre, added new tests, existing tests.


DragonMinded:
Untested code sections will be tested in part 2 when I introduce parsing.

libcst/_nodes/expression.py
```diff
@@ -1670,6 +1698,14 @@ class Parameters(CSTNode):
     #: Optional parameter that captures unspecified kwargs.
     star_kwarg: Optional[Param] = None
 
+    #: Positional-only parameters, with or without defaults. Positional-only
+    #: parameters with defaults must all be after those without defaults.
+    posonly_params: Sequence[Param] = ()
```
jimmylai: The Parameters class becomes fairly complex after this change. It worth provide some examples with mixed usage of those different type of parameter.

libcst/_nodes/expression.py
```diff
@@ -1550,6 +1550,34 @@ def _codegen_impl(self, state: CodegenState) -> None:
         self.comma._codegen(state)
 
 
+@add_slots
+@dataclass(frozen=True)
+class ParamSlash(CSTNode):
+    """
+    A sentinel indicator on a :class:`Parameter` list to denote that the previous
```
jimmylai: Parameter is not a valid class name in LibCST. You probably mean Param.

libcst/_nodes/tests/test_funcdef.py
```diff
@@ -365,6 +431,54 @@ class FunctionDefCreationTest(CSTNodeTest):
...
+                "node": cst.FunctionDef(
+                    cst.Name("foo"),
+                    cst.Parameters(
+                        posonly_params=(cst.Param(cst.Name("bar")),),
+                        star_kwarg=cst.Param(cst.Name("baz")),
+                    ),
+                    cst.SimpleStatementSuite((cst.Pass(),)),
+                ),
+                "code": "def foo(bar, /, **baz): pass\n",
+            },
```
jimmylai: Can you add more advanced test case like `def name(p1, p2=None, /, p_or_kw=None, *, kw):`?

DragonMinded: We don't need to, we have plenty of tests that cover a host of permutations, we don't need to keep doubling the test count every time we add a new feature.

PR #213, opened 2020-01-16
Support `f{x=}` debugging syntax

DragonMinded:
## Summary

This was snuck into 3.8 without any updates to the documentation or grammar. However, it works in python3.8 and so we need to support it as well.

## Test Plan

Verified against several tests in the stdlib, tried a variety of f-strings at the 3.8 REPL and verified that we could parse the same, pyre, new tests, existing tests.


libcst/_nodes/expression.py
```diff
@@ -683,6 +683,10 @@ class FormattedStringExpression(BaseFormattedStringContent):
         ""
     )
 
+    #: Equal sign for formatted string expression uses self-documenting expressions,
+    #: such as ``f"{x=}"``.
```
jimmylai: Can you link to CPython self-documenting expressions doc?

PR #217, opened 2020-01-18
Refactor string split helper method into core attributes.

DragonMinded:
## Summary

We have a helper in one of our codemods that should really be universally available. So, move the relevant bits into libcst core, remove them from the codemod, update the codemod to use the new behavior and update the tests accordingly.

## Test Plan

tox, pyre, new tests

### DIFF
--- a/libcst/_nodes/expression.py
+++ b/libcst/_nodes/expression.py
@@ -516,9 +516,27 @@ class BaseString(BaseExpression, ABC):
     pass
 
 
+StringQuoteLiteral = Literal['"', "'", '"""', "'''"]
+
+
 class _BasePrefixedString(BaseString, ABC):
-    @abstractmethod
-    def _get_prefix(self) -> str:
+    @property
+    def prefix(self) -> str:
+        """
+        Returns the string's prefix, if any exists.
+
+        See `String and Bytes literals
+        <https://docs.python.org/3.7/reference/lexical_analysis.html#string-and-bytes-literals>`_
+        for more information.
+        """
+        ...
+
+    @property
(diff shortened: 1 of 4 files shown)


libcst/_nodes/expression.py
```diff
@@ -517,8 +517,23 @@ class BaseString(BaseExpression, ABC):
...
+        Returns the string's prefix, if any exists.
+
+        See `String and Bytes literals
+        <https://docs.python.org/3.7/reference/lexical_analysis.html#string-and-bytes-literals>`_
+        for more information.
+        """
+        ...
+
+    @property
+    def quote(self) -> str:
```
jimmylai: Can we use Literal to be specific?
```suggestion
    def quote(self) -> Literal['"', "'", ...]:
```

DragonMinded: I think that will fix typing errors when sending the result back to a new string/f-string construction too.

libcst/_nodes/expression.py
```diff
@@ -516,9 +516,27 @@ class BaseString(BaseExpression, ABC):
...
 
 
+StringQuoteLiteral = Literal['"', "'", '"""', "'''"]
+
+
 class _BasePrefixedString(BaseString, ABC):
-    @abstractmethod
-    def _get_prefix(self) -> str:
+    @property
+    def prefix(self) -> str:
```
jimmylai: Can we also use Literal to type prefix? It's also a limited set of strings.

DragonMinded: Literal makes less sense here, as prefix validity is version-dependent.

PR #218, opened 2020-01-21
Full Repo Metadata support in Codemod Tool

DragonMinded:
## Summary

This makes it such that when you write a codemod, you need only to depend on the `TypeInferenceProvider` in your codemod, the way you would a non-full-repo metadata provider. `libcst.tool` will handle the rest, including cache computation and passing values around.

## Test Plan

pyre, existing tests, also installed libcst into a new virtualenv, initialized `libcst.tool` and `pyre` in a new directory, added a file with types and added a codemod that printed the type of each name it encountered. Types were printed as expected.


DragonMinded:
Pyre seems to be failing to install, but it works locally? I retriggered the job and it still fails.

DragonMinded:
This is blocked by https://github.com/facebook/pyre-check/issues/216

PR #220, opened 2020-01-22
Fix list behavior

DragonMinded:
## Summary

If a module is naughty and does work on import, and that work crashes for some reason, we don't want to take down the list command. So, be more permissive in list command exceptions swallowed.

## Test Plan

Tested internally, also existing tests, pyre, tox.


jimmylai:
After catch the exception, can we surface the exception so user can see and fix it instead of just swallow it?

PR #224, opened 2020-01-25
[codemod] add ConvertPercentFormatStringCommand

jimmylai:
## Summary
Port the codemod from lint.
With the prefix, quote, raw_value, evalurated_value helpers provided by SimpleString. The implementation became much simpler now.
Questions to be discussed: (can be addressed by follow up PRs)
- can we rename VisitorBasedCodemodCommand as TransformerBasedCodemodCommand? Visitor implies the ready only visitor pattern and Transformer represents the mutable leave_ API.
- can we consolidate the interface provided by lint framework and codemod framework? That way we can easily convert a codemod as a lint fixer very easily. Currently the conversion wasn't trivial and can be error-prone.
- add a msg as third parameter to assertCodemod? The standard test assert helpers (e.g. assertEqual) takes a msg parameter which can be a useful message when test failed. It's convenient to users used to it if we can provide it.

## Test Plan
Unit tests.

### DIFF
--- a/libcst/codemod/commands/convert_precent_format_to_fstring.py
+++ b/libcst/codemod/commands/convert_precent_format_to_fstring.py
@@ -1,0 +1,126 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+#
+# pyre-strict
+import re
+from typing import Callable, cast
+
+import libcst as cst
+import libcst.matchers as m
+from libcst.codemod import VisitorBasedCodemodCommand
+
+
+USE_FSTRING_SIMPLE_EXPRESSION_MAX_LENGTH = 30
+
+
+def _match_simple_string(node: cst.CSTNode) -> bool:
+    if isinstance(node, cst.SimpleString) and not node.prefix.lower().startswith("b"):
+        # SimpleString can be a bytes and fstring don't support bytes
+        return re.fullmatch("[^%]*(%s[^%]*)+", node.raw_value) is not None
+    return False
+
+
+def _gen_match_simple_expression(module: cst.Module) -> Callable[[cst.CSTNode], bool]:
+    def _match_simple_expression(node: cst.CSTNode) -> bool:
+        # either each element in Tuple is simple expression or the entire expression is simple.
+        if (
+            isinstance(node, cst.Tuple)
+            and all(
+                len(module.code_for_node(elm.value))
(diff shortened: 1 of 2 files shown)


libcst/codemod/commands/convert_precent_format_to_fstring.py
```diff
@@ -0,0 +1,121 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
DragonMinded: Does this need a FB copyright? I have a TODO on my plate to add a copyright/preamble check to the lint job so I don't keep forgetting.

libcst/codemod/commands/convert_precent_format_to_fstring.py
```diff
@@ -0,0 +1,121 @@
...
+
+class EscapeStringQuote(cst.CSTTransformer):
+    def __init__(self, quote: str) -> None:
+        self.quote = quote
+        super().__init__()
+
+    def leave_SimpleString(
+        self, original_node: cst.SimpleString, updated_node: cst.SimpleString
+    ) -> cst.SimpleString:
+        # prefix, quote, innards = _string_prefix_and_quotes(original_node.value)
```
DragonMinded: Can this comment be removed?

libcst/codemod/commands/convert_precent_format_to_fstring.py
```diff
@@ -0,0 +1,121 @@
...
+        if self.quote == original_node.quote:
+            for quo in ["'", '"', "'''", '"""']:
+                if quo != original_node.quote and quo not in original_node.raw_value:
```
DragonMinded: Should we also check the literal value of the string before and after the change, just for safety?

libcst/codemod/commands/convert_precent_format_to_fstring.py
jimmylai: Yes, every Python file needs a copyright unless it's whitelisted.
Add a check in the lint is a good idea for getting feedback early while developing.

DragonMinded: I created #227 for this, since we don't have lint open sourced yet.

libcst/codemod/commands/convert_precent_format_to_fstring.py
jimmylai: Checked and raise exception when the literal value isn't the same.

libcst/codemod/commands/convert_precent_format_to_fstring.py
jimmylai: Removed.

PR #229, opened 2020-01-31
Add "evaluated_value" support to number types

DragonMinded:
## Summary

Extend the support for `evaluated_value` from just strings to numbers as well. Follow up with a few tweaks to code that used `literal_eval` and swap it over to using the calculated properties. Update the docs to encourage using the properties instead of `literal_eval` as well.

## Test Plan

new and existing tests, pyre, tox


DragonMinded:
Coverage looks funny because it turns out that coverage is not picking up the libcst.helpers.tests directory on account of a missing `__init__.py` file. I fixed that in #228 but I didn't want to base this PR on top of that, so it'll get fixed once that lands.

PR #231, opened 2020-02-03
add get_full_name_for_node_or_raise helper

jimmylai:
## Summary
add get_full_name_for_node_or_raise helpe. It's more convenient for some callsites.

## Test Plan
unit test.

### DIFF
--- a/libcst/helpers/expression.py
+++ b/libcst/helpers/expression.py
@@ -30,3 +30,14 @@ def get_full_name_for_node(node: Union[str, cst.CSTNode]) -> Optional[str]:
     elif isinstance(node, cst.Decorator):
         return get_full_name_for_node(node.decorator)
     return None
+
+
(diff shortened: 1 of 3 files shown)


libcst/helpers/expression.py
```diff
@@ -27,3 +27,23 @@ def get_full_name_for_node(node: Union[str, cst.CSTNode]) -> Optional[str]:
     elif isinstance(node, (cst.FunctionDef, cst.ClassDef)):
         return get_full_name_for_node(node.name)
     return None
+
+
+def get_full_name_for_node_or_raise(node: Union[str, cst.CSTNode]) -> str:
```
DragonMinded: This should probably just call `get_full_name_for_node` and then raise if it gets `None` back. That way fixes to the original function (such as my PR up now) won't get lost.

PR #234, opened 2020-02-06
Add support for current package to codemod context.

DragonMinded:
## Summary

I want to add helpers to import nodes and in helper directory for getting the module/object/alias for an import in a way compatible with add/remove import transforms. Part of that includes support for relative imports. Until now, we had no way of figuring out the absolute package name from a relative one. So, add that support in preparation of closing that hole in LibCST.

## Test Plan

Added some tests for the new helper method, also modified the noop codemod to print the package name of each file and looked it over for irregularities.

### DIFF
--- a/libcst/codemod/_context.py
+++ b/libcst/codemod/_context.py
@@ -39,6 +39,12 @@ class CodemodContext:
     #: running codemods from the command line.
     filename: Optional[str] = None
 
+    #: The current module if a codemod is being executed against a file that
+    #: lives on disk, and the repository root is correctly configured. This
+    #: Will take the form of a dotted name such as ``foo.bar.baz`` for a file
+    #: in the repo named ``foo/bar/baz.py``.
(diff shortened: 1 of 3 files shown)


libcst/codemod/_context.py
```diff
@@ -39,6 +39,12 @@ class CodemodContext:
     #: running codemods from the command line.
     filename: Optional[str] = None
 
+    #: The current package if a codemod is being executed against a file that
+    #: lives on disk, and the repository root is correctly configured. This
+    #: Will take the form of a dotted name such as ``foo.bar.baz`` for a file
+    #: in the repo named ``foo/bar/baz.py``.
+    package: Optional[str] = None
```
jimmylai: I'm worrying about this name is confusing. `package` means directory with `__init__.py` file in Python.
`full_module_name` might be a better name here.

DragonMinded: Updated.

PR #239, opened 2020-02-12
Open-Source Import Removal Code

DragonMinded:
## Summary

This was the last bit of codemod framework code that was incubated inside Instagram. This PR open-sources that code and improves it a bit by using the recently-added import helpers. This enables the existing code to handle relative imports successfully, and integrates removing imports into the codemod framework in the same manner as adding imports. With this PR in place, we have a consistent story around imports for everything except for references from type hints that are written as forward reference strings.

This PR copies the files to open-source, rewrites a few sections to use `libcst.helpers`, adds a ton of documentation and adds tests for relative imports since that is enabled by using `libcst.helpers`.

## Test Plan

Existing and new tests, pyre, tox.

### DIFF
--- a/libcst/codemod/visitors/_add_imports.py
+++ b/libcst/codemod/visitors/_add_imports.py
@@ -19,9 +19,10 @@ class AddImportsVisitor(ContextAwareTransformer):
     """
     Ensures that given imports exist in a module. Given a
     :class:`~libcst.codemod.CodemodContext` and a sequence of tuples specifying
-    a module to import from as a string and optionally an object to import from
-    that module, ensures that that import exists. It will modify existing imports
-    as necessary if the module in question is already being imported from.
+    a module to import from as a string. Optionally an object to import from
+    that module and any alias to assign that import, ensures that that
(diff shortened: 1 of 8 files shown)


libcst/codemod/visitors/_add_imports.py
```diff
@@ -19,17 +19,19 @@ class AddImportsVisitor(ContextAwareTransformer):
     """
     Ensures that given imports exist in a module. Given a
     :class:`~libcst.codemod.CodemodContext` and a sequence of tuples specifying
-    a module to import from as a string and optionally an object to import from
-    that module, ensures that that import exists. It will modify existing imports
-    as necessary if the module in question is already being imported from.
+    a module to import from as a string. optionally an object to import from
```
miggaiowski: Should "optionally" be capitalized?
```suggestion
    a module to import from as a string. Optionally an object to import from
```

DragonMinded: Oh yeah, good catch~

PR #246, opened 2020-02-28
Optimize Scope Analysis memory usage by removing unnecessary Assignment creation

jimmylai:
## Summary
In one of the use case, **the Scope Analysis consumed 6GB memory**.

The use case has lots of object creation (Call node) using keyword argument.

`Arg.keyword` is created as an `Assignment` and stored in the parent scope which doesn't make sense.

E.g. An "arg" `Assignment` was created and stored in `GlobalScope`.
```
call(arg=val)
```

Unless we create a `LocalScope` for `Call` node, no `Assignment` should be created in this case.

Adding a `LocalScope` for `Call` node probably won't be very useful since there won't be references.

For now, we just remove the `Assignment` creation.

This reduces memory usage from 6GB to 75MB and makes it **80X more memory** efficient for this specific benchmark.

Also include another small optimization which add slots to `Access`.

CC @zsol

## Test Plan
Added test case and it only pass with the new change.

Generate a Python file with lots of keyword args:
```
lines = []
for i in range(1000):
    offset = i * 1000
    args = ", ".join([f"arg{offset+j}=val{offset+j}" for j in range(100)])
    lines.append(f"func({args})\n")

with open("generated_test_file_massive_keyword_arg_calls.py", "w") as fp:
    fp.writelines(lines)
```
Benchmark script: `bench.py`
```
import libcst as cst
from libcst.metadata import MetadataWrapper, ScopeProvider


txt = open(
    "generated_test_file_massive_keyword_arg_calls.py"
).read()
m = cst.parse_module(txt)
r = MetadataWrapper(m, unsafe_skip_copy=True)
scopes = r.resolve(ScopeProvider)
```
Benchmark command: `time -f "%M" python bench.py`
Before this optimization: 470800
After this optimization: 332496
Memory usage -29.3%

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -556,6 +556,8 @@ class ComprehensionScope(LocalScope):
 
 
 class ScopeVisitor(cst.CSTVisitor):
+    # TODO: Don't provide scope for formatting nodes (semicolon, which space, etc.)
+    # since it's probably not useful. That can makes this visitor cleaner.
     def __init__(self, provider: "ScopeProvider") -> None:
         self.provider: ScopeProvider = provider
         self.scope: Scope = GlobalScope()
@@ -700,7 +702,7 @@ class ScopeVisitor(cst.CSTVisitor):
                 attr.visit(self)
         return False
 
-    def visit_Arg(self, node: cst.Arg) -> Optional[bool]:
+    def visit_Arg(self, node: cst.Arg) -> bool:
         # The keyword of Arg is neither an Assignment nor an Access and we explicitly don't visit it.
         for attr in [
             node.value,
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -36,6 +36,8 @@ from libcst.metadata.expression_context_provider import (
 )
 
 
+@add_slots
+@dataclass(frozen=False)
 class Access:
     """
     An Access records an access of an assignment.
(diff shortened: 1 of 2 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -698,10 +700,19 @@ def visit_Param(self, node: cst.Param) -> Optional[bool]:
...
+    def visit_Arg(self, node: cst.Arg) -> Optional[bool]:
+        # The keyword of Arg is neither an Assignment nor an Access and we explicitly don't visit it.
+        for attr in [
+            node.value,
+            node.equal,
+            node.comma,
+            node.star,
+            node.whitespace_after_star,
+            node.whitespace_after_arg,
+        ]:
```
zsol: I don't really understand why this visits everything other than `node.value`, there can't possibly be assignments and accesses in the rest of them. But then again, the rest of the methods in this class do the same so 🤷

libcst/metadata/scope_provider.py
```diff
@@ -698,10 +700,19 @@ def visit_Param(self, node: cst.Param) -> Optional[bool]:
                 attr.visit(self)
         return False
 
-    def visit_Arg_keyword(self, node: cst.Arg) -> None:
-        keyword = node.keyword
-        if keyword is not None:
-            self.scope.record_assignment(keyword.value, node)
+    def visit_Arg(self, node: cst.Arg) -> Optional[bool]:
```
zsol: nit:
```suggestion
    def visit_Arg(self, node: cst.Arg) -> bool:
```

thatch:
The change looks fine; can you reproduce the memory reduction on a public file like one of libcst itself?

libcst/metadata/scope_provider.py
jimmylai: Correct, we don't expect to find Name in those attributes.
Initially, when ScopeProvide was implemented, the goal was to provide a scope as a container for all nodes in the syntax tree. There is a trick in on_leave that we set the scope object to a node automatically when it's visited and on leave.
https://github.com/Instagram/LibCST/blob/8c77811d701e8c8d58a82f62acd8e08d9bfd7048/libcst/metadata/scope_provider.py#L846-L847
By returning a False in a visit function, we stop the traversal of a subtree and that'll cause some node doesn't have a scope associated.

When we look back to review this design, it probably isn't necessary to provide a scope for a formatting node (e.g. comma, white space, comment.) The use cases are most likely just need the scope object from a non-formatting node. I can add a TODO here to clean up scope node associates to formatting for follow up task.

jimmylai:
> The change looks fine; can you reproduce the memory reduction on a public file like one of libcst itself?

I've included a reproducible benchmark in the test plan. Didn't find good file to test in libcst. Generated one with many keyword arg calls. The result showed memory was reduced by 29%.

PR #247, opened 2020-02-28
Make get_qualified_names_for and __contains__ not change the scope

zsol:
## Summary
Calling `get_qualified_names_for` or `__contains__` for items not in the scope used to result in a `self._assignments[name]` call which inserted a new empty set into itself.

This was a problem because all iterators to assignments were invalidated. This PR prevents the above from happening.

## Test Plan
Added unit tests


zsol:
conflicts resolved

PR #252, opened 2020-03-03
[ScopeProvider][optimization] don't visit formatting information nodes

jimmylai:
## Summary
Don't visit formatting information nodes.
Provide scope object for formatting information is not useful. In most cases, we just need scope for Name, Attribute, etc.
Remove those unnecessary visits makes the visitor run faster and keeps the code cleaner.
CC @zsol 

## Test Plan
Existing unit tests.

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -556,7 +556,6 @@ class ComprehensionScope(LocalScope):
 
 
 class ScopeVisitor(cst.CSTVisitor):
-    # TODO: Don't provide scope for formatting nodes (semicolon, which space, etc.)
     # since it's probably not useful. That can makes this visitor cleaner.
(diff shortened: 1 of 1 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -868,6 +773,8 @@ class ScopeProvider(BatchableMetadataProvider[Optional[Scope]]):
     more advanced static analysis. E.g. given a :class:`~libcst.FunctionDef`
     node, we can check the type of its Scope to figure out whether it is a class method
     (:class:`ClassScope`) or a regular function (:class:`GlobalScope`).
+    Scope metadata is available for most node types other than formatting information nodes
```
thatch: Consider adding a blank line above so this starts a paragraph.

jimmylai: Fixed.

PR #253, opened 2020-03-03
[ScopeProvider][optimization] use batch set union in infer_accesses

jimmylai:
## Summary
Aggregate access with the same name and batch add with set union as an optimization.
In worst case, all accesses (m) and assignments (n) refer to the same name,
the time complexity is O(m x n), this optimizes it as O(m + n).

This also changed the return of scope assignments from Tuple to Set to make it faster without creating Tuple.
Checked current use cases and all of them just iterate on the returned assignments which don't explicit use it as a Tuple. This change is safe.

@zsol 

## Test Plan
Existing unit tests.

Test on a file with 70k lines with 50k access of the same name. Combine both the optimization of not visiting formatting info and this infer access optimization.
Before the optimization: `7min 49.90secs`
After both optimization: `2min 25.19secs` (3.2X faster)

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -556,7 +556,6 @@ class ComprehensionScope(LocalScope):
 
 
 class ScopeVisitor(cst.CSTVisitor):
-    # TODO: Don't provide scope for formatting nodes (semicolon, which space, etc.)
     # since it's probably not useful. That can makes this visitor cleaner.
     def __init__(self, provider: "ScopeProvider") -> None:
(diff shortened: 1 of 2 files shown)


jimmylai:
> I'm not wild about making a type change (or even changing the ordering) in a minor version, but this does seem relatively minor. My other code also just iterates the returned value.

Yeah, I was also thinking about whether to make a major version release for this. After checking all our internal use cases, I cannot find any case that assume the assignments as a Tuple. So I'll just include this in a minor version release.

PR #254, opened 2020-03-04
[helper] add get_docstring() to Module, ClassDef and FunctionDef

jimmylai:
## Summary
Add `get_docstring()` to `Module`, `ClassDef` and `FunctionDef`.
Docstring is the first expression (ignoring comments and empty lines) of `Module`, `ClassDef` or `FunctionDef` and it needs to be a `SimpleString` or `ConcatenatedString`.
Note that `ConcatenatedString` can be a tree of nested `ConcatenatedString`. So we need to traverse the entire tree. It returns `None` if any part of the tree is a f-string.
The reason why docstring cannot be a f-string was discussed in https://bugs.python.org/issue28739.

References:
- `ast.get_docstring` https://docs.python.org/3/library/ast.html#ast.get_docstring
    and https://github.com/python/cpython/blob/89aa4694fc8c6d190325ef8ed6ce6a6b8efb3e50/Lib/ast.py#L254
- PEP 257 https://www.python.org/dev/peps/pep-0257/

## Test Plan
Unit tests.

### DIFF
--- a/libcst/_nodes/expression.py
+++ b/libcst/_nodes/expression.py
@@ -1020,6 +1020,23 @@ class ConcatenatedString(BaseString):
             self.whitespace_between._codegen(state)
             self.right._codegen(state)
 
+    @property
+    def evaluated_value(self) -> Optional[str]:
+        """
+        Return an :func:`ast.literal_eval` evaluated str of recursively concatenated :py:attr:`left` and :py:attr:`right`
+        if and only if both :py:attr:`left` and :py:attr:`right` are composed by :class:`SimpleString` or :class:`ConcatenatedString`
+        (:class:`FormattedString` cannot be evaluated).
+        """
+        left = self.left
+        right = self.right
+        if isinstance(left, FormattedString) or isinstance(right, FormattedString):
+            return None
+        left_val = left.evaluated_value
+        right_val = right.evaluated_value
+        if right_val is None:
+            return None
+        return left_val + right_val
+
 
 @add_slots
 @dataclass(frozen=True)
--- a/libcst/_nodes/statement.py
+++ b/libcst/_nodes/statement.py
@@ -3,6 +3,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import inspect
 import re
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
@@ -20,6 +21,7 @@ from libcst._nodes.expression import (
     BaseDelTargetExpression,
     BaseExpression,
     Call,
+    ConcatenatedString,
     ExpressionPosition,
     From,
     LeftParen,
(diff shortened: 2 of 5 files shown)


libcst/_nodes/expression.py
```diff
@@ -1020,6 +1020,23 @@ def _codegen_impl(self, state: CodegenState) -> None:
...
+        if and only if both :py:attr:`left` and :py:attr:`right` are composed by :class:`SimpleString` or :class:`ConcatenatedString`
+        (:class:`FormattedString` cannot be evaluated).
+        """
+        left = self.left
+        right = self.right
+        if isinstance(left, FormattedString) or isinstance(right, FormattedString):
+            return None
+        left_val = left.evaluated_value
+        right_val = right.evaluated_value
+        if right_val is None:
```
thatch: Why is this check only for right_val?

libcst/_nodes/statement.py
```diff
@@ -1472,6 +1475,40 @@ def _codegen_impl(self, state: CodegenState) -> None:
...
+        if len(expr.body) == 0:
+            return None
+        expr = expr.body[0]
+    if not isinstance(expr, Expr):
+        return None
+    val = expr.value
+    if isinstance(val, (SimpleString, ConcatenatedString)):
+        evaluated_value = val.evaluated_value
+    else:
+        return None
```
thatch: Technically grouping parens are allowed, I'm not sure this handles that case.

```
>>> def f():
...     ("doc")
... 
>>> f.__doc__
'doc'
```

libcst/_nodes/expression.py
jimmylai: https://libcst.readthedocs.io/en/latest/nodes.html#libcst.ConcatenatedString
`left` can be either SimpleString or FormattedString, we just check it's not an FormattedString, so it must be a SimpleString. SimpleString.evaluated_value is always a str.

The `right` part can be a ConcatenatedString, which can have None as the evaluated_value whenever it contains any FormattedString. So we only apply None check on `right` here.

libcst/_nodes/statement.py
jimmylai: Yes, it works since the parentheses are formatting data which parsed as attribute of the SimpleString. See the paste for more detail:

```
SimpleStatementLine(
                body=[
                    Expr(
                        value=SimpleString(
                            value='"doc"',
                            lpar=[
                                LeftParen(
                                    whitespace_after=SimpleWhitespace(
                                        value='',
                                    ),
                                ),
                            ],
                            rpar=[
                                RightParen(
                                    whitespace_before=SimpleWhitespace(
                                        value='',
                                    ),
                                ),
                            ],
                        ),
                        semicolon=MaybeSentinel.DEFAULT,
                    ),
                ],
```


Also add it as a test case.

PR #256, opened 2020-03-05
Better validate `except` without a space after it.

thatch:
## Summary

Testing libcst against real-world code, this comes up a few times, probably from people used to C.  Instead of blanket requiring a space, only require the space if the type is a Name.

There are still many ways to make an invalid `except`, but this finds the most common.

## Test Plan

Included tests fail before the fix, pass after.  There is a `test_invalid` that ensures the exception is still raised on the most obvious cases.

### DIFF
--- a/libcst/_nodes/statement.py
+++ b/libcst/_nodes/statement.py
@@ -785,10 +785,15 @@ class ExceptHandler(CSTNode):
             raise CSTValidationError(
                 "Must use a Name node for AsName name inside ExceptHandler."
             )
-        if self.type is not None and self.whitespace_after_except.empty:
-            raise CSTValidationError(
-                "Must have at least one space after except when ExceptHandler has a type."
-            )
+        type_ = self.type
+        if type_ is not None and self.whitespace_after_except.empty:
+            # Space is only required when the first char in `type` could start
+            # an identifier.  In the most common cases, we want to allow
+            # grouping or tuple parens.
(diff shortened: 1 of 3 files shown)


libcst/_nodes/statement.py
```diff
@@ -786,9 +786,13 @@ def _validate(self) -> None:
                 "Must use a Name node for AsName name inside ExceptHandler."
             )
         if self.type is not None and self.whitespace_after_except.empty:
-            raise CSTValidationError(
-                "Must have at least one space after except when ExceptHandler has a type."
-            )
+            # Space is only required when the first char in `type` could start
+            # an identifier.  In the most common cases, we want to allow
+            # grouping or tuple parens.
+            if isinstance(self.type, Name) and not self.type.lpar:
```
zsol: you'll have to extract `self.type` to a local variable to make the type checker happy.

thatch:
> So as far as I understand, it's no longer possible to trigger this validationerror after parsing well-formed python code

Yes, I think that's correct.  In any case, this switches it to being default-allow if there's a case I missed, which would allow us to roundtrip, instead of failing to parse.

jimmylai:
LGTM
The test cases are well written that every new lines are covered. https://codecov.io/github/Instagram/LibCST/commit/d9f0666ae9388b0fd92d75dc39976337be46c7be

PR #257, opened 2020-03-05
Improve KeyError message for missing metadata

thatch:
## Summary

Adds a more useful message to KeyError when a metadata provider is missing.  This is easy to trigger if you call `.resolve` and let it finish before accessing (because it clears) or if you forget to `.visit` on the MetadataWrapper instead of Module.

## Test Plan

New test.


jimmylai:
Thanks for improving the exception message. I also got confused by the `KeyError` a while ago.

Issue #184, opened 2020-01-04
Compatibility with older python syntax


thatch:
Because I've noticed a couple of references in other projects' bugs: I'm already working on this and don't anticipate that it will be super difficult.  I will be adding support back to 2.7 (but probably stopping there).  I have future-import parsing working as well, which is necessary for print statements and barry_as_FLUFL handling.

https://github.com/thatch/python-grammar-changes has a hand-compiled list of what changed, but I hope to create snippets and mechanically verify what versions they work on as well.

zsol:
I think supporting Python 2 syntax is more burden than it's worth at this point, so @thatch unless you have a compelling argument for this work, I'd propose closing this issue and dropping #275

PR #261, opened 2020-03-09
Support Python 3 syntax back to 3.0

thatch:
## Summary

This adds support for machine-readable supported versions, and the minor changes necessary to support 3.0, 3.1, and 3.3.  This includes future detection for future Python 2 parsing, but isn't required here given the PEP 401 findings.

Machine-readable versions are at `libcst.KNOWN_PYTHON_VERSION_STRINGS`.

## Test Plan

New included tests, and ran against my `python-grammar-changes` project.  Each entry should either have ".." or "oo" to say that the LibCST success matches that cpython version.

```
[tim@rygel python-grammar-changes]$ ../LibCST/.tox/py37/bin/python metarun.py
                                                  30 31 33 35 36 37 38
examples/py31-0002-with1.py                       oo oo oo oo oo oo oo 
examples/py31-0002-with2.py                       .. oo oo oo oo oo oo 
examples/py31-0002-with3.py                       .. oo oo oo oo oo oo 
examples/py33-0011-yield-from.py                  .. .. oo oo oo oo oo 
examples/py33-0012-uprefix1.py                    .. .. oo oo oo oo oo 
examples/py35-0008-else-in-call1.py               oo oo oo oo oo oo oo 
examples/py35-0008-else-in-call2.py               oo oo oo oo oo oo oo 
examples/py35-0009-async-def.py                   .. .. .. oo oo oo oo 
examples/py35-0010-dict-literals.py               .. .. .. oo oo oo oo 
examples/py36-0005-types1.py                      .. .. .. .. oo oo oo 
examples/py36-0006-fstrings.py                    .. .. .. .. oo oo oo 
examples/py36-0007-trailing-comma-def.py          .. .. .. .. oo oo oo 
examples/py36-0028-variable-annotations-pep526.py .. .. .. .. oo oo oo 
examples/py38-0003-walrus1.py                     .. .. .. .. .. .. oo 
examples/py38-0003-walrus2.py                     .. .. .. .. .. .. oo 
examples/py38-0003-walrus3.py                     .. .. .. .. .. .. .. 
examples/py38-0003-walrus4.py                     .. .. .. .. .. .. oo 
Legend:
 green header means will test with libcst

 first result is python, second [optional] result is libcst
 o   parses
 .   does not parse
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,8 +33,8 @@ A Concrete Syntax Tree (CST) parser and serializer library for Python
 
 .. intro-start
 
-LibCST parses Python 3.5, 3.6, 3.7 or 3.8 source code as a CST tree that keeps all
-formatting details (comments, whitespaces, parentheses, etc). It's useful for
+LibCST parses Python 3.0, 3.1, 3.3, 3.5, 3.6, 3.7 or 3.8 source code as a CST tree that keeps
+all formatting details (comments, whitespaces, parentheses, etc). It's useful for
 building automated refactoring (codemod) applications and linters.
 
 .. intro-end
--- a/libcst/__init__.py
+++ b/libcst/__init__.py
@@ -185,7 +185,10 @@ from libcst._nodes.whitespace import (
     TrailingWhitespace,
 )
 from libcst._parser.entrypoints import parse_expression, parse_module, parse_statement
-from libcst._parser.types.config import PartialParserConfig
+from libcst._parser.types.config import (
+    KNOWN_PYTHON_VERSION_STRINGS,
+    PartialParserConfig,
+)
 from libcst._removal_sentinel import RemovalSentinel, RemoveFromParent
 from libcst._version import LIBCST_VERSION
 from libcst._visitors import CSTNodeT, CSTTransformer, CSTVisitor, CSTVisitorT
@@ -201,6 +204,7 @@ from libcst.metadata.wrapper import MetadataWrapper
 
 
 __all__ = [
+    "KNOWN_PYTHON_VERSION_STRINGS",
     "LIBCST_VERSION",
     "BatchableCSTVisitor",
     "CSTNodeT",
--- a/libcst/_nodes/tests/base.py
+++ b/libcst/_nodes/tests/base.py
@@ -41,6 +41,20 @@ def _cst_node_equality_func(
         raise AssertionError(f"\n{a!r}\nis not deeply equal to \n{b!r}{suffix}")
 
 
+def parse_expression_as(**config: Any) -> Callable[[str], cst.BaseExpression]:
+    def inner(code: str) -> cst.BaseExpression:
+        return cst.parse_expression(code, config=cst.PartialParserConfig(**config))
+
+    return inner
+
+
+def parse_statement_as(**config: Any) -> Callable[[str], cst.BaseStatement]:
+    def inner(code: str) -> cst.BaseStatement:
+        return cst.parse_statement(code, config=cst.PartialParserConfig(**config))
+
+    return inner
+
+
 # We can't use an ABCMeta here, because of metaclass conflicts
 class CSTNodeTest(UnitTest):
     def setUp(self) -> None:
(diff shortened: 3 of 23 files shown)


thatch:
This is the first half of #184

thatch:
@jimmylai I could use another pair of eyes on the failing pyre test; I do not see what's different about `test_matrix_multiply` vs e.g. `test_atom` which has similar imports and the same `test_versions` signature.

README.rst
```diff
@@ -33,8 +33,8 @@ A Concrete Syntax Tree (CST) parser and serializer library for Python
 
 .. intro-start
 
-LibCST parses Python 3.5, 3.6, 3.7 or 3.8 source code as a CST tree that keeps all
-formatting details (comments, whitespaces, parentheses, etc). It's useful for
+LibCST parses Python 3.0, 3.1, 3.3, 3.5, 3.6, 3.7 or 3.8 source code as a CST tree that keeps
```
zsol: nit: I would reword this to say
```suggestion
LibCST parses source code compatible with all Python versions between 3.0 and 3.8 (inclusive) as a CST tree that keeps
```

jimmylai: Python 3.2 and 3.4 are not mentioned here, assuming they don't have new syntax change?

jimmylai:
Regarding the weird type errors about `parse_expression_as` and `parse_statement_as`, it's a chick and egg situation.
`pyre-check` depends on `libcst`, so when we install pyre in the venv, an older version of libcst is also installed.
When you do `pyre --search-path ${VIRTUAL_ENV}/lib/python*/site-packages check` or `pyre --preserve-pythonpath check`, it read the old libcst and thus don't know the exist of `parse_expression_as` and `parse_statement_as`.

One way to fix this is to use the LibCST source of truth in the local repo.
By doing this, it installs libcst from local git repo.
`pip uninstall libcst`
`pip install -e .`
Then pyre understands new code and don't emit those errors. The missing parameter annotation errors still apply.

README.rst
thatch: @jimmylai there were no grammar changes in 3.2 or 3.4, so I didn't include them (they would be duplicates of the 2.1 and 3.3 grammar if allowed). My intent with KNOWN_PYTHON_VERSION_STRINGS is that you could do

```
for v in reversed(KNOWN_PYTHON_VERSION_STRINGS):
  try:
    mod = parse(...python_version=v)
    return mod
  except ParseError:
    pass
```

to see if it parses on any.  I'm happy to add some helpers here (e.g. if you have a `/usr/bin/pythonx.y` and want to pick an ideal grammar version.

thatch:
These changes should almost work, although CI will claim a failure and I don't think I can do anything about that.  I only have two issues remaining when I run pyre on my machine:

- tokenize.Intnumber etc (there's an entry in stubs/ but https://github.com/python/typeshed/pull/3839 should fix this for everyone)
- argspec kwonlydefaults can be None:

> libcst/codemod/_command.py:169:20 Undefined attribute [16]: Optional type has no attribute `__getitem__`.

jimmylai:
> These changes should almost work, although CI will claim a failure and I don't think I can do anything about that. I only have two issues remaining when I run pyre on my machine:
> 
> * tokenize.Intnumber etc (there's an entry in stubs/ but [python/typeshed#3839](https://github.com/python/typeshed/pull/3839) should fix this for everyone)
> * argspec kwonlydefaults can be None:
> 
> > libcst/codemod/_command.py:169:20 Undefined attribute [16]: Optional type has no attribute `__getitem__`.

I also see the type error in `libcst/codemod/_command.py` on my dev machine but it doesn't show up in CI. Not sure what's the cause but I've been ignoring it for a long time. It could be related to platform if you also develop on a Mac. (@DragonMinded doesn't have this issue.)

jimmylai:
Regarding the tokenize type error. The installed tokenize source code is not typed. When Pyre read the source code, it generates the type error. We used a stub file to help pyre understand the types https://github.com/Instagram/LibCST/blob/master/stubs/tokenize.pyi
I'm working on a fix to teach Pyre to read both the stubs and the site-packages correctly in https://github.com/Instagram/LibCST/pull/262

thatch:
Regarding the tokenize type error. https://github.com/python/typeshed/pull/3839 is merged, you shouldn't need the stub anymore if you `--typeshed=/path/to/checkout` too.

jimmylai:
Looks like Pyre packs the typeshed. We can probably get your fix for free in the next release.

PR #262, opened 2020-03-11
[CI] Use the latest LibCST source code in Pyre check

jimmylai:
## Summary
After we ask pyre to read site-packages, Pyre cannot understand new changes in LibCST repo..
CI installs pyre-check from pip requirements which installs the public version of libcst.
This removes the installed libcst and install from the checked out repo.

## Test Plan
CI build.


thatch:
Can you rename `.pyre_configuration` to `.pyre_configuration.sample` or so, now that it's ignored?  (The circleci doesn't need to save and restore then, just delete the copy after running.)

jimmylai:
@thatch Renamed `.pyre_configuration` as `.pyre_configuration.example` and updated the README accordingly.

PR #265, opened 2020-03-16
Apply type annotations from a stub.

pradeep90:
Basically move the [`apply_annotations`](https://github.com/facebook/pyre-check/blob/7ee7f4b15fc4666371e7538cc44931a298229a2d/client/apply_annotations.py) code from Pyre. I had discussed this with Jimmy Lai. The main motivation for moving it to libcst is that the code deals purely with libcst concepts like `CSTVisitor` and `CSTTransformer` and we want to use it in two separate projects: Pyre and MonkeyType.

This is not implemented as a normal Codemod because, instead of working on just one module, it works on two modules: the stub module from which to collect annotations and the source module to which to apply the annotations.

## Test Plan

Added the tests as well.

### DIFF
--- a/docs/source/codemods.rst
+++ b/docs/source/codemods.rst
@@ -153,3 +153,5 @@ inside codemods. As of now, the list includes the following helpers.
   :exclude-members: CONTEXT_KEY, visit_Module, leave_ImportFrom, leave_Module
 .. autoclass:: libcst.codemod.visitors.RemoveImportsVisitor
   :exclude-members: CONTEXT_KEY, METADATA_DEPENDENCIES, visit_Module, leave_ImportFrom, leave_Import
+.. autoclass:: libcst.codemod.visitors.ApplyTypeAnnotationsVisitor
+  :exclude-members: CONTEXT_KEY, transform_module_impl, visit_ClassDef, visit_Comment, visit_FunctionDef, leave_Assign, leave_ClassDef, leave_FunctionDef, leave_ImportFrom, leave_Module
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -1,0 +1,476 @@
+# Copyright (c) 2016-present, Facebook, Inc.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree
+#
+# pyre-strict
+
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional, Sequence, Set, Tuple, Union
+
+import libcst as cst
+from libcst.codemod._context import CodemodContext
+from libcst.codemod._visitor import ContextAwareTransformer
+from libcst.codemod.visitors._add_imports import AddImportsVisitor
+from libcst.codemod.visitors._gather_imports import GatherImportsVisitor
+from libcst.helpers import get_full_name_for_node
+
+
+def _get_import_alias_names(import_aliases: Sequence[cst.ImportAlias]) -> Set[str]:
+    import_names = set()
+    for imported_name in import_aliases:
+        asname = imported_name.asname
+        if asname:
+            import_names.add(get_full_name_for_node(asname.name))
+        else:
+            import_names.add(get_full_name_for_node(imported_name.name))
+    return import_names
+
+
+def _get_import_names(imports: Sequence[Union[cst.Import, cst.ImportFrom]]) -> Set[str]:
+    import_names = set()
+    for _import in imports:
+        if isinstance(_import, cst.Import):
+            import_names.update(_get_import_alias_names(_import.names))
+        else:
+            names = _import.names
+            if not isinstance(names, cst.ImportStar):
+                import_names.update(_get_import_alias_names(names))
+    return import_names
+
+
+@dataclass(frozen=True)
+class FunctionAnnotation:
+    parameters: cst.Parameters
+    returns: Optional[cst.Annotation]
+
+
+class TypeCollector(cst.CSTVisitor):
+    """
+    Collect type annotations from a stub module.
+    """
+
+    def __init__(self, existing_imports: Set[str], context: CodemodContext) -> None:
+        # Qualifier for storing the canonical name of the current function.
+        self.qualifier: List[str] = []
+        # Store the annotations.
+        self.function_annotations: Dict[str, FunctionAnnotation] = {}
+        self.attribute_annotations: Dict[str, cst.Annotation] = {}
+        self.existing_imports: Set[str] = existing_imports
+        self.class_definitions: Dict[str, cst.ClassDef] = {}
+        self.context = context
+
+    def visit_ClassDef(self, node: cst.ClassDef) -> None:
+        self.qualifier.append(node.name.value)
+        self.class_definitions[node.name.value] = node
+
+    def leave_ClassDef(self, original_node: cst.ClassDef) -> None:
+        self.qualifier.pop()
+
+    def visit_FunctionDef(self, node: cst.FunctionDef) -> bool:
+        self.qualifier.append(node.name.value)
+        returns = node.returns
+        if returns is not None:
+            return_annotation = self._create_import_from_annotation(returns)
+            parameter_annotations = self._import_parameter_annotations(node.params)
+            self.function_annotations[".".join(self.qualifier)] = FunctionAnnotation(
+                parameters=parameter_annotations, returns=return_annotation
+            )
+        # pyi files don't support inner functions, return False to stop the traversal.
+        return False
+
+    def leave_FunctionDef(self, original_node: cst.FunctionDef) -> None:
+        self.qualifier.pop()
+
+    def visit_AnnAssign(self, node: cst.AnnAssign) -> bool:
+        name = get_full_name_for_node(node.target)
+        if name is not None:
+            self.qualifier.append(name)
+        annotation_value = self._create_import_from_annotation(node.annotation)
+        self.attribute_annotations[".".join(self.qualifier)] = annotation_value
+        return True
+
+    def leave_AnnAssign(self, original_node: cst.AnnAssign) -> None:
+        self.qualifier.pop()
+
+    def visit_ImportFrom(self, node: cst.ImportFrom) -> None:
+        module = node.module
+        names = node.names
+
+        # module is None for relative imports like `from .. import foo`.
+        # We ignore these for now.
+        if module is None or isinstance(names, cst.ImportStar):
+            return
+        module_name = get_full_name_for_node(module)
+        if module_name is not None:
+            for import_name in _get_import_alias_names(names):
+                AddImportsVisitor.add_needed_import(
+                    self.context, module_name, import_name
+                )
+
+    def _add_annotation_to_imports(
+        self, annotation: cst.Attribute
+    ) -> Union[cst.Name, cst.Attribute]:
+        key = get_full_name_for_node(annotation.value)
+        if key is not None:
+            # Don't attempt to re-import existing imports.
+            if key in self.existing_imports:
+                return annotation
+            import_name = get_full_name_for_node(annotation.attr)
+            if import_name is not None:
+                AddImportsVisitor.add_needed_import(self.context, key, import_name)
+        return annotation.attr
+
+    def _handle_Index(self, slice: cst.Index, node: cst.Subscript) -> cst.Subscript:
+        value = slice.value
+        if isinstance(value, cst.Subscript):
+            new_slice = slice.with_changes(value=self._handle_Subscript(value))
+            return node.with_changes(slice=new_slice)
+        elif isinstance(value, cst.Attribute):
+            new_slice = slice.with_changes(value=self._add_annotation_to_imports(value))
+            return node.with_changes(slice=new_slice)
+        else:
+            return node
+
+    def _handle_Subscript(self, node: cst.Subscript) -> cst.Subscript:
+        slice = node.slice
+        if isinstance(slice, list):
+            new_slice = []
+            for item in slice:
+                value = item.slice.value
+                if isinstance(value, cst.Attribute):
+                    name = self._add_annotation_to_imports(item.slice.value)
+                    new_index = item.slice.with_changes(value=name)
+                    new_slice.append(item.with_changes(slice=new_index))
+                else:
+                    if isinstance(item.slice, cst.Index) and not isinstance(
+                        item.slice.value, cst.Name
+                    ):
+                        new_index = item.slice.with_changes(
+                            value=self._handle_Index(item.slice, item)
+                        )
+                        item = item.with_changes(slice=new_index, comma=None)
+                    new_slice.append(item)
+            return node.with_changes(slice=new_slice)
+        elif isinstance(slice, cst.Index):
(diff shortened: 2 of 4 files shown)


libcst/codemod/commands/tests/test_apply_stub_annotations.py
```diff
@@ -0,0 +1,679 @@
...
+            parse_module(textwrap.dedent(stub.rstrip())),
+            parse_module(textwrap.dedent(source.rstrip())),
+            textwrap.dedent(expected.rstrip()),
+        )
+
+    def assert_annotations(self, stub: str, source: str, expected: str) -> None:
+        stub_file, source_file, expected = self.format_files(stub, source, expected)
+        self.assertEqual(apply_stub_annotations(stub_file, source_file).code, expected)
+
+    def test_annotate_functions(self) -> None:
```
thatch: Can you break this up with `data_provider`?

libcst/codemod/commands/apply_stub_annotations.py
```diff
@@ -0,0 +1,506 @@
...
+    if attribute is not None:
+        names.append(_get_name_as_string(attribute.value))
+    return ".".join(names)
+
+
+def _get_name_as_string(node: Union[cst.CSTNode, str]) -> str:
+    if isinstance(node, cst.Name):
+        return node.value
+    else:
+        # pyre-fixme[7]: Expected `str` but got `Union[CSTNode, str]`.
```
thatch: Are the pyre-fixme's in the original? I agree with the typechecker on many of them, like this one.

libcst/codemod/commands/apply_stub_annotations.py
```diff
@@ -0,0 +1,506 @@
...
+        # Don't attempt to re-import existing imports.
+        if key in self.existing_imports:
+            return annotation
+        self._add_to_imports(
+            [cst.ImportAlias(name=annotation.attr)], annotation.value, key
+        )
+        return annotation.attr
+
+    @overload
+    def _handle_Index(self, slice: cst.Index, node: cst.Subscript) -> cst.Subscript:
```
thatch: I'm unfamiliar with this code, but I'm lost on the type hints here.  The function body makes assumptions about `node` (namely that it has a `.slice`) without enforcing that.  Is the overload necessary?

libcst/codemod/commands/apply_stub_annotations.py
pradeep90: Yes, they were in the original. Sorry about this. I'm unfamiliar with this code too, but I'll try to fix as many as I can.

libcst/codemod/commands/apply_stub_annotations.py
```diff
@@ -0,0 +1,506 @@
...
+                *statements_before_imports,
+                *toplevel_statements,
+                *statements_after_imports,
+            ]
+        )
+
+
+def apply_stub_annotations(stubs: cst.Module, source: cst.Module) -> cst.Module:
+    import_visitor = ImportCollector()
+    source.visit(import_visitor)
```
jimmylai: We have a helper `GatherImportsVisitor` for collecting imports. Can we reuse it to be simpler? https://libcst.readthedocs.io/en/latest/codemods.html#libcst.codemod.visitors.GatherImportsVisitor

libcst/codemod/commands/apply_stub_annotations.py
```diff
@@ -0,0 +1,506 @@
...
+                    parameter = parameter.with_changes(
+                        annotation=self._create_import_from_annotation(annotation)
+                    )
+                updated_parameters.append(parameter)
+            return updated_parameters
+
+        return parameters.with_changes(params=update_annotations(parameters.params))
+
+
+class TypeTransformer(cst.CSTTransformer):
```
jimmylai: Can we make this class a `ContextAwareTransformer` and have the `apply_stub_annotations` as static method?
This is to make it consistent to our other transformer helpers like `AddImportsVisitor` and `RemovedNodeVisitor ` to prevent future confusion.

```
class TypeTransformer(ContextAwareTransformer):
    """ Add detailed documentation. """
    @staticmethod
    def apply_stub_annotations(sub_module: cst.Module):
        # collect stub type annotations using TypeCollector and store the collected data.

```
We can use GatherImportsVisitor to collect imports in `visit_Module` like this:
https://github.com/Instagram/LibCST/blob/9f7018902925f6404f4df8a36045c81cb53424b8/libcst/codemod/visitors/_add_imports.py#L169-L173

Once you finish that, you can add to the library of transform document by updating  https://github.com/Instagram/LibCST/blob/0ff9a6c5b7f2380dfeca4be01644c9d7e9146a30/docs/source/codemods.rst#L142-L150

carljm:
@jimmylai Yes, this will meet MonkeyType's needs! Will let you and @thatch review the details since you've already done thorough reviews.

pradeep90:
+ Make it a `ContextAwareTransformer`, rename it to `ApplyTypeAnnotationsVisitor`, and move it to codemod/visitors/. 

+ Use `GatherImportsVisitor` to gather imports. 

+ Add a static method `add_stub_to_context` so that users can schedule a stub whose types are to be applied later. 

+ Use `assertCodemod` and `data_provider` in the tests. 

+ Add documentation. 

+ Remove fixmes.

docs/source/codemods.rst
```diff
@@ -153,3 +153,5 @@ inside codemods. As of now, the list includes the following helpers.
   :exclude-members: CONTEXT_KEY, visit_Module, leave_ImportFrom, leave_Module
 .. autoclass:: libcst.codemod.visitors.RemoveImportsVisitor
   :exclude-members: CONTEXT_KEY, METADATA_DEPENDENCIES, visit_Module, leave_ImportFrom, leave_Import
+.. autoclass:: libcst.codemod.visitors.ApplyTypeAnnotationsVisitor
```
thatch: @jimmylai  is there a reason these aren't alphabetized?

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -0,0 +1,558 @@
+# Copyright (c) 2016-present, Facebook, Inc.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree
+
+
```
thatch: does this still work if pyre-strict?

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -0,0 +1,558 @@
...
+        asname = imported_name.asname
+        if asname:
+            import_names.add(_get_name_or_attribute_as_string(asname.name))
+        else:
+            import_names.add(_get_name_or_attribute_as_string(imported_name.name))
+    return import_names
+
+
+def get_import_names(imports: Sequence[Union[cst.Import, cst.ImportFrom]]) -> Set[str]:
+    import_names = set()
```
thatch: I'm a little surprised this doesn't require an annotation.  Is this because it's not in strict mode?

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -0,0 +1,558 @@
...
+        if isinstance(_import, cst.Import):
+            import_names.update(get_import_alias_names(_import.names))
+        else:
+            names = _import.names
+            if not isinstance(names, cst.ImportStar):
+                import_names.update(get_import_alias_names(names))
+    return import_names
+
+
+class FunctionAnnotation(NamedTuple):
```
thatch: @jimmylai I think the rest uses dataclasses, is there a reason to stick with namedtuple here?

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -0,0 +1,558 @@
...
+        self.attribute_annotations[".".join(self.qualifier)] = annotation_value
+        return True
+
+    def leave_AnnAssign(self, original_node: cst.AnnAssign) -> None:
+        self.qualifier.pop()
+
+    def visit_ImportFrom(self, node: cst.ImportFrom) -> None:
+        module = node.module
+        names = node.names
+        if module is None or isinstance(names, cst.ImportStar):
```
thatch: do you know off the top of your head when module is None?  This could use a comment.

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -0,0 +1,558 @@
...
+            return
+        value = module.value
+        name = (
```
thatch: I'm not sure the types check out here, I think just `name = _get_name_or_attribute_as_string(module)` might work.  See https://libcst.readthedocs.io/en/latest/nodes.html#libcst.ImportFrom

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -0,0 +1,558 @@
...
+from typing import Dict, List, NamedTuple, Optional, Sequence, Set, Tuple, Union
+
+import libcst as cst
+from libcst.codemod._context import CodemodContext
+from libcst.codemod._visitor import ContextAwareTransformer
+from libcst.codemod.visitors._gather_imports import GatherImportsVisitor
+
+
+def _get_name_or_attribute_as_string(
+    name_or_attribute: cst.BaseExpression,
```
thatch: Can you give this a more concrete Union type? As I'm going through the ones below it looks like this might be `Union[cst.Name, cst.Attribute, None]` and anything else is intended as an error?

libcst/codemod/visitors/_apply_type_annotations.py
pradeep90: No, I have to add it. Thanks for catching that.

Side note: I notice that the `.pyre_configuration` file was [recently](https://github.com/Instagram/LibCST/commit/ea8619ade5eecaf02136be1d588ee075145097c8) moved into the .circleci directory. I don't know if that was intentional, but I had to copy-paste the configuration file into the root directory for local testing.

Also, given that almost all the files are marked `pyre-strict`, you could make them strict by default using `"strict": true`.

libcst/codemod/visitors/_apply_type_annotations.py
pradeep90: No, local variables without annotations are fine.

libcst/codemod/visitors/_apply_type_annotations.py
pradeep90: It's None for `from .. import foo`. Added a comment and a test.

libcst/codemod/visitors/_apply_type_annotations.py
pradeep90: It's passed a `BaseExpression` in a few places. For example, `Attribute.value` gives a `BaseExpression`. Given that it's a tradeoff between doing `isinstance` in the caller or returning `Optional` in this function, I feel it's better to return `Optional` since it returns `Optional` for some Attributes anyway.

docs/source/codemods.rst
jimmylai: No, we basically just add them sequentially when we add new visitors.

libcst/codemod/visitors/_apply_type_annotations.py
jimmylai: It's a very common pattern to get the dot-concatenated name given a Name, Attribute or other type of nodes, and we already built a helper for it.
https://libcst.readthedocs.io/en/latest/helpers.html#libcst.helpers.get_full_name_for_node
It returns exactly `Optional[str]`. Please use this one.

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -0,0 +1,563 @@
...
+        if key is not None:
+            # Don't attempt to re-import existing imports.
+            if key in self.existing_imports:
+                return annotation
+            value = annotation.value
+            if isinstance(value, (cst.Attribute, cst.Name)):
+                self._add_to_imports(
+                    [cst.ImportAlias(name=annotation.attr)], value, key
+                )
+        return annotation.attr
```
jimmylai: I'm really sorry that just realized a lot of code here are for handling adding missing import.
Good news is, we already have that implemented as a helper `AddImportsVisitor`.
https://libcst.readthedocs.io/en/latest/codemods.html#libcst.codemod.visitors.AddImportsVisitor

To add `from typing import Optional`,
you just need to call: `AddImportsVisitor.add_needed_import(self.context, "typing", "Optional")`.

To add `import typing`,
you just need to call: `AddImportsVisitor.add_needed_import(self.context, "typing")`.

`AddImportsVisitor` will do the work to make sure there is no duplicated imports for you. So a lot of code in `TypeCollector` and `ApplyTypeAnnotationsVisitor` can be saved. You also don't need to use `GatherImportsVisitor` to collect imports. You can safely add all import needed by type annotation.
I should have point this out sooner.

libcst/codemod/visitors/_apply_type_annotations.py
jimmylai: Yes, it's intentional, we updated our README and ask developer to follow the steps to copy `.pyre_configuration.example` as `.pyre_configuration` as well as adding custom python package path.
Each developer may have a venv in different location and we want to include the venv site-packages path in the config. So Pyre parses all installed Python packages and can understand more types from more packages.
https://github.com/Instagram/LibCST/blob/master/README.rst

pradeep90:
+ Use `get_full_name_for_node`.
+ Use `AddImports` instead of manually adding imports.
+ Remove old import code and ImportStatement.

Note: I discovered a couple of edge cases regarding imports with the same name, which I'll address in a follow-up PR.

libcst/codemod/visitors/tests/test_apply_type_annotations.py
```diff
@@ -0,0 +1,629 @@
...
+                def foo():
+                    return returns_baz()
+                """,
+                """
+                import bar
+
+                def foo() -> bar.Baz:
+                    return returns_baz()
+                """,
+            ),
```
pradeep90: I'm keeping the use of `GatherImports` to get and check against the imports that already exist in the source module, because otherwise it adds an extra import `from bar import Baz` and makes the return type of `foo` be `Baz`.

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -0,0 +1,483 @@
...
+    def leave_Module(
+        self, original_node: cst.Module, updated_node: cst.Module
+    ) -> cst.Module:
+        fresh_class_definitions = [
+            definition
+            for name, definition in self.annotations.class_definitions.items()
+            if name not in self.visited_classes
+        ]
+        if self.is_generated:
+            return original_node
```
jimmylai: `@generated` is FB specific logic which is not generic.
Our CLI includes the option to skip generated file.

https://github.com/Instagram/LibCST/blob/c21ae59a6bb2765d750748418c6bafc4af6cf8da/libcst/tool.py#L481

https://github.com/Instagram/LibCST/blob/c21ae59a6bb2765d750748418c6bafc4af6cf8da/libcst/tool.py#L657-L659

We should probably reuse those existing logic or at least make the current implementation configurable (e.g. add `skip_generated: bool = True` parameter to `add_stub_to_context`).

pradeep90: Ok. I removed our ad hoc handling since it looks like your tool already supports it.

PR #278, opened 2020-03-30
Ignore version markers from pyenv and common virtualenv names

jreese:
## Summary

These are added when adding directory-based version preferences via `pyenv local`. Also, common practice is to put the virtualenvs in the root of the project, named either `venv/` or `.venv/`, so let's ignore those too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ build/
 .coverage
 .hypothesis/
 .pyre_configuration
+.python-version
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ build/
 .hypothesis/
 .pyre_configuration
 .python-version
+.venv/
+venv/


jimmylai:
The issue of having venv in current directory is that: Pyre will type check the code.
That's probably the reason why your Pyre run is much slower. @jreese 
Unless we can confirm having venv in current directory don't cause the issue, I won't suggest add those directories.

jimmylai:
@jreese we recommend use CPython builtin `venv` to create develop environment (rather than `pyenv`). 
https://github.com/Instagram/LibCST/blob/master/README.rst#development

Let me know if you still think this change is needed.

thatch:
pyenv is very convenient for installing arbitrary versions of python.  It does more/different than just venv.

jimmylai:
We recommend venv instead of pyenv in the development process and CI builds.
You're free to use pyenv locally and add the ignore locally in `.git/info/exclude`.
Closing this for now.

PR #284, opened 2020-04-16
[ScopeProvider] Expose more granular Assignments and Accesses for dotted imports

### LINKED ISSUE #281: [RFC] Analyze attribute access from imports to support dead import clean up better
Per @zsol 's request, propose this change. Feedback is welcome.

In this example, the access `a.b` is recorded in assignment `a.b` and `a.c`. So the `RemoveUnusedImportsCommand` cleans up nothing.
```
import a.b
import a.c

a.b
```

It's because we only record the top level name of assignments and accesses.
https://github.com/Instagram/LibCST/blob/master/libcst/metadata/scope_provider.py#L310-L311

For example, when we see an access `a.b`, `a` can be imported or a local object.
To provide the dead import clean up support better, we can record the best assignment and access when multiple imports are available in `infer_access`.

https://github.com/Instagram/LibCST/blob/master/libcst/metadata/scope_provider.py#L786-L800

That means, given an access, we can check all import assignments having the same top level name (`a.b` and `a.c`) and only assign the access when it startswith the fully qualified name of the access.

zsol:
## Summary
This PR adds extra `Assignment`s created by the `ScopeProvider` for dotted imports like `import a.b.c`. For such an import, there are now three assignments, keyed by `a`, `a.b`, and `a.b.c`. For accesses, only the most specific access is recorded and attached to the appropriate assignment.

## Test Plan

Added unit tests.

Fixes #281.

### DIFF
--- a/libcst/codemod/commands/tests/test_remove_unused_imports.py
+++ b/libcst/codemod/commands/tests/test_remove_unused_imports.py
@@ -48,3 +48,32 @@ class RemoveUnusedImportsCommandTest(CodemodTest):
             x: a = 1
         """
         self.assertCodemod(before, before)
+
+    def test_dotted_imports(self) -> None:
+        before = """
+            import a.b, a.b.c
+            import e.f
+            import g.h
+            import x.y, x.y.z
+
+            def foo() -> None:
+                a.b
+                e.g
+                g.h.i
+                x.y.z
+        """
+
+        after = """
+            import a.b, a.b.c
+            import e.f
+            import g.h
+            import x.y.z
+
+            def foo() -> None:
+                a.b
+                e.g
+                g.h.i
+                x.y.z
+        """
+
+        self.assertCodemod(before, after)
--- a/libcst/codemod/visitors/tests/test_remove_imports.py
+++ b/libcst/codemod/visitors/tests/test_remove_imports.py
@@ -377,46 +377,6 @@ class TestRemoveImportsCodemod(CodemodTest):
             context_override=CodemodContext(full_module_name="a.b.foobar"),
         )
 
-    def test_remove_dotted_import(self) -> None:
-        before = """
-            import a.b, a.b.c
-            import e.f
-            import g.h
-            import x.y, x.y.z
-
-            def foo() -> None:
-                a.b
-                e.g
-                g.h.i
-                x.y.z
-        """
-
-        after = """
-            import a.b, a.b.c
-            import e.f
-            import g.h
-            import x.y.z
-
-            def foo() -> None:
-                a.b
-                e.g
-                g.h.i
-                x.y.z
-        """
-
-        self.assertCodemod(
-            before,
-            after,
-            [
-                ("a.b", None, None),
-                ("a.b.c", None, None),
-                ("e.f", None, None),
-                ("g.h", None, None),
-                ("x.y", None, None),
-                ("x.y.z", None, None),
-            ],
-        )
-
     def test_remove_import_complex(self) -> None:
         """
         Should remove complex module as import
--- a/libcst/codemod/visitors/tests/test_remove_imports.py
+++ b/libcst/codemod/visitors/tests/test_remove_imports.py
@@ -433,6 +433,7 @@ class TestRemoveImportsCodemod(CodemodTest):
             from d.e import f
             from h.i import j as k
             from l.m import n as o
+            from x import *
 
             def foo() -> None:
                 f()
@@ -442,6 +443,7 @@ class TestRemoveImportsCodemod(CodemodTest):
             from bar import qux
             from d.e import f
             from h.i import j as k
+            from x import *
 
             def foo() -> None:
                 f()
--- a/libcst/codemod/visitors/tests/test_remove_imports.py
+++ b/libcst/codemod/visitors/tests/test_remove_imports.py
@@ -387,21 +387,25 @@ class TestRemoveImportsCodemod(CodemodTest):
             import baz, qux
             import a.b
             import c.d
+            import x.y.z
             import e.f as g
             import h.i as j
 
             def foo() -> None:
                 c.d()
+                x.u
                 j()
         """
         after = """
             import bar
             import qux
             import c.d
+            import x.y.z
             import h.i as j
 
             def foo() -> None:
                 c.d()
+                x.u
                 j()
         """
 
@@ -414,6 +418,7 @@ class TestRemoveImportsCodemod(CodemodTest):
                 ("c.d", None, None),
                 ("e.f", None, "g"),
                 ("h.i", None, "j"),
+                ("x.y.z", None, None),
             ],
         )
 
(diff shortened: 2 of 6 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -584,12 +587,32 @@ class ComprehensionScope(LocalScope):
...
+        yield (node.value, node)
+    else:
+        value = node.value
+        if not isinstance(value, (cst.Attribute, cst.Name)):
+            raise ValueError(f"Unexpected name value in import: {value}")
+        name_values = iter(_gen_dotted_names(value))
+        (next_name, next_node) = next(name_values)
+        yield (f"{next_name}.{node.attr.value}", node)
+        yield (next_name, next_node)
+        yield from name_values
```
jimmylai, line 606: This is duplicated to `get_full_name_for_node`. Let's reuse the helper.
https://libcst.readthedocs.io/en/latest/helpers.html#libcst.helpers.get_full_name_for_node

jimmylai (changes requested):

libcst/metadata/scope_provider.py
jimmylai: I didn't read it right, so it generates all possible dotted names, e.g. a, a.b, a.b.c
This is specific for Attribute, should we just make it a helper of `Attribute`, so it's easier to use?

libcst/metadata/scope_provider.py
```diff
@@ -788,10 +808,20 @@ def infer_accesses(self) -> None:
...
-            name = access.node.value
+        for (access, enclosing_attribute) in self.__deferred_accesses:
+            if enclosing_attribute is not None:
+                names = _gen_dotted_names(enclosing_attribute)
+                name = None
+                for name, node in names:
+                    if name in access.scope:
+                        access.node = node
+                        break
+                assert name is not None
```
jimmylai, line 819:
```suggestion
                for name, node in _gen_dotted_names(enclosing_attribute):
                    if name in access.scope:
                        access.node = node
                        break
```

libcst/codemod/visitors/tests/test_remove_imports.py
```diff
@@ -377,6 +377,46 @@ def foo() -> None:
...
+            after,
+            [
+                ("a.b", None, None),
+                ("a.b.c", None, None),
+                ("e.f", None, None),
+                ("g.h", None, None),
+                ("x.y", None, None),
+                ("x.y.z", None, None),
+            ],
+        )
```
jimmylai, line 418: If you add this as a test case of `RemoveUnusedImportsCommandTest`, you don't need to manually iterate all the imports.
https://github.com/Instagram/LibCST/blob/master/libcst/codemod/commands/tests/test_remove_unused_imports.py#L12-L18

libcst/metadata/tests/test_scope_provider.py
```diff
@@ -782,7 +817,7 @@ def test_multiple_assignments(self) -> None:
             },
         )
 
-    def test_assignemnts_and_accesses(self) -> None:
+    def test_assignments_and_accesses(self) -> None:
```
jimmylai, line 820: Nice catch on typos.

jimmylai:
This is a pretty good enhancement. Import `a.b.c` not only imports module `a.b.c` but also `a` and `a.b`. It's a new concept added to scope analysis. Can you provide some example and documentation to explain it?
https://libcst.readthedocs.io/en/latest/metadata.html#scope-metadata

libcst/metadata/scope_provider.py
```diff
@@ -54,7 +57,7 @@ def __new__(cls) -> "Tree":
 
     #: The name node of the access. A name is an access when the expression context is
     #: :attr:`ExpressionContext.LOAD`.
-    node: cst.Name
+    node: Union[cst.Name, cst.Attribute]
```
jimmylai, line 60: Can you also update the inline comments to explain in what situation the node will be an `Attribute`?

libcst/metadata/scope_provider.py
jimmylai: To make it clear, my suggestion is:

Add a helper in Attribute and maybe name it as `get_dotted_prefix_names`  to make it clearer. https://github.com/Instagram/LibCST/blob/4544539485c0d5ab8dad776465b9597fb1e16baa/libcst/_nodes/expression.py#L1358
Then when you see a `Name` node in `ScopeVisitor`, you can just do `node.value` to get the string value.
And only do `node.get_dotted_prefix_names` when the node is an `Attribute`.

zsol: Hmm I don't think this helper function makes sense for all `Attribute`s. For example: `a[b].c` or `(a+b).c`.
I think this only makes sense in the context of `Attribute`s inside import-like nodes.

libcst/metadata/scope_provider.py
zsol: That's a type error unfortunately :(
`_gen_dotted_names` might result in an empty generator (ofcourse it doesn't in practice), so the type checker warns about `name` being undefined down below.

libcst/codemod/visitors/tests/test_remove_imports.py
zsol: Good point, I haven't thought about that :)

libcst/metadata/scope_provider.py
zsol: done!

libcst/codemod/commands/tests/test_remove_unused_imports.py
```diff
@@ -48,3 +48,32 @@ def test_type_annotations(self) -> None:
...
+
+            def foo() -> None:
+                a.b
+                e.g
+                g.h.i
+                x.y.z
+        """
+
+        after = """
+            import a.b, a.b.c
```
jimmylai, line 67: Why `a.b.c` is not removed?

jimmylai: I understand a.b.c may have side effect and thus is not safe to remove. My question is how do we not remove it?
The ScopeVisitor looks like will use name `a.b` to store access to assignment `a.b`.
Is it because `import a.b.c` creates three assignment lookup entries (a, a.b, a.b.c), `a.b` access is recorded to the second one so, the import is not removed?
In the case of `x.y.z`, we only associate access `x.y.z` to `import x.y.z` instead of `import x.y` (which are x and x.y).

jimmylai (approved):
Overall looks great.
One last thing, can you add a test case to cover this line?
https://codecov.io/gh/Instagram/LibCST/pull/284/diff?src=pr&el=tree#D2-642...643
Add a `from module import *` in a test case should make it covered.

libcst/codemod/commands/tests/test_remove_unused_imports.py
zsol: It's because `a.b.c` is looked at in isolation (without considering the existence of the import for `a.b`).
So essentially because of the same reason this is not fixed:
```
import a
import a
a
```
This is a limitation of the `RemoveUnusedImportsVisitor`, it only removes unused imports, not duplicates.

PR #291, opened 2020-04-28
[apply-annotations] Add argument for ignoring existing annotations.

pradeep90:
## Summary
This will allow us to override existing types based on the stub.

## Test Plan

Added test case.

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -203,7 +203,7 @@ class ApplyTypeAnnotationsVisitor(ContextAwareTransformer):
     This is one of the transforms that is available automatically to you when
     running a codemod. To use it in this manner, import
     :class:`~libcst.codemod.visitors.ApplyTypeAnnotationsVisitor` and then call the static
-    :meth:`~libcst.codemod.visitors.ApplyTypeAnnotationsVisitor.add_stub_to_context` method,
+    :meth:`~libcst.codemod.visitors.ApplyTypeAnnotationsVisitor.store_stub_in_context` method,
     giving it the current context (found as ``self.context`` for all subclasses of
     :class:`~libcst.codemod.Codemod`), the stub module from which you wish to add annotations.
 
@@ -211,7 +211,7 @@ class ApplyTypeAnnotationsVisitor(ContextAwareTransformer):
 
         stub_module = parse_module("x: int = ...")
 
-        ApplyTypeAnnotationsVisitor.add_stub_to_context(self.context, stub_module)
+        ApplyTypeAnnotationsVisitor.store_stub_in_context(self.context, stub_module)
 
     You can apply the type annotation using::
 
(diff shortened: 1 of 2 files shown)


zsol:
Functionality looks good to me. I have one nit: `ignore_existing_annotations` is not expressive enough IMO. I'd be much happier with something like `overwrite_existing_annotations`. Wdyt?

jimmylai:
> Functionality looks good to me. I have one nit: `ignore_existing_annotations` is not expressive enough IMO. I'd be much happier with something like `overwrite_existing_annotations`. Wdyt?

Agree. `overwrite_existing_annotations` is more direct.

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -223,19 +223,26 @@ class ApplyTypeAnnotationsVisitor(ContextAwareTransformer):
...
     """
 
     CONTEXT_KEY = "ApplyTypeAnnotationsVisitor"
 
     def __init__(
-        self, context: CodemodContext, annotations: Optional[Annotations] = None
+        self,
+        context: CodemodContext,
+        annotations: Optional[Annotations] = None,
+        ignore_existing_annotations: bool = False,
```
jimmylai, line 237: Can we add `overwrite_existing_annotations` to the static function `add_stub_to_context` (we can also do both)?
So overwrite_existing_annotations is stored in the context (probably use a different key).
That can make it easier for user since in some codemod application, user doesn't need to explicitly create the visitor instance and it's done by the high level Codemod helper.
E.g. In this codemod, user just calls the static method  to add needed imports without needed to create `AddImportsVisitor` instance. https://github.com/Instagram/LibCST/blob/379c9d06e766ec48d283bd622d82034c41c178f0/libcst/codemod/commands/strip_strings_from_types.py#L45

pradeep90: Yeah, added the argument to the static method as well. 

Design choice: Setting the argument to True in any call will overwrite annotations for all stubs. The alternative was to store a flag per stub (and thus per annotation), which seemed like overkill.

pradeep90: Found that API a bit confusing since it wasn't clear what would happen if you called `add_stub_to_context(stub1, overwrite_existing_annotations=True)` and `add_stub_to_context(stub2, overwrite_existing_annotations=False)`.

Decided to go with a separate static method `set_overwrite_existing_annotations_flag_in_context` such that calling it at any point will set the flag to True. (It does not accept a bool flag to avoid the ambiguity I mentioned above with multiple calls.)

jimmylai: Do you expect user calls `ApplyTypeAnnotationsVisitor.add_stub_to_context` multiple times?
The function takes a stub module and apply the stub on current module. When you run over a list of stub files, you'll need to call `add_stub_to_context` with different stub module and context. Separate `set_overwrite_existing_annotations_flag_in_context` as another flag adds one more step to config the overwrite which doesn't seem make it easier to use due to the overwrite flag is stored in the context. That means each module needs to call `set_overwrite_existing_annotations_flag_in_context` separately if they want to enable overwrite.

pradeep90: > Do you expect user calls ApplyTypeAnnotationsVisitor.add_stub_to_context multiple times?

Yes, as it stands, `add_stub_to_context` can be called multiple times to add annotations for the same module, just like `add_needed_import`. That is what creates the ambiguity I pointed out above.

Decided to store only one stub in the context instead of maintaining a list, i.e., `ApplyTypeAnnotationsVisitor.store_stub_in_context(context, stub, overwrite_existing_annotations)`. This is more limited than the option of adding multiple stubs for the same module, but is less ambiguous and matches the notion of one stub per Python source file. Let me know if that works.

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -223,33 +223,49 @@ class ApplyTypeAnnotationsVisitor(ContextAwareTransformer):
         x: int = 1
 
     If the function or attribute already has a type annotation, it will not be overwritten.
+
+    To overwrite existing annotations when applying annotations from a stub,
+    use the keyword argument ``overwrite_existing_annotations=True`` when
+    constructing the codemod or when calling ``store_stub_in_context``.
```
jimmylai, line 229: @pradeep90 Can you add some explanation here and mention only the `overwrite_existing_annotations ` passed by last call take effect when multiple calls happen?

PR #298, opened 2020-05-21
Fix StopIteration exception during scope analysis

zsol:
## Summary

During scope analysis all attribute accesses are collected for matching on
import names. The matching code (specifically `_gen_dotted_names`) was not
prepared for all types of expressions. In particular, complex expressions like
`foo[0].bar.baz()` caused a `StopIteration` exception when `_gen_dotted_names`
calls itself recursively. The nested call doesn't yield any values, and so
calling `next()` on it raises.

This commit fixes these types of errors.

## Test Plan

Added test case

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -616,7 +616,10 @@ def _gen_dotted_names(
             # this is not an import
             return
         name_values = iter(_gen_dotted_names(value))
-        (next_name, next_node) = next(name_values)
+        next_pair = next(name_values, None)
+        if next_pair is None:
+            return
+        (next_name, next_node) = next_pair
         yield (f"{next_name}.{node.attr.value}", node)
         yield (next_name, next_node)
         yield from name_values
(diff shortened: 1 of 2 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -616,7 +616,10 @@ def _gen_dotted_names(
             # this is not an import
             return
         name_values = iter(_gen_dotted_names(value))
-        (next_name, next_node) = next(name_values)
+        next_pair = next(name_values, None)
+        if next_pair is None:
+            return
```
jimmylai, line 621: Can we just add `try ... except StopIteration`? That can be more efficient.

zsol: Hmm, sure. Why is it more efficient?

zsol:
I haven't yet.

libcst/metadata/scope_provider.py
jimmylai: That way we saved one extra `next` call.

zsol: Then I don't understand your suggestion :)
Apart from the way it's currently written, I can imagine this:
```
try:
    (next_name, next_node) = next(name_values)
    yield (f"...", ...)
    ...
except StopIteration:
    return
```

which is the same amount of `next` calls

libcst/metadata/scope_provider.py
```diff
@@ -616,7 +616,10 @@ def _gen_dotted_names(
...
+        (next_name, next_node) = next_pair
```
jimmylai, line 622: We just need try catch on this line. Then we can save the `next` in `next_pair = next(name_values, None)`.

jimmylai:
Merge this for now and will submit the proposed change in another PR.

PR #299, opened 2020-05-25
Add NamedTuple to dataclass conversion codemod

josieesh:
## Summary
Conversion of a NamedTuple class declaration to Python 3.7 `dataclasses.dataclass` type as an annotation. Perform removal of unused NamedTuple import and add the import of `dataclasses.dataclass` if not already imported. 
## Test Plan
Updated test suite:
```
python -m unittest libcst.codemod.commands.tests.test_convert_namedtuple_to_dataclass
```
Run entire test suite:
```
tox test
```

### DIFF
--- a/libcst/codemod/commands/convert_namedtuple_to_dataclass.py
+++ b/libcst/codemod/commands/convert_namedtuple_to_dataclass.py
@@ -1,0 +1,77 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+#
+# pyre-strict
+from typing import Sequence
+
+import libcst as cst
+from libcst import MaybeSentinel, ensure_type, parse_expression
+from libcst.codemod import VisitorBasedCodemodCommand
+from libcst.codemod.visitors import AddImportsVisitor, RemoveImportsVisitor
+from libcst.metadata import (
+    ProviderT,
+    QualifiedName,
+    QualifiedNameProvider,
+    QualifiedNameSource,
+)
+
+
+class ConvertNamedTupleToDataclassCommand(VisitorBasedCodemodCommand):
+    """
+    Convert NamedTuple class declarations to Python 3.7 dataclasses.
+
+    This only performs a conversion at the class declaration level.
+    It does not perform type annotation conversions, nor does it convert
+    NamedTuple-specific attributes and methods.
+    """
+
+    DESCRIPTION: str = "Convert legacy NamedTuple class declarations to Python 3.7 dataclasses."
+    METADATA_DEPENDENCIES: Sequence[ProviderT] = (QualifiedNameProvider,)
+
+    MODULE: str = "typing"
+    OBJECT: str = "NamedTuple"
+
+    def leave_ClassDef(
+        self, original_node: cst.ClassDef, updated_node: cst.ClassDef
+    ) -> cst.ClassDef:
+        new_bases = []
+        namedtuple_base = None
+
+        for base_class in original_node.bases:
+            # Compare the base class's qualified named against the expected typing.NamedTuple
+            if not QualifiedNameProvider.has_name(
+                self,
+                base_class.value,
+                QualifiedName(
+                    name=f"{self.MODULE}.{self.OBJECT}",
+                    source=QualifiedNameSource.IMPORT,
+                ),
+            ):
+                # Keep all bases that are not of type typing.NamedTuple
+                new_bases.append(base_class)
+            else:
+                namedtuple_base = base_class
+
+        # We still want to return the updated node in case some of its children have been modified
+        if not namedtuple_base:
+            return updated_node
+
+        AddImportsVisitor.add_needed_import(self.context, "dataclasses", "dataclass")
+        RemoveImportsVisitor.remove_unused_import_by_node(
+            self.context, namedtuple_base.value
+        )
+
+        call = ensure_type(
+            parse_expression(
+                "dataclass(frozen=True)", config=self.module.config_for_parsing
+            ),
+            cst.Call,
+        )
+        return updated_node.with_changes(
+            lpar=MaybeSentinel.DEFAULT,
+            rpar=MaybeSentinel.DEFAULT,
+            bases=new_bases,
+            decorators=[*original_node.decorators, cst.Decorator(decorator=call)],
+        )
--- a/libcst/codemod/commands/convert_namedtuple_to_dataclass.py
+++ b/libcst/codemod/commands/convert_namedtuple_to_dataclass.py
@@ -53,7 +53,7 @@ class ConvertNamedTupleToDataclassCommand(VisitorBasedCodemodCommand):
                 namedtuple_base = base_class
 
         # We still want to return the updated node in case some of its children have been modified
-        if not namedtuple_base:
+        if namedtuple_base is None:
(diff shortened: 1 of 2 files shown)


libcst/codemod/commands/convert_namedtuple_to_dataclass.py
```diff
@@ -0,0 +1,77 @@
...
+    MODULE: str = "typing"
+    OBJECT: str = "NamedTuple"
+
+    def leave_ClassDef(
+        self, original_node: cst.ClassDef, updated_node: cst.ClassDef
+    ) -> cst.ClassDef:
+        new_bases = []
+        namedtuple_base = None
+
+        for base_class in original_node.bases:
```
thatch, line 42: Are you sure you want `original_node` here?  If so, worth a comment.

libcst/codemod/commands/convert_namedtuple_to_dataclass.py
```diff
@@ -0,0 +1,77 @@
...
+    DESCRIPTION: str = "Convert legacy NamedTuple class declarations to Python 3.7 dataclasses."
+    METADATA_DEPENDENCIES: Sequence[ProviderT] = (QualifiedNameProvider,)
+
+    MODULE: str = "typing"
+    OBJECT: str = "NamedTuple"
+
+    def leave_ClassDef(
+        self, original_node: cst.ClassDef, updated_node: cst.ClassDef
+    ) -> cst.ClassDef:
+        new_bases = []
```
thatch, line 39: I had expected this to need an annotation of `List[cst.Arg]` or so... can you double-check that type checking is covering this?

libcst/codemod/commands/convert_namedtuple_to_dataclass.py
```diff
@@ -0,0 +1,77 @@
...
+        namedtuple_base = None
+
+        for base_class in original_node.bases:
+            # Compare the base class's qualified named against the expected typing.NamedTuple
+            if not QualifiedNameProvider.has_name(
+                self,
+                base_class.value,
+                QualifiedName(
```
thatch, line 47: If you hoist this out constant of the loop, the has_name call might fit on one line.

josieesh: Good point

libcst/codemod/commands/convert_namedtuple_to_dataclass.py
josieesh: Seems that the metadata dict is directly tied to the original node's bases, so `updated_node.bases` does not work with the QualifiedNameProvider here. Added a comment to acknowledge this 👍

libcst/codemod/commands/convert_namedtuple_to_dataclass.py
```diff
@@ -0,0 +1,75 @@
...
+class ConvertNamedTupleToDataclassCommand(VisitorBasedCodemodCommand):
+    """
+    Convert NamedTuple class declarations to Python 3.7 dataclasses.
+
+    This only performs a conversion at the class declaration level.
+    It does not perform type annotation conversions, nor does it convert
+    NamedTuple-specific attributes and methods.
+    """
+
+    DESCRIPTION: str = "Convert legacy NamedTuple class declarations to Python 3.7 dataclasses."
```
carljm, line 29: minor nit, but probably best to leave out the word "legacy" here and avoid provoking any pedantry or misleading anyone, since in upstream Python NamedTuple is not (as far as I know) deprecated or discouraged in any way. In our codebases we prefer to convert NamedTuple to dataclass, but that doesn't make NamedTuple "legacy" in general.

libcst/codemod/commands/convert_namedtuple_to_dataclass.py
```diff
@@ -0,0 +1,75 @@
...
+    ) -> cst.ClassDef:
+        new_bases: List[cst.Arg] = []
+        namedtuple_base: Optional[cst.Arg] = None
+        qualified_namedtuple: QualifiedName = QualifiedName(
+            name=f"{self.MODULE}.{self.OBJECT}", source=QualifiedNameSource.IMPORT
+        )
+
+        # Need to examine the original node's bases since they are directly tied to import metadata
+        for base_class in original_node.bases:
+            # Compare the base class's qualified named against the expected typing.NamedTuple
```
carljm, line 46:
```suggestion
            # Compare the base class' qualified name against the expected typing.NamedTuple
```

libcst/codemod/commands/convert_namedtuple_to_dataclass.py
```diff
@@ -0,0 +1,75 @@
...
+    OBJECT: str = "NamedTuple"
+
+    def leave_ClassDef(
+        self, original_node: cst.ClassDef, updated_node: cst.ClassDef
+    ) -> cst.ClassDef:
+        new_bases: List[cst.Arg] = []
+        namedtuple_base: Optional[cst.Arg] = None
+        qualified_namedtuple: QualifiedName = QualifiedName(
+            name=f"{self.MODULE}.{self.OBJECT}", source=QualifiedNameSource.IMPORT
+        )
```
carljm, line 42: This is immutable and creating it has no side effects; any reason not to just make the entire `QualifiedName` object a class attribute, and not have to construct it anew each time we leave a class definition?

libcst/codemod/commands/convert_namedtuple_to_dataclass.py
```diff
@@ -0,0 +1,75 @@
...
+
+        call = cst.ensure_type(
+            cst.parse_expression(
+                "dataclass(frozen=True)", config=self.module.config_for_parsing
+            ),
+            cst.Call,
+        )
+        return updated_node.with_changes(
+            lpar=cst.MaybeSentinel.DEFAULT,
+            rpar=cst.MaybeSentinel.DEFAULT,
```
carljm, line 72: Are these two lines necessary? What fails without them?

josieesh: They account for the scenario where all of the base classes have been removed, so that we default to `class Foo:` instead of `class Foo():`

libcst/codemod/commands/convert_namedtuple_to_dataclass.py
josieesh: Very good point

libcst/codemod/commands/convert_namedtuple_to_dataclass.py
thatch: @carljm  `class's` is legal for singular, which this is one at a time.  did you mean `classes'` to talk about the whole operation?

jimmylai:
> Do we want documentation for this, or will we get autogenerated reference docs from the docstring?

We only have the document for transform of libraries and we can automatically generate the list of available CodemodCommands to keep the document update easier. We'll cover documentation later in Milestone 3 and @josieesh can work on it at that time (It requires writing some code to generate the rst for Sphinx and we already have something similar for our lint rules.)
https://libcst.readthedocs.io/en/latest/codemods.html#library-of-transforms

libcst/codemod/commands/convert_namedtuple_to_dataclass.py
carljm: @thatch Oops, yeah I was thinking of the plural. Made the comment due to the extraneous `d` and the other part was an afterthought that clearly needed a bit more thought :)

Issue #297, opened 2020-05-14
ImportError under Python 3.9


Zac-HD:
Via https://github.com/Zac-HD/stdlib-property-tests/issues/12, attempting to `import libcst` fails on Python 3.9 with

```python-traceback
Traceback (most recent call last):
  File "t.py", line 3, in <module>
    import libcst
  File "libcst/__init__.py", line 187, in <module>
    from libcst._parser.entrypoints import parse_expression, parse_module, parse_statement
  File "libcst/_parser/entrypoints.py", line 25, in <module>
    _DEFAULT_PARTIAL_PARSER_CONFIG: PartialParserConfig = PartialParserConfig()
  File "<string>", line 8, in __init__
  File "libcst/_parser/types/config.py", line 132, in __post_init__
    raise ValueError(
ValueError: LibCST can only parse code using one of the following versions of
		    Python's grammar: 3.0, 3.1, 3.3, 3.5, 3.6, 3.7, 3.8. More versions may 
		    be supported by future releases.
```

Given that I'm using `libcst` in `hypothesmith` to test (among other things) pre-release Python versions, it would be lovely if libcst defaulted to the newest known grammar on newer-than-known Python versions rather than raising an error on import :smile:

SimiCode:
I'm getting this error too when using MonkeyType
cc: @carljm

SimiCode:
What can we do to make the change happen? 
Does it need testing? expansion?
@jimmylai

Zac-HD:
Closed by #306.

(@thatch - the magic auto-close syntax only works in the PR description, not latter comments)

Issue #305, opened 2020-05-31
type_enforce test failure on 3.9


thatch:
Once #297 is fixed, there are a couple of failures.  This one is a behavior change in `typing.Iterable`.

```
======================================================================
ERROR: test_basic_pass_19 (libcst.tests.test_type_enforce.TypeEnforcementTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tim/code/LibCST/libcst/testing/utils.py", line 89, in new_test
    return member(self, *data)
  File "/home/tim/code/LibCST/libcst/tests/test_type_enforce.py", line 135, in test_basic_pass
    is_value_of_type(value, expected_type),
  File "/home/tim/code/LibCST/libcst/_type_enforce.py", line 136, in is_value_of_type
    elif issubclass(expected_origin_type, Iterable) and not issubclass(
TypeError: issubclass() arg 1 must be a class
```

It appears to be a difference in `typing_inspect.get_origin`.

Issue #304, opened 2020-05-31
Codegen test failure on 3.9


thatch:
Once #297 is fixed, there are a couple of failures.  This one is that the codegen test fails, and indeed it would create a rather significant diff if you run codegen under 3.9.

https://gist.github.com/thatch/7f402910a46fdbe6c4e17a9493ef91a5

```
======================================================================
FAIL: test_codegen_clean_matcher_classes (libcst.codegen.tests.test_codegen_clean.TestCodegenClean)
Verifies that codegen of matcher classes would not result in a
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tim/code/LibCST/libcst/codegen/tests/test_codegen_clean.py", line 82, in test_codegen_clean_matcher_classes
    self.assertTrue(
AssertionError: False is not true : libcst.matchers.__init__ needs new codegen!
```

PR #306, opened 2020-05-31
Improve default-version selection logic

thatch:
## Summary

We need this to be able to import at all on 3.9, because there aren't any syntax changes that would warrant it being a separate entry in the list.

## Test Plan

Compiled 3.9b1 and this diff enables it in tox.ini.
```
tox -p all
```

Failures are filed as separate issues, #304 and #305


thatch:
Closes #297

thatch:
@jimmylai can you take a look?

PR #308, opened 2020-06-09
Add codemod RenameCommand

josieesh:
## Summary
Rename Codemod allows to rename imported and local objects. If imported object is renamed, the import is force-removed and a new import is added.

## Test Plan
- Repo test suite should pass
```
tox
```
- Unittests updated
```
python -m unittest libcst.codemod.commands.tests.test_rename.TestRenameCommand
```
- Assert codemod works as expected on a test file `z.py`
```
python -m libcst.tool codemod rename.RenameCommand z.py --orig_object=x.bar --new_object=x.baz
```

### DIFF
--- a/libcst/codemod/commands/rename.py
+++ b/libcst/codemod/commands/rename.py
@@ -5,7 +5,7 @@
 #
 # pyre-strict
 import argparse
-from typing import Dict, Sequence, Set, Union
+from typing import Sequence, Set, Union
 
 import libcst as cst
 from libcst.codemod import CodemodContext, VisitorBasedCodemodCommand
@@ -53,13 +53,6 @@ class RenameCommand(VisitorBasedCodemodCommand):
         self.old_mod_or_obj: str = old_mod_or_obj
         self.new_mod_or_obj: str = new_mod_or_obj
 
-        # Mapping for easy lookup later.
-        self.equivalent_mod_or_objs: Dict[str, str] = {
-            self.old_mod_or_obj: self.new_mod_or_obj
-        }
-
-        self.equivalent_modules: Dict[str, str] = {self.old_module: self.new_module}
-
         self.scheduled_removals: Set[cst.CSTNode] = set()
         self.bypass_imports = False
         self.schedule_import = False
@@ -71,11 +64,10 @@ class RenameCommand(VisitorBasedCodemodCommand):
 
                 # If the import statement is exactly equivalent to the old name, it will be taken care of in `leave_Name` or `leave_Attribute`.
                 if alias_name == self.old_name:
-                    self.record_asname(import_alias, alias_name)
                     self.bypass_imports = True
                 # If the import is a parent module of `self.old_name`, it will be renamed in `leave_Import`.
                 elif self.old_name.startswith(alias_name + "."):
-                    self.record_asname(import_alias, alias_name)
+                    pass
                 # If we are renaming a top-level module of the import, we know that the rename will be taken care of in `leave_Name` or `leave_Attribute`.
                 elif alias_name.startswith(self.old_name + "."):
                     self.bypass_imports = True
@@ -89,6 +81,7 @@ class RenameCommand(VisitorBasedCodemodCommand):
             import_alias_full_name = get_full_name_for_node(import_alias_name)
             if import_alias_full_name is None:
                 raise Exception("Could not parse full name for ImportAlias.name node.")
+
             if isinstance(import_alias_name, cst.Name) and self.old_name.startswith(
                 import_alias_full_name + "."
             ):
@@ -120,6 +113,15 @@ class RenameCommand(VisitorBasedCodemodCommand):
                 self.bypass_imports = True
             else:
                 new_names.append(import_alias)
+
+        # Finally, we need to record the 'as' name using the original node.
+        for import_alias in original_node.names:
+            alias_name = get_full_name_for_node(import_alias.name)
+            if alias_name is not None:
+                if alias_name == self.old_name or self.old_name.startswith(
+                    alias_name + "."
+                ):
+                    self.record_asname(import_alias, alias_name)
         return updated_node.with_changes(names=new_names)
 
     def visit_ImportFrom(self, node: cst.ImportFrom) -> None:
@@ -159,12 +161,14 @@ class RenameCommand(VisitorBasedCodemodCommand):
                 if alias_name is not None:
                     qual_name = f"{imported_module_name}.{alias_name}"
                     if self.old_name == qual_name:
-                        self.record_asname(import_alias, alias_name)
 
                         replacement_module = self.gen_replacement_module(
                             imported_module_name
                         )
                         replacement_obj = self.gen_replacement(alias_name)
+
+                        # Record the 'as' name after we finish generating the new names for the import.
+                        self.record_asname(import_alias, alias_name)
                         new_import_alias_name: cst.BaseExpression = cst.parse_expression(
                             replacement_obj
                         )
@@ -193,10 +197,10 @@ class RenameCommand(VisitorBasedCodemodCommand):
                         # and if the module is the same, add an extra alias, if not, schedule a whole new import.
                         new_names.append(import_alias)
                         self.scheduled_removals.add(original_node)
-                        if imported_module_name in self.equivalent_modules:
-                            new_module_name = self.gen_replacement_module(
-                                imported_module_name
-                            )
+                        new_module_name = self.gen_replacement_module(
+                            imported_module_name
+                        )
+                        if new_module_name:
                             if new_module_name == imported_module_name:
                                 new_names.append(
                                     cst.ImportAlias(
@@ -271,35 +275,24 @@ class RenameCommand(VisitorBasedCodemodCommand):
         return updated_node
 
     def gen_replacement(self, original_name: str) -> str:
-        index = original_name.rfind(".")
-        parts = (
-            (original_name[: index + 1], original_name[index + 1 :])
-            if index > -1
-            else (original_name,)
-        )
-        replacement_parts = []
-        for part in parts:
-            if part in self.equivalent_mod_or_objs:
-                replacement_parts.append(self.equivalent_mod_or_objs[part])
-            elif part.rstrip(".") in self.equivalent_modules:
-                replacement_parts.append(self.equivalent_modules[part.rstrip(".")])
-                replacement_parts.append(".")
-
-        return "".join(replacement_parts).rstrip(".")
+        if original_name == self.old_mod_or_obj:
+            return self.new_mod_or_obj
+        elif original_name == ".".join([self.old_module, self.old_mod_or_obj]):
+            return self.new_name
+        elif original_name.endswith("." + self.old_mod_or_obj):
+            return self.new_mod_or_obj
+        else:
+            return self.gen_replacement_module(original_name)
 
     def gen_replacement_module(self, original_module: str) -> str:
-        return self.equivalent_modules.get(original_module, "")
+        return self.new_module if original_module == self.old_module else ""
 
     def record_asname(self, import_alias: cst.ImportAlias, import_name: str) -> None:
-        # Record the import's `as` name if it has one, and map it to the equivalent new name.
+        # Record the import's `as` name if it has one, and set the corresponding attribute to it.
         as_name = import_alias.asname
         name = as_name.name if as_name is not None else None
         if name is not None and isinstance(name, cst.Name):
-            if import_name in self.equivalent_mod_or_objs:
-                self.equivalent_mod_or_objs[name.value] = self.equivalent_mod_or_objs[
-                    import_name
-                ]
-            elif import_name in self.equivalent_modules:
-                self.equivalent_modules[name.value] = self.equivalent_modules[
-                    import_name
-                ]
+            if import_name == self.old_mod_or_obj:
+                self.old_mod_or_obj = name.value
+            elif import_name == self.old_module:
+                self.old_module = name.value
--- a/libcst/codemod/commands/rename.py
+++ b/libcst/codemod/commands/rename.py
@@ -5,13 +5,31 @@
 #
 # pyre-strict
 import argparse
-from typing import Sequence, Set, Union
+from typing import Callable, Optional, Sequence, Set, Tuple, Union
 
 import libcst as cst
 from libcst.codemod import CodemodContext, VisitorBasedCodemodCommand
 from libcst.codemod.visitors import AddImportsVisitor, RemoveImportsVisitor
 from libcst.helpers import get_full_name_for_node
 from libcst.metadata import QualifiedNameProvider
+
+
+def leave_import_decorator(
+    method: Callable[..., Union[cst.Import, cst.ImportFrom]]
+) -> Callable[..., Union[cst.Import, cst.ImportFrom]]:
+    # We want to record any 'as name' that is relevant but only after we leave the corresponding Import/ImportFrom node as we don't want the 'as name'
+    # to interfere with children 'Name' and 'Attribute' nodes.
+    def wrapper(
+        self: VisitorBasedCodemodCommand,
+        original_node: Union[cst.Import, cst.ImportFrom],
+        updated_node: Union[cst.Import, cst.ImportFrom],
+    ) -> Union[cst.Import, cst.ImportFrom]:
+        updated_node = method(self, original_node, updated_node)
+        if original_node != updated_node:
+            getattr(self, "record_asname")(original_node)
+        return updated_node
+
+    return wrapper
 
 
 class RenameCommand(VisitorBasedCodemodCommand):
@@ -53,6 +71,8 @@ class RenameCommand(VisitorBasedCodemodCommand):
         self.old_mod_or_obj: str = old_mod_or_obj
         self.new_mod_or_obj: str = new_mod_or_obj
 
+        self.as_name: Optional[Tuple[str, str]] = None
+
         self.scheduled_removals: Set[cst.CSTNode] = set()
         self.bypass_imports = False
         self.schedule_import = False
@@ -65,13 +85,11 @@ class RenameCommand(VisitorBasedCodemodCommand):
                 # If the import statement is exactly equivalent to the old name, it will be taken care of in `leave_Name` or `leave_Attribute`.
                 if alias_name == self.old_name:
                     self.bypass_imports = True
-                # If the import is a parent module of `self.old_name`, it will be renamed in `leave_Import`.
-                elif self.old_name.startswith(alias_name + "."):
-                    pass
                 # If we are renaming a top-level module of the import, we know that the rename will be taken care of in `leave_Name` or `leave_Attribute`.
                 elif alias_name.startswith(self.old_name + "."):
                     self.bypass_imports = True
 
+    @leave_import_decorator
     def leave_Import(
         self, original_node: cst.Import, updated_node: cst.Import
     ) -> cst.Import:
@@ -114,14 +132,6 @@ class RenameCommand(VisitorBasedCodemodCommand):
             else:
                 new_names.append(import_alias)
 
-        # Finally, we need to record the 'as' name using the original node.
-        for import_alias in original_node.names:
-            alias_name = get_full_name_for_node(import_alias.name)
-            if alias_name is not None:
-                if alias_name == self.old_name or self.old_name.startswith(
-                    alias_name + "."
-                ):
-                    self.record_asname(import_alias, alias_name)
         return updated_node.with_changes(names=new_names)
 
     def visit_ImportFrom(self, node: cst.ImportFrom) -> None:
@@ -131,17 +141,14 @@ class RenameCommand(VisitorBasedCodemodCommand):
         imported_module_name = get_full_name_for_node(module)
         if imported_module_name is None:
             return
-
         if imported_module_name == self.old_name:
             # If the imported module is exactly equivalent to the old name, it will be taken care of in `leave_Name` or `leave_Attribute`.
             self.bypass_imports = True
-        elif self.old_name.startswith(imported_module_name + "."):
-            # If the imported module is a parent module of `self.old_name`, it will be renamed in `leave_ImportFrom`.
-            pass
         elif imported_module_name.startswith(self.old_name + "."):
             # If we are renaming a parent module of the current module, we know that the rename will be taken care of in `leave_Name` or `leave_Attribute`.
             self.bypass_imports = True
 
+    @leave_import_decorator
     def leave_ImportFrom(
         self, original_node: cst.ImportFrom, updated_node: cst.ImportFrom
     ) -> cst.ImportFrom:
@@ -167,8 +174,6 @@ class RenameCommand(VisitorBasedCodemodCommand):
                         )
                         replacement_obj = self.gen_replacement(alias_name)
 
-                        # Record the 'as' name after we finish generating the new names for the import.
-                        self.record_asname(import_alias, alias_name)
                         new_import_alias_name: cst.BaseExpression = cst.parse_expression(
                             replacement_obj
                         )
@@ -275,6 +280,15 @@ class RenameCommand(VisitorBasedCodemodCommand):
         return updated_node
 
     def gen_replacement(self, original_name: str) -> str:
+        module_as_name = self.as_name
+        if module_as_name is not None:
+            if original_name == module_as_name[0]:
+                original_name = module_as_name[1]
+            elif original_name.startswith(module_as_name[0] + "."):
+                original_name = original_name.replace(
+                    module_as_name[0] + ".", module_as_name[1] + ".", 1
+                )
+
         if original_name == self.old_mod_or_obj:
             return self.new_mod_or_obj
         elif original_name == ".".join([self.old_module, self.old_mod_or_obj]):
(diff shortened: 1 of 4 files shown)


jimmylai:
Can you generate a git diff to show the changes done for fixing the import removal issue? That can make the code review easier. You can either pasted the diff in PR summary/comment or a GH gist.

josieesh:
> Can you generate a git diff to show the changes done for fixing the import removal issue? That can make the code review easier. You can either pasted the diff in PR summary/comment or a GH gist.

Sure thing! A diff between this file and the one in distillery? :)

jimmylai:
> > Can you generate a git diff to show the changes done for fixing the import removal issue? That can make the code review easier. You can either pasted the diff in PR summary/comment or a GH gist.
> 
> Sure thing! A diff between this file and the one in distillery? :)

Yes.

josieesh:
> > > Can you generate a git diff to show the changes done for fixing the import removal issue? That can make the code review easier. You can either pasted the diff in PR summary/comment or a GH gist.
> > 
> > 
> > Sure thing! A diff between this file and the one in distillery? :)
> 
> Yes.

Gist available here: https://gist.github.com/josieesh/4c0262add03d1911b4896af575f221e7

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,192 @@
...
+            module_asname = (
+                None
+                if module_name == self.orig_module
+                else ".".join(name.split(".")[:-1])
+            )
+
+            if self.new_module is not None:
+                AddImportsVisitor.add_needed_import(
+                    self.context,
+                    # pyre-ignore[6]: Expected `str` for 2nd parameter `module` to call
```
thatch, line 147: Instead of pyre-ignore, these can be silenced with something like `module = self.new_module; assert module is not None`

jimmylai: Correct. We don't usually use pyre-ignore unless it's something Pyre cannot understand. When adding a pyre-ignore, we usually also report the issue to Pyre team as feature request.
Other than use `assert` to teach Pyre a variable is not None, we usually use an if check since assert is a noop in production mode unlike an `if`.
E.g.
```
module = self.new_module
if module is not None:
    # Pyre won't complain module is an `Optional` in this scope.
```

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,192 @@
...
+        )
+
+    def leave_Import(
+        self, original_node: cst.Import, updated_node: cst.Import
+    ) -> Union[cst.Import, cst.RemovalSentinel]:
+        for import_alias in original_node.names:
+            name = get_full_name_for_node(import_alias.name)
+            if name is not None and name == self.orig_module:
+                # Remove old import altogether
+                return cst.RemoveFromParent()
```
jimmylai, line 91: In the case of `import a, b, c`, there will be multiple `names`.
If only `b` is the target to be removed, this will remove the entire import statement which is wrong.
We should just create a new list of `names` by filtering the existing names and then return the new `names` in an updated `Import` node.
Let's also add test cases to cover this case.

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,192 @@
...
+            or not isinstance(names, Sequence)
+        ):
+            return updated_node
+
+        if (
+            len(names) == 1
+            and get_full_name_for_node(names[0].name) == self.orig_object
+        ):
+            # Remove this ImportFrom altogether
+            return cst.RemoveFromParent()
```
jimmylai, line 116: This is a pretty clever way to implement it!

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,202 @@
...
+        self.assertCodemod(
+            before,
+            after,
+            orig_module="foo",
+            orig_object="qux",
+            new_module="baz",
+            new_object="quux",
+        )
+
+    def test_rename_module(self) -> None:
```
apadmarao, line 114: This tests a class rename not a module rename. Using an uppercase class name `B` instead of `b` would make this more clear.

Separately, we should test a module rename as well!

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,202 @@
...
+        self.assertCodemod(
+            before,
+            after,
+            orig_module="a",
+            orig_object="b",
+            new_module="c",
+            new_object="b",
+        )
+
+    def test_rename_local_variable(self) -> None:
```
apadmarao, line 137: This is very cool! How do we anticipate that people use this feature, if there are many local variables with the same name in different functions in a module? (This comment also applies to all of the other "rename local" features.)

This leads me to think we may want a separate codemod command to rename local variables.

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,202 @@
...
+                pass
+        """
+        after = """
+            from a import c
+
+            class Foo(c):
+                pass
+        """
+        self.assertCodemod(
+            before, after, orig_module="a", orig_object="b", new_object="c"
```
apadmarao, line 165: Reading this, I was initially a bit confused that we implicitly assume `new_module` is "a". It could be more obvious if we required the `new_module` parameter any time there is an `old_module` parameter.

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,199 @@
...
+import libcst as cst
+from libcst.codemod import CodemodContext, VisitorBasedCodemodCommand
+from libcst.codemod.visitors import AddImportsVisitor
+from libcst.helpers import get_full_name_for_node
+from libcst.metadata import QualifiedName, QualifiedNameProvider, QualifiedNameSource
+
+
+class RenameCommand(VisitorBasedCodemodCommand):
+    """
+    Rename all instances of a variable or import.
```
apadmarao, line 19: The rename command works for all Python objects (module, class, function, name). I _think_ this is what you intend with "import" but different phrasing could be more clear.

libcst/codemod/commands/rename.py
apadmarao: We do not want to use an import if it is referenced in code (The internal version of rename handles this I believe). A `from` import can also be used to import a module which is used in multiple places (so solution 1 is not a viable solution).

libcst/codemod/commands/tests/test_rename.py
jimmylai: We'll always need a `orig_module` to locate the import module of `orig_object`. When `new_module` is not provided, it means the developer don't want to rename the module name. So at least one of `new_module` or `new_object` should be provided in a rename use case.

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+
+    DESCRIPTION: str = (
+        "Rename all instances of a variable or import. "
+        + "The existing import may not be removed and RenameCommand needs more work to be perfect."
```
jimmylai, line 24: The `not be removed` issue should be fixed by this PR. Let's update the description.

libcst/codemod/commands/rename.py
josieesh: Yeah this was on my mind as well. But then I realized that the way arguments are passed to the codemod, the imported module won't get removed. And this is checked in the test [test_no_removal_of_import_from_in_use](https://github.com/Instagram/LibCST/pull/308/files#diff-511afb8b206c354e493c782e1c50abadR245)

libcst/codemod/commands/tests/test_rename.py
josieesh: I have allowed the omission of both for the case of renaming a local variable. Do we want to disallow this altogether?

apadmarao: Renaming a local variable is a very cool feature! 

The reservation I have with the current implementation is that you oftentimes a local variable name is repeated within multiple classes, methods, or functions within a module. I don't think it makes sense to rename all of the local variables in the same manner.

So, for renaming a local variable, I imagine that you want to target a scope that may be more granular than a file / module. 

This is why I was thinking that it could make sense to have a different interface for renaming local variables, and hence a separate codemod command.

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,170 @@
...
+        for import_alias in original_node.names:
+            name = get_full_name_for_node(import_alias.name)
+            if name is not None and name == self.orig_module:
+                # Schedule this import to be potentially removed
+                RemoveImportsVisitor.remove_unused_import_by_node(
+                    self.context, original_node
+                )
+        return original_node
+
+    def leave_ImportFrom(
```
apadmarao, line 93: I'm worried that this code will not work when the name you import is referenced in other places in code (for example if a module is imported); this code does not use `RemoveImportsVisitor`as `leave_Import` which checks that the import is unused before removing it.

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,290 @@
...
+            from a import m
+            import blah
+
+            class Foo(blah.CstNode):
+                bar: m.Attribute
+        """
+        self.assertCodemod(
+            before,
+            after,
+            orig_module="a",
```
apadmarao, line 262: Does this work if `orig_module="a.m"` and `orig_object="CstNode"`?

josieesh: Just added another test: https://github.com/Instagram/LibCST/pull/308/files#diff-511afb8b206c354e493c782e1c50abadR268

libcst/codemod/commands/rename.py
josieesh: The problem with using RemoveImportVisitor is that in the scenario that the new `module` + `object` combination may be the same as the old one eg: `from fixit.common import CstLintRule` to replace `from package.module.common import CstLintRule`. In this scenario, the qualified names of the base class of `SomeRule(CstLintRule)` will be an array of both of those modules and RemoveImportsVisitor will not remove the old import since it's still being "used". There's really no way to avoid this with RemoveImportsVisitor.

libcst/codemod/commands/tests/test_rename.py
carljm: There are multiple levels of "local"; here it just means module-local (not imported). This could mean a module-level "global" variable, which would be reasonable to want to rename module-wide.

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,170 @@
...
+            "--new_module",
+            dest="new_module",
+            help="New name of the module. Leave blank if renaming a local variable, or if same as orig_module.",
+        )
+        parser.add_argument(
+            "--new_object",
+            dest="new_object",
+            required=True,
+            help="New name of the object in module or variable.",
+        )
```
carljm, line 49: I wonder about simplifying this so that we just take `oldname` and `newname`, where either one or both can be a dotted path. If either one contains one or more dots, we internally split it on the final dot to split module name from object name. Otherwise the logic remains the same.

So instead of typing `--orig_module foo.bar --orig_object baz --new_module one.two --new_object three` I would just type `--oldname foo.bar.baz --newname one.two.three`, which is a lot less verbose and I think clearer to read also.

One question then might be whether the meaning of `--oldname foo.bar.baz --newname bing` should be shorthand for `--oldname foo.bar.baz --newname foo.bar.bing` (that's what the current logic would suggest) or should instead be "rename baz to bing and remove import of baz from foo.bar without replacing it" implying that it's becoming module-local. I would favor the latter, I think it's kinda non-obvious for omitting new-module to mean "don't change old module".

libcst/codemod/commands/rename.py
carljm: I think this code should be safe against removing any still-used import; it should only ever remove the single from-import of the name we are renaming.

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,170 @@
...
+    def leave_Attribute(
+        self, original_node: cst.Attribute, updated_node: cst.Attribute
+    ) -> cst.Attribute:
+        if QualifiedNameProvider.has_name(
+            self, original_node, QualifiedName(name=self.orig_name, source=self.source),
+        ):
+            full_name = get_full_name_for_node(original_node)
+            if full_name is None:
+                raise Exception("Logic error!")
+            module_name = ".".join(full_name.split(".")[:-1])
```
carljm, line 135: Maybe `module_name = full_name.rpartition(".")[0]`?

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,170 @@
...
+            # New assignment for pyre check.
```
carljm, line 136: probably don't need this comment?

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,170 @@
...
+                AddImportsVisitor.add_needed_import(
+                    self.context, module=new_import, asname=None,
+                )
+            # New assignment for pyre check.
+            new_value = self.new_module if self.new_module is not None else module_name
+            # If we reach here, new_value should never technically be None since new_module and module_asname cannot both be None.
+            if new_value is not None:
+                return updated_node.with_changes(
+                    value=cst.parse_expression(new_value),
+                    attr=cst.Name(value=self.new_object.split(".")[-1]),
```
carljm, line 148: in what case would `self.new_object` be expected to have dots in it?

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,313 @@
...
+            orig_object="m.CstNode",
```
carljm, line 263: it surprises me a bit that we support this form (with dots in the object name) at all. But IMO the fact that we need to test both ways just strengthens the idea of having a two-argument API where both arguments are full dotted names (or undotted local names), as discussed above.

carljm:
This is a super-cool codemod, can't wait to use it!

libcst/codemod/commands/rename.py
josieesh: This could be a good solution to disambiguate the codemod ie provide the full qualified name of the object one is renaming. One scenario I can see arising though is what about the case where we just want to rename a module eg `foo.bar.baz` where `baz` is still a module and not an object. But perhaps we could just do two checks in the codemod for each node we encounter and see if one of them is a match...

carljm: It's clearer if we talk about which syntactic forms we want to support rather than what `baz` _is_, because the codemod doesn't actually know or care whether `baz` in `foo.bar.baz` is a module or some other kind of object; even in the current form if you say `--orig_module foo.bar --orig_object baz`, `baz` could in fact be a module but the codemod will still happily change `from foo.bar import baz` to the requested new name. That's part of the reason I think it's more intuitive and will lead to fewer surprises if we just ask for a single dotted name and replace any syntactic form we see that matches it, rather than making implicit assertions about what is or isn't a module.

So I'm guessing the syntactic form you mean here is if we have `import foo.bar.baz` or `from foo.bar.baz import qux, quux` and we want to change `foo.bar.baz` to something else, without changing any names imported from it or accessed as attributes on it? Yes, I think in my API proposal if we specified `--oldname foo.bar.baz` that case should be supported.

libcst/codemod/commands/rename.py
apadmarao: I looked at the implementation again. It only removes a `from ...` import if the fully-qualified name being imported exactly matches the fully-qualified name being renamed. 

This is why the test [below](https://github.com/Instagram/LibCST/pull/308/files#r442561556) pass - because we're renaming an object in a module but we're importing a module.

I don't think this code will remove an unreferenced module import, where we're renaming an object in the module.

For example,
```
from a import b

b.c = 1
```
where we're renaming "a.b.c" to "d.e.f"

libcst/codemod/commands/tests/test_rename.py
apadmarao: I looked a bit more into how `QualifiedNameProvider` works. Presumably the reason why this is the case is because the qualified name of a function local `y` in `x`

```
def x():
  y = 1
  ...
```
is `x.<locals>.y` and not simply `y`. Does this mean that we can rename function or class locals by correctly providing the right qualified name?

apadmarao: Also, we generally design codemod commands to run over an entire codebase and not a single module or file. Renaming local variables (whether module-local or not) only really makes sense to run on a single module - it's unlikely you want to rename a local variable across the codebase.

The question I have is: should we provide a separate interface for renames of local variables that takes the target module as well?

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,192 @@
...
+        self, original_node: cst.ImportFrom, updated_node: cst.ImportFrom
+    ) -> cst.ImportFrom:
+        module = updated_node.module
+        if module is not None:
+            imported_module_name = get_full_name_for_node(module)
+            names = original_node.names
+
+            if imported_module_name is None or not isinstance(names, Sequence):
+                return updated_node
+            elif imported_module_name == self.old_name:
```
carljm, line 78: Similar to above, seems like we're missing the case here where the changed name is a prefix of `imported_module_name`, e.g. we're renaming `a.b` to `x.y` and we have `from a.b.c import d`, which should become `from x.y.c import d`

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,192 @@
...
+                                imported_module_name + "."
+                            )
+                            replacement_obj = self.gen_replacement(alias_name)
+                            new_import_alias_name: cst.BaseExpression = cst.parse_expression(
+                                replacement_obj
+                            )
+                            if not isinstance(
+                                new_import_alias_name, (cst.Attribute, cst.Name)
+                            ):
+                                raise Exception("Something went wrong!")
```
carljm, line 102: This is a pretty classic not-useful error message :) I understand it should never happen and you're just keeping the type-checker happy, but it still seems like nothing would be lost by changing it to `"parse_expression() on dotted path returned non-Attribute-or-Name"`.

Another alternative to consider might be a simple `assert`: `assert isinstance(new_import_alias_name, (cst.Attribute, cst.Name))`

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,385 @@
...
+    def test_rename_attr(self) -> None:
+
+        before = """
+            import a.b
+
+            def test() -> None:
+                a.b.c(5)
+        """
+        after = """
+            import a.b
```
carljm, line 63: Shouldn't this be removed as now-unused?

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,385 @@
...
+        """
+        after = """
+            import c.b
+
+            class Foo(c.b.C):
+                pass
+        """
+
+        self.assertCodemod(
+            before, after, old_name="a.b", new_name="c.b",
```
carljm, line 108: Per comments above, I think we should also add a test where `old_name="a", new_name="c"` with the same before and after as in this test.

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,385 @@
...
+        """
+        after = """
+            from c import b
+
+            class Foo(b.C):
+                pass
+        """
+
+        self.assertCodemod(
+            before, after, old_name="a.b", new_name="c.b",
```
carljm, line 126: And here also we should verify the same before/after with `old_name="a", new_name="b"`

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,385 @@
...
+            before, after, old_name="aa.aaaa", new_name="b.c",
+        )
+
+    def test_import_star_attr(self) -> None:
+        before = """
+            from foo import *
+
+            def baz():
+                foo.bar(5)
+        """
```
carljm, line 329: assuming the fully-qualified name of the thing is `foo.bar`, this initial code is wrong, if you do `from foo import *` you'd have just `bar` in your namespace, not `foo.bar`.

I think this codemod should basically ignore `*` imports. So this test would be correct if we have an unrelated `*` import and then an `import foo`, but we should never make any assumptions about names coming from `*` imports.

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,385 @@
...
+        self.assertCodemod(
+            before, after, old_name="foo.bar", new_name="qux.bar",
+        )
+
+    def test_import_star_obj(self) -> None:
+        before = """
+            from foo import *
+
+            class Bar(foo.Bar):
```
carljm, line 345: again, like above, the combination of `from foo import *` with later use of `foo.Bar` doesn't make sense as working code in the first place (unless the `foo` has a submodule also named `foo` and the fully qualified name of `Bar` is `foo.foo.Bar`)

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,192 @@
...
+        module = updated_node.module
+        if module is not None:
+            imported_module_name = get_full_name_for_node(module)
+            names = original_node.names
+
+            if imported_module_name is None or not isinstance(names, Sequence):
+                return updated_node
+            elif imported_module_name == self.old_name:
+                # Simply rename the imported module on the spot.
+                return updated_node.with_changes(
```
carljm, line 80: I noticed that according to [the coverage report](https://codecov.io/gh/Instagram/LibCST/pull/308/diff) no test actually hits this line?

(Test coverage of this PR is otherwise quite excellent!)

libcst/codemod/commands/rename.py
carljm: Looks like this case is now handled and there's a test for it!

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,192 @@
...
+
+                # If the import statement is exactly equivalent to the old name, replace it.
+                if alias_name == self.old_name:
+                    self.cleanup_imports(node=node, new_module=self.new_name)
+
+    def leave_ImportFrom(
+        self, original_node: cst.ImportFrom, updated_node: cst.ImportFrom
+    ) -> cst.ImportFrom:
+        module = updated_node.module
+        if module is not None:
```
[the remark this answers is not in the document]
josieesh: ah yes, good point

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,192 @@
...
+        self.old_name = old_name
+        self.new_name = new_name
+
+        old_names = self.old_name.rpartition(".")
+        new_names = self.new_name.rpartition(".")
+
+        # Mapping for easy lookup later.
+        self.equivalents: Dict[str, str] = {
+            old_names[2]: new_names[2],
+            "".join(old_names[:2]): new_names[0],
```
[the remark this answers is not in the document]
josieesh: The trailing dot is there for cases such as `foo.foo`. I thought about having two separate dictionaries as well, and maybe it is the way to go. I'll tinker with this and add comments 👍

libcst/codemod/commands/tests/test_rename.py
josieesh: If the passed arguments are `old_name="a"`, `new_name="c"` with the same before and after, shouldn't we expect no change from this codemod as we are only looking for names whose full qualified names match that passed-in arguments? I'll add a test to check that there is indeed no codemod in this case.

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,345 @@
...
+
+    def __init__(self, context: CodemodContext, old_name: str, new_name: str) -> None:
+        super().__init__(context)
+
+        new_module, has_colon, new_mod_or_obj = new_name.rpartition(":")
+        # Exit early if improperly formatted args.
+        if ":" in new_module:
+            raise Exception("Error: `new_name` should contain at most one colon.")
+        if ":" in old_name:
+            raise Exception("Error: `old_name` should not contain any colons.")
```
mvismonte, line 71: I'd probably have a little more specific exception.  Either define a custom one or at least use `ValueError`

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,345 @@
...
+from typing import Callable, Optional, Sequence, Set, Tuple, Union
+
+import libcst as cst
+from libcst.codemod import CodemodContext, VisitorBasedCodemodCommand
+from libcst.codemod.visitors import AddImportsVisitor, RemoveImportsVisitor
+from libcst.helpers import get_full_name_for_node
+from libcst.metadata import QualifiedNameProvider
+
+
+def leave_import_decorator(
```
mvismonte, line 17: What's the value of making this a decorator as opposed to doing this inline or making a helper function?  Not opposed to it, but just want to understand =)

libcst/codemod/commands/tests/test_rename.py
```diff
@@ -0,0 +1,588 @@
...
+#
+# pyre-strict
+
+from libcst.codemod import CodemodTest
+from libcst.codemod.commands.rename import RenameCommand
+
+
+class TestRenameCommand(CodemodTest):
+
+    ONCALL_SHORTNAME = "instagram_server_framework"
```
mvismonte, line 14: We probably don't need this =D

libcst/codemod/commands/rename.py
josieesh: There are so many codepaths that lead to a return statement in both `leave_Import` and `leave_ImportFrom` that I thought it would be easier to intercept it with a decorator :)

libcst/codemod/commands/rename.py
```diff
@@ -0,0 +1,346 @@
...
+        if not has_colon or not new_module:
+            new_module, _, new_mod_or_obj = new_name.rpartition(".")
+
+        self.new_name: str = new_name.replace(":", ".").strip(".")
+        self.new_module: str = new_module.replace(":", ".").strip(".")
+        self.new_mod_or_obj: str = new_mod_or_obj
+
+        if not self.new_mod_or_obj:
+            old_module = old_name
+            old_mod_or_obj = ""
```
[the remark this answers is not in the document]
josieesh: This is the case where a user specifies a `new_name` with a colon at the end, similarly to [this](https://github.com/Instagram/LibCST/pull/308/files#diff-511afb8b206c354e493c782e1c50abadR557) test.

PR #314, opened 2020-06-16
ApplyTypeAnnotationsVisitor: fix default value

hauntsaninja:
## Summary

ApplyTypeAnnotationsVisitor erases default values of positional-only and keyword-only args. See https://github.com/Instagram/MonkeyType/issues/198 for a repro

## Test Plan

Added tests for the relevant issues in test_apply_type_annotations.py

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -398,7 +398,10 @@ class ApplyTypeAnnotationsVisitor(ContextAwareTransformer):
         return annotations.parameters.with_changes(
             params=update_annotation(
                 updated_node.params.params, annotations.parameters.params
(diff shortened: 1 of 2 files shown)


hauntsaninja:
Thanks for the quick review! I was wondering what LibCST's release process is like? This feels like a fairly major regression for MonkeyType (which apparently only recently started using LibCST), so might be nice to cut a new release soon

jimmylai:
> Thanks for the quick review! I was wondering what LibCST's release process is like? This feels like a fairly major regression for MonkeyType (which apparently only recently started using LibCST), so might be nice to cut a new release soon

We currently release occasionally and we do prioritize critical bug fixes. There are currently other fixes on master and @josieesh can make a release later this week.

PR #334, opened 2020-07-08
Only remove trailing comma if the last alias is removed

zsol:
## Summary
`RemoveUnusedImportsVisitor` always removed trailing commas from import statements, which caused the CST to always change. This is undesirable because it causes unnecessary post-processing (like formatting being applied) even if there were no "real" changes to the CST.

This PR changes visitor's behavior so that the trailing comma is only removed if the last alias is actually removed.
## Test Plan

Added unit tests

### DIFF
--- a/libcst/codemod/visitors/tests/test_remove_imports.py
+++ b/libcst/codemod/visitors/tests/test_remove_imports.py
@@ -797,3 +797,17 @@ class TestRemoveImportsCodemod(CodemodTest):
             after,
             RemoveImportTransformer(CodemodContext()).transform_module(module).code,
         )
+
+    def test_remove_comma(self) -> None:
+        """
+        Trailing commas should be removed if and only if the last alias is removed.
+        """
+        before = """
+            from m import (a, b,)
+            import x, y
+        """
+        after = """
+            from m import (b,)
+            import x
+        """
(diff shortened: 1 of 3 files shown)


libcst/codemod/visitors/tests/test_remove_imports.py
```diff
@@ -797,3 +797,17 @@ def visit_ImportFrom(self, node: cst.ImportFrom) -> None:
             after,
             RemoveImportTransformer(CodemodContext()).transform_module(module).code,
         )
+
+    def test_remove_comma(self) -> None:
+        """
+        Trailing commas should be removed if and only if the last alias is removed.
+        """
+        before = """
+            from m import (a, b,)
```
jimmylai, line 806: In the case of `from n import (c, d,)` and remove `("n", "d", None)`, the trailing comma should still exist `from n import (c,)` (to make it still a valid Tuple.
Can you also test it?

zsol: Are you sure? `from n import (c)` is perfectly valid Python (and will be reformatted to `from n import c` by the formatter)

jimmylai: You are right. Both `from a import (b)` and `from a import (b,)` are valid code and the trailing comma should be removed in the codemod.

PR #335, opened 2020-07-10
[Pyre] Handle type subscripts when applying annotations.

MaggieMoss:
## Summary
Just as we special case "Type" when applying annotations here: https://github.com/Instagram/LibCST/blob/master/libcst/codemod/visitors/_apply_type_annotations.py#L166
We need to do the same in Subscripts

## Test Plan

python -m unittest -v libcst.codemod.visitors.tests.test_apply_type_annotations

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -134,10 +134,7 @@ class TypeCollector(cst.CSTVisitor):
 
     def _handle_Subscript(self, node: cst.Subscript) -> cst.Subscript:
         slice = node.slice
-        if (
-            m.matches(node.value, m.Name())
-            and cst.ensure_type(node.value, cst.Name).value == "Type"
-        ):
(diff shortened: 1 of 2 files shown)


libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -133,6 +134,11 @@ def _handle_Index(self, slice: cst.Index, node: cst.Subscript) -> cst.Subscript:
 
     def _handle_Subscript(self, node: cst.Subscript) -> cst.Subscript:
         slice = node.slice
+        if (
+            m.matches(node.value, m.Name())
+            and cst.ensure_type(node.value, cst.Name).value == "Type"
+        ):
```
jimmylai, line 140: Use matcher to match the value to be simpler and easier to read.
```suggestion
        if m.matches(node.value, m.Name(value="type")):
```

MaggieMoss: I did not know you could use this syntax. 😍

PR #343, opened 2020-07-14
Insert imports after module docstring

jimmylai:
## Summary
Module docstring was reordered wrongly.

```
'''Docstring.'''
import typing
```

Became

```
from __future__ import annotations
'''Docstring.'''
import typing
```

Instead of
```
'''Docstring.'''
from __future__ import annotations
import typing
```

## Test Plan
Added a test case.

### DIFF
--- a/libcst/codemod/visitors/_add_imports.py
+++ b/libcst/codemod/visitors/_add_imports.py
@@ -248,7 +248,9 @@ class AddImportsVisitor(ContextAwareTransformer):
     ) -> Tuple[
         List[Union[libcst.SimpleStatementLine, libcst.BaseCompoundStatement]],
         List[Union[libcst.SimpleStatementLine, libcst.BaseCompoundStatement]],
+        List[Union[libcst.SimpleStatementLine, libcst.BaseCompoundStatement]],
     ]:
+        statement_before_import_location = 0
         import_add_location = 0
 
         # never insert an import before initial __strict__ flag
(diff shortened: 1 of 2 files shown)


sagarbadiyani:
Instead of keeping the logic to check for the module docstring inside the `for` loop, can we keep it outside the `for` loop? Should we not be directly checking if the first element of the module body is a docstring before stepping inside the `for` loop, if it is, set `statement_before_import_location = import_add_location = 1`. 

The reason I am asking for this, I found a case in one of the repositories where the docstrings were written outside the function like how we have in Java, obviously, this is unconventional and unpythonic, but my only concern is why are we coupling the logic to check docstring inside the `for` loop? 

Consider a case
```
from __future__ import annotations
from __future__ import division

import abc
from xyz import def

"""Some docstring"""
def some_func():
    pass
```

Let's say we are trying to add an import `from typing import List`, the final code after applying the add imports tranformation will look like
```
from __future__ import annotations
from typing import List
from __future__ import division

import abc
from xyz import def

"""Some docstring"""
def some_func():
    pass
```
This is because the docstring is a part of the body of the module and when the for loop reached there, it set the `import_add_location` to 1, which should not have been the case
And this will break the code because `from __future__ imports` must occur at the beginning of the file.

I will raise a PR for this if you find this satisfactory
@luciawlli @carljm @josieesh @zsol

PR #353, opened 2020-07-28
[RemoveUnusedImports] Support string type annotations

zsol:
## Summary
This PR adds support for detecting imports being used by string type
annotations, as well as imports suppressed by comments.

It breaks up the existing visitor into multiple smaller, single-purpose
visitors, and composes them together.

TODO:

- [x] add docstrings
- [x] document behavioral changes to `RemoveImportsVisitor`

## Test Plan
existing and new tests

### DIFF
--- a/libcst/codemod/commands/remove_unused_imports.py
+++ b/libcst/codemod/commands/remove_unused_imports.py
@@ -33,10 +33,7 @@ class RemoveUnusedImportsCommand(VisitorBasedCodemodCommand):
         "Note: only considers the file in isolation. "
     )
 
-    METADATA_DEPENDENCIES: Tuple[ProviderT] = (
-        PositionProvider,
-        *GatherCommentsVisitor.METADATA_DEPENDENCIES,
-    )
+    METADATA_DEPENDENCIES: Tuple[ProviderT] = (PositionProvider,)
 
     def __init__(self, context: CodemodContext) -> None:
         super().__init__(context)
--- a/libcst/codemod/commands/remove_unused_imports.py
+++ b/libcst/codemod/commands/remove_unused_imports.py
@@ -4,12 +4,14 @@
 # LICENSE file in the root directory of this source tree.
 #
 
+from typing import Union
+
 from libcst import Import, ImportFrom, Module
 from libcst.codemod import VisitorBasedCodemodCommand
-from libcst.codemod.visitors import RemoveImportsVisitor, GatherCommentsVisitor
+from libcst.codemod.visitors import GatherCommentsVisitor, RemoveImportsVisitor
 from libcst.helpers import get_absolute_module_for_import
 from libcst.metadata import PositionProvider
-from typing import Union
+
 
 DEFAULT_SUPPRESS_COMMENT_REGEX = (
     r".*\W(noqa|lint-ignore: ?unused-import|lint-ignore: ?F401)(\W.*)?$"
--- a/libcst/codemod/commands/remove_unused_imports.py
+++ b/libcst/codemod/commands/remove_unused_imports.py
@@ -4,9 +4,16 @@
 # LICENSE file in the root directory of this source tree.
 #
 
-from libcst import Import, ImportFrom
+from libcst import Import, ImportFrom, Module
 from libcst.codemod import VisitorBasedCodemodCommand
-from libcst.codemod.visitors import RemoveImportsVisitor
+from libcst.codemod.visitors import RemoveImportsVisitor, GatherCommentsVisitor
+from libcst.helpers import get_absolute_module_for_import
+from libcst.metadata import PositionProvider
+from typing import Union
+
+DEFAULT_SUPPRESS_COMMENT_REGEX = (
+    r".*\W(noqa|lint-ignore: ?unused-import|lint-ignore: ?F401)(\W.*)?$"
+)
 
 
 class RemoveUnusedImportsCommand(VisitorBasedCodemodCommand):
@@ -17,21 +24,60 @@ class RemoveUnusedImportsCommand(VisitorBasedCodemodCommand):
     to track cross-references between them. If a symbol is imported in a file
     but otherwise unused in it, that import will be removed even if it is being
     referenced from another file.
-
-    It currently doesn't keep track of string type annotations, so an import
-    for `MyType` used only in `def f() -> "MyType"` will be removed.
     """
 
     DESCRIPTION: str = (
         "Remove all imports that are not used in a file. "
         "Note: only considers the file in isolation. "
-        "Note: does not account for usages in string type annotations. "
     )
 
+    METADATA_DEPENDENCIES = (
+        PositionProvider,
+        *GatherCommentsVisitor.METADATA_DEPENDENCIES,
+    )
+
+    def visit_Module(self, node: Module) -> bool:
+        comment_visitor = GatherCommentsVisitor(
+            self.context, DEFAULT_SUPPRESS_COMMENT_REGEX
+        )
+        node.visit(comment_visitor)
+        self._ignored_lines = set(comment_visitor.comments.keys())
+        return True
+
     def visit_Import(self, node: Import) -> bool:
-        RemoveImportsVisitor.remove_unused_import_by_node(self.context, node)
+        self._handle_import(node)
         return False
 
     def visit_ImportFrom(self, node: ImportFrom) -> bool:
-        RemoveImportsVisitor.remove_unused_import_by_node(self.context, node)
+        self._handle_import(node)
         return False
+
+    def _handle_import(self, node: Union[Import, ImportFrom]) -> None:
+        node_start = self.get_metadata(PositionProvider, node).start.line
+        if node_start in self._ignored_lines:
+            return
+
+        for alias in node.names:
+            position = self.get_metadata(PositionProvider, alias)
+            lines = set(range(position.start.line, position.end.line + 1))
+            if lines.isdisjoint(self._ignored_lines):
+                if isinstance(node, Import):
+                    RemoveImportsVisitor.remove_unused_import(
+                        self.context,
+                        module=alias.evaluated_name,
+                        asname=alias.evaluated_alias,
+                    )
+                else:
+                    module_name = get_absolute_module_for_import(
+                        self.context.full_module_name, node
+                    )
+                    if module_name is None:
+                        raise ValueError(
+                            f"Couldn't get absolute module name for {alias.evaluated_name}"
+                        )
+                    RemoveImportsVisitor.remove_unused_import(
+                        self.context,
+                        module=module_name,
+                        obj=alias.evaluated_name,
+                        asname=alias.evaluated_alias,
+                    )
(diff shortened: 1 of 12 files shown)


libcst/codemod/visitors/_gather_unused_imports.py
```diff
@@ -0,0 +1,121 @@
...
+from libcst.codemod.visitors._gather_exports import GatherExportsVisitor
+from libcst.codemod.visitors._gather_comments import GatherCommentsVisitor
+from libcst.codemod.visitors._gather_string_annotation_names import (
+    GatherNamesFromStringAnnotationsVisitor,
+)
+from libcst.metadata.scope_provider import _gen_dotted_names
+
+
+DEFAULT_SUPPRESS_COMMENT_REGEX = (
+    r".*\W(lint-ignore: ?unused-import|noqa|lint-ignore: ?F401)(\W.*)?$"
```
Kronuz, line 23: I suggest this one instead:

```py
r".*\W(noqa[^:]|noqa: ?F401|lint-ignore: ?unused-import|lint-ignore: ?F401)(\W.*)?$"
```

So it ignores things like `# noqa`, `# noqa: F401`, but not `# noqa: IG61`

libcst/codemod/visitors/_gather_unused_imports.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def visit_Module(self, node: cst.Module) -> bool:
+        export_collector = GatherExportsVisitor(self.context)
+        node.visit(export_collector)
+        self._exported_names = export_collector.explicit_exported_objects
+        comment_visitor = GatherCommentsVisitor(
+            self.context,
+            self.context.scratch.get(
+                self.SUPPRESS_COMMENT_REGEX_CONTEXT_KEY, DEFAULT_SUPPRESS_COMMENT_REGEX,
+            ),
```
Kronuz, line 55: The thing with suppressing comments at `GatherUnusedImportsVisitor` level is that in some use cases (linter rules in frameworks that already silent lines with such comments) this ends up in doing more work than needed.

Maybe the suppression of comments could be moved to some other visitor which adds lines to be excluded, so one could choose whether to have them collected or not by using that one.

zsol: I suppose you can set the regex to something that will never match (like `^$`), but we could add an explicit bool option to disable it. Would that help?

Kronuz: Well, faster code is code never executed. Best thing would be if we could put these in a whole different, also reusable, place so they're never executed when they're not needed. Something like `GatherIgnoredLines` or something like that. But that's only a suggestion for those use cases where this isn't really needed.

libcst/codemod/visitors/_gather_string_annotation_names.py
```diff
@@ -0,0 +1,65 @@
...
+    ) -> None:
+        value = node.evaluated_value
+        if value is None:
+            return
+        mod = cst.parse_module(value)
+        extracted_nodes = m.extractall(
+            mod,
+            m.Name(value=m.SaveMatchedNode(m.DoNotCare(), "name"))
+            | m.SaveMatchedNode(m.Attribute(), "attribute"),
+        )
```
Kronuz, line 46: I suggest using this instead:

```py
        extracted_nodes = m.extractall(
            mod,
            m.Name(
                value=m.SaveMatchedNode(m.DoNotCare(), "name"),
                metadata=m.MatchMetadataIfTrue(
                    cst.metadata.ParentNodeProvider,
                    lambda parent: not isinstance(parent, cst.Attribute),
                ),
            )
            | m.SaveMatchedNode(m.Attribute(), "attribute"),
            metadata_resolver=MetadataWrapper(mod, unsafe_skip_copy=True),
        )
```

That fixes the issue about annotations like `"api.http_exceptions.HttpException"` flagging import `from api.http_exceptions import HttpException` as used, when we use `GatherNamesFromStringAnnotationsVisitor` for finding unused imports, because in the original case `HttpException` (as well as `http_exceptions`) are amongst the names returned by this visitor.

libcst/codemod/commands/remove_unused_imports.py
```diff
@@ -17,21 +26,60 @@ class RemoveUnusedImportsCommand(VisitorBasedCodemodCommand):
...
 
     DESCRIPTION: str = (
         "Remove all imports that are not used in a file. "
         "Note: only considers the file in isolation. "
-        "Note: does not account for usages in string type annotations. "
     )
 
+    METADATA_DEPENDENCIES = (
+        PositionProvider,
+        *GatherCommentsVisitor.METADATA_DEPENDENCIES,
```
jimmylai, line 38: We should only need to list the direct metadata dependency of `RemoveUnusedImportsCommand` here.

libcst/codemod/visitors/_gather_string_annotation_names.py
jimmylai: The `MatchMetadataIfTrue` check @Kronuz mentioned can directly remove the inner `Name` from `Attribute`.

jimmylai:
Overall looks good and the test coverage is high. I'll let @Kronuz to do another review and accept to see if he need more changes for reusability.

libcst/codemod/visitors/_gather_string_annotation_names.py
zsol: i forgot to apply this one. I'll do it and add a test case tomorrow morning.

libcst/codemod/commands/remove_unused_imports.py
```diff
@@ -4,9 +4,18 @@
...
+from libcst import Import, ImportFrom, ImportStar, Module
+from libcst.codemod import CodemodContext, VisitorBasedCodemodCommand
+from libcst.codemod.visitors import GatherCommentsVisitor, RemoveImportsVisitor
+from libcst.helpers import get_absolute_module_for_import
+from libcst.metadata import PositionProvider, ProviderT
+
+
+DEFAULT_SUPPRESS_COMMENT_REGEX = (
+    r".*\W(noqa|lint-ignore: ?unused-import|lint-ignore: ?F401)(\W.*)?$"
+)
```
Kronuz, line 18: Also, don't forget about this one as well, I think it's better than the current one:

```py
DEFAULT_SUPPRESS_COMMENT_REGEX = (
    r".*\W(noqa[^:]|noqa: ?F401|lint-ignore: ?unused-import|lint-ignore: ?F401)(\W.*)?$"
)
```

libcst/codemod/commands/remove_unused_imports.py
Kronuz: @jimmylai, so the line `*GatherCommentsVisitor.METADATA_DEPENDENCIES,` should be removed, right?

libcst/codemod/visitors/_gather_unused_imports.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self._string_annotation_names: Set[str] = set()
+        self._exported_names: Set[str] = set()
+        self.unused_imports: Set[
+            Tuple[cst.ImportAlias, Union[cst.Import, cst.ImportFrom]]
+        ] = set()
+
+    def visit_Module(self, node: cst.Module) -> bool:
+        export_collector = GatherExportsVisitor(self.context)
+        node.visit(export_collector)
```
Kronuz, line 40: @jimmylai, just wondering... won't calling `node.visit(export_collector)` and `node.visit(annotation_visitor)` here traverses the tree multiple times? wouldn't this be inefficient?

I was thinking maybe using multiple inheritance instead of this, so we make it that it only traverses the tree once.

libcst/codemod/visitors/_gather_unused_imports.py
```diff
@@ -0,0 +1,97 @@
...
+        self._exported_names = export_collector.explicit_exported_objects
+        annotation_visitor = GatherNamesFromStringAnnotationsVisitor(self.context)
+        node.visit(annotation_visitor)
+        self._string_annotation_names = annotation_visitor.names
+        return True
+
+    @m.visit(
+        m.Import()
```
Kronuz, line 48: How could we add other exceptions to this reusable class, like if we wanted to ignore `import __strict__`, for example?

libcst/codemod/commands/remove_unused_imports.py
zsol: This doesn't match `# noqa` though, which is the most common case

libcst/codemod/visitors/_gather_unused_imports.py
zsol: Yes it will traverse the tree multiple times. I don't think that's a big deal for a codemod like this. Running the codemod on a bunch of large files is still not visibly slower than before. I don't have benchmarks for you though :)

libcst/codemod/visitors/_gather_unused_imports.py
zsol: I think it's best to apply filtering on the results of this visitor (just like how you pointed out that the suppression filtering belongs outside of this class). So I would filter for special module names after running `GatherUnusedImportsVisitor`, and keep the exceptions here to a minimum (i.e. the ones implied by Python itself)

libcst/codemod/commands/remove_unused_imports.py
Kronuz: You are right!

What do you think of:

```py
r” .*\W(noqa|lint-ignore)(: ?(F401|unused-import)|(?= )|$)(\W.*)?$”
```

libcst/codemod/visitors/_gather_unused_imports.py
Kronuz: If we use this in a linter, which will run over and over, while the user types, we want this as fast as possible, otherwise linting times will hit the experience. Although I’m not sure how expensive traversing the tree is, processing time would be exponential, potentially noticeable on big files. What do you think @jimmylai ?

libcst/codemod/commands/remove_unused_imports.py
jimmylai: Yes.

libcst/codemod/visitors/_gather_unused_imports.py
jimmylai: Performance is a concern for lint but not for codemod. In lint, we try to avoid creating a visitor and calling `node.visit(visitor)`.

jimmylai: The `@m.visit`used in this codemod is not supported in the lint framework since we worried about the the pattern can introduce some expensive checks. So we only use `m.matches()` in lint.

zsol:
I'm ok with that but won't have time to implement it this week. How about we land this PR and open a new one for these proposed refactors?

libcst/codemod/visitors/_gather_unused_imports.py
Kronuz: Yes, I agree :)

Kronuz:
@jimmylai, sounds good!

@zsol, thank you for working on this! It'll be super helpful!! 😀

PR #355, opened 2020-08-02
fix: correctly extract wildcard matchers

sk-:
##  Summary
`SaveMatchedNodes` within sequences was not working as expected, the fixes in this PR are:
- Fixes #337 the returned matched sequence now reflect the original match and it's not extended until the end of the original sequence
- Fixes #338 a node matched without a wildcard is now returned as a node and not a sequence 

## Test Plan
Unit tests

Issue #337, opened 2020-07-11
[BUG] Incorrect saved node


sk-:
`SavedMatchedNode` does not seem to work properly with wildcard matchers.

`libcst` version: 0.3.7 (installed with pip)

See the test below:

```py
import libcst.matchers as m
import libcst

node = libcst.parse_expression("[1, 2, 3, 5]")
matcher = m.List(
    elements=[
        m.SaveMatchedNode(m.AtMostN(n=2), "head"),
        m.SaveMatchedNode(m.Element(value=m.Integer(value="3")), "element"),
        m.AtMostN(n=1),
    ]
)
print(m.extract(node, matcher))
```

```
{'head': [Element(
    value=Integer(
        value='1',
        lpar=[],
        rpar=[],
    ),
    comma=Comma(
        whitespace_before=SimpleWhitespace(
            value='',
        ),
        whitespace_after=SimpleWhitespace(
            value=' ',
        ),
    ),
), Element(
    value=Integer(
        value='2',
        lpar=[],
        rpar=[],
    ),
    comma=Comma(
        whitespace_before=SimpleWhitespace(
            value='',
        ),
        whitespace_after=SimpleWhitespace(
            value=' ',
        ),
    ),
), Element(
    value=Integer(
        value='3',
        lpar=[],
        rpar=[],
    ),
    comma=Comma(
        whitespace_before=SimpleWhitespace(
            value='',
        ),
        whitespace_after=SimpleWhitespace(
            value=' ',
        ),
    ),
), Element(
    value=Integer(
        value='5',
        lpar=[],
        rpar=[],
    ),
    comma=MaybeSentinel.DEFAULT,
)], 'element': [Element(
    value=Integer(
        value='3',
        lpar=[],
        rpar=[],
    ),
    comma=Comma(
        whitespace_before=SimpleWhitespace(
            value='',
        ),
        whitespace_after=SimpleWhitespace(
            value=' ',
        ),
    ),
), Element(
    value=Integer(
        value='5',
        lpar=[],
        rpar=[],
    ),
    comma=MaybeSentinel.DEFAULT,
)]}
```

we see two errors with the output:
1) `head` contains the whole list
2) `element` contains the matched element plus the tail

jimmylai:
Nice catch! The bug is reproducible.

PR #356, opened 2020-08-02
fix: SaveMachedNode now matches with trailing empty wildcards

### LINKED ISSUE #336: [BUG] SavedMatchedNode affects matching
Wrapping a matcher with `SaveMatchedNode` sometimes breaks the matching.

`libcst` version: 0.3.7 (installed with pip)

See the test below

```py
import libcst.matchers as m
import libcst

node = libcst.parse_expression("[1, 2, 3]")

matcher = m.List(
    elements=[
        m.SaveMatchedNode(m.AtMostN(n=2), "head"),
        m.SaveMatchedNode(m.Element(value=m.Integer(value="3")), "element"),
        m.AtMostN(n=1),
    ]
)
print(m.extract(node, matcher))
print(m.matches(node, matcher))

matcher = m.List(
    elements=[
        m.SaveMatchedNode(m.AtMostN(n=2), "head"),
        m.SaveMatchedNode(m.Element(value=m.Integer(value="3")), "element"),
        m.SaveMatchedNode(m.AtMostN(n=1), "tail"),
    ]
)
print(m.extract(node, matcher))
print(m.matches(node, matcher))
```

The first matcher fails, while the second succeeds. The only difference is the `tail` `SaveMatchedNode`.

sk-:
## Summary

Note that SaveMatchedNode was already matching leading empty wildcards, however its value is incorrect due to #337. This is why the test for the leading wildcards are failing, and will be so until #355 is merged.

This fixes #336.

## Test Plan

Unit tests


jimmylai:
@sk- can you rebase to fix the conflicts and failed tests?

sk-:
@jimmylai Just rebased it, now the tests pass.

PR #358, opened 2020-08-04
Fix dotted names

Kronuz:
## Summary

When importing things like `import os.path` and using it as `os.path.join("a", "b").lower()`,
references ended up being in the `["os"]` assignment instead of `["os.path"]`.
This fixes the problem by updating the dotted names generator in the scope provider·

## Test Plan

```
tox -e py37
```

### DIFF
--- a/libcst/metadata/tests/test_scope_provider.py
+++ b/libcst/metadata/tests/test_scope_provider.py
@@ -21,6 +21,7 @@ from libcst.metadata.scope_provider import (
     QualifiedNameSource,
     Scope,
     ScopeProvider,
+    _gen_dotted_names,
 )
 from libcst.testing.utils import UnitTest, data_provider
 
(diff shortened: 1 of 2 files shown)


libcst/metadata/tests/test_scope_provider.py
```diff
@@ -203,6 +203,39 @@ def test_dotted_import_access(self) -> None:
...
+        self.assertTrue("os" in scope_of_module)
+
+        a_b_c_assignment = cast(Assignment, list(scope_of_module["os.path"])[0])
+        a_b_c_assignment_references = list(a_b_c_assignment.references)
+        self.assertNotEqual(len(a_b_c_assignment_references), 0)
+        a_b_c_access = a_b_c_assignment_references[0]
+        self.assertEqual(scope_of_module.accesses["os"], set())
+        self.assertEqual(scope_of_module.accesses["os.path"], {a_b_c_access})
+        self.assertEqual(scope_of_module.accesses["os.path.join"], set())
+        self.assertEqual(a_b_c_access.node, attr)
```
jimmylai, line 237: The test case looks good.
Can we also add another test cases to just test the `_gen_dotted_names`? That can keep the test cases simpler and easier to add for future updates.

Kronuz: Done! I added one test for that one

PR #360, opened 2020-08-04
Fix pyre syntax error in tokenizer stub

Kronuz:
## Summary

The pyre stub for the tokenizer module had a syntax error.
Fixing it removes other pyre errors.

## Test Plan

```
pyre check
```


stubs/tokenize.pyi
```diff
@@ -94,4 +94,4 @@ class TokenInfo(Tuple[int, str, Tuple[int, int], Tuple[int, int], int]):
     def __repr__(self) -> str: ...
 
 def detect_encoding(readline: Callable[[], bytes]) -> Tuple[str, Sequence[bytes]]: ...
-def tokenize(Callable[[], bytes]) -> TokenInfo: ...
+def tokenize(readline: Callable[[], bytes]) -> TokenInfo: ...
```
jimmylai, line 97: @Kronuz  Based on the official doc, it should return an Iterator, can you try update the return type?
https://github.com/python/typeshed/blob/5d553c9584eb86793cfa61019ddee5c7b62bc286/stdlib/3/tokenize.pyi#L62

PR #362, opened 2020-08-05
Fix enclosing attribute for attributes in call arguments

Kronuz:
## Summary

Fixes enclosed arguments like `c.d` in `x.y(c.d()).z()` were badly being resolved as `x.y` instead.

This also clarifies the intent in `infer_accesses()` so it no longer shadows variable `name` and also fixes the case where no node is actually found in the scope.

## Test Plan

Run tests

### DIFF
--- a/libcst/codemod/commands/tests/test_remove_unused_imports.py
+++ b/libcst/codemod/commands/tests/test_remove_unused_imports.py
@@ -77,6 +77,16 @@ class RemoveUnusedImportsCommandTest(CodemodTest):
 
         self.assertCodemod(before, after)
 
+    def test_enclosed_attributes(self) -> None:
+        before = """
+            from a.b import c
+            import x
+
+            def foo() -> None:
+                x.y(c.d()).z()
+        """
+        self.assertCodemod(before, before)
+
     def test_access_in_assignment(self) -> None:
         before = """
             from a import b
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -639,7 +639,7 @@ class ScopeVisitor(cst.CSTVisitor):
         self.provider: ScopeProvider = provider
         self.scope: Scope = GlobalScope()
         self.__deferred_accesses: List[Tuple[Access, Optional[cst.Attribute]]] = []
-        self.__top_level_attribute: Optional[cst.Attribute] = None
+        self.__top_level_attribute_stack: List[Optional[cst.Attribute]] = [None]
 
     @contextmanager
     def _new_scope(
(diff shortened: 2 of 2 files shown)


Kronuz:
I added a test which fails without this:

https://github.com/Instagram/LibCST/blob/1c750648a4ab248b4518e40378e52a4ee4bd10a1/libcst/codemod/commands/tests/test_remove_unused_imports.py#L80-L88

I'm thinking the real reason could also be the incorrect detection of `__top_level_attribute` for `c`, detecting `d.x` as "`enclosing_attribute`", and ultimately leading to lose `c`'s access when `name` gets reassigned from `c` to `d`. I'll dig into this hypothesis.

zsol:
Yes that name shadowing was intentional. The first `name` assignment is the default, and then the loop is there to find a better, more specific name.

libcst/codemod/commands/tests/test_remove_unused_imports.py
```diff
@@ -77,6 +77,16 @@ def foo() -> None:
 
         self.assertCodemod(before, after)
 
+    def test_enclosed_attributes(self) -> None:
+        before = """
+            from a.b import c
+            import d
+
+            def foo() -> None:
+                d.x(c(y)).w()
```
jimmylai, line 86: `y` is an undefined name. Shall we just update it to make it valid use case?

jimmylai:
@Kronuz, when @zsol added the variable name shadow, he tries to remove the `x.y` and keep `x.y.z`, that means for Attribute access `x.y.z` we want to associate it with `x.y.z` rather than `x.y` or both. Can you keep this test work while making it also work for your new test case?

> https://github.com/Instagram/LibCST/blob/6a5e7394c023df75bb4d98fc10fb5362b4f3a0bb/libcst/codemod/commands/tests/test_remove_unused_imports.py#L51-L63
>

Kronuz:
@jimmylai, I think I know what the real problem is. I’ll try to fix it!

Kronuz:
@jimmylai, fixed!

jimmylai:
@Kronuz the fix broke a test for rename codemod.
https://app.circleci.com/pipelines/github/Instagram/LibCST/1309/workflows/5b05474b-e290-4de4-9de2-886cd5d911c9/jobs/8019

https://github.com/Instagram/LibCST/blob/030df0625c4509dd606b9f0e7ea5f9d1b7d35da5/libcst/codemod/commands/tests/test_rename.py#L52-L67

You can run `tox -e py37` to run all tests locally before submit changes to save you some time.
There is also a formatting error and you can run `tox -e autofix` to fix it.

Kronuz:
> You can run `tox -e py37` to run all tests locally before submit changes to save you some time.
> There is also a formatting error and you can run `tox -e autofix` to fix it.

Done! sorry about that :P

These fixes will be needed for the linter to work.

PR #366, opened 2020-08-12
replace matchers with explicit visitation in gatherers

zsol:
## Summary
These gatherer visitors are supposed to be reusable, but the matchers make them
* unnecessarily slow
* hard to configure

So this PR eliminates the matchers and uses a more standard visitor API.

## Test Plan

Existing tests

### DIFF
--- a/libcst/codemod/visitors/_gather_string_annotation_names.py
+++ b/libcst/codemod/visitors/_gather_string_annotation_names.py
@@ -28,15 +28,10 @@ class GatherNamesFromStringAnnotationsVisitor(ContextAwareVisitor):
     METADATA_DEPENDENCIES = (QualifiedNameProvider,)
 
     def __init__(
-        self, context: CodemodContext, typing_functions: Optional[Set[str]] = None
+        self, context: CodemodContext, typing_functions: Collection[str] = FUNCS_CONSIDERED_AS_STRING_ANNOTATIONS
     ) -> None:
         super().__init__(context)
-
-        self._typing_functions: Set[str] = (
-            typing_functions
-            if typing_functions is not None
-            else FUNCS_CONSIDERED_AS_STRING_ANNOTATIONS
-        )
+        self._typing_functions: Collection[str] = typing_functions
         self._annotation_stack: List[cst.CSTNode] = []
         #: The set of names collected from string literals.
         self.names: Set[str] = set()
--- a/libcst/codemod/visitors/_gather_string_annotation_names.py
+++ b/libcst/codemod/visitors/_gather_string_annotation_names.py
@@ -3,7 +3,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Set, Union, cast
+from typing import List, Optional, Set, Union, cast
 
 import libcst as cst
 import libcst.matchers as m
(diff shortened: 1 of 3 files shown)


libcst/codemod/visitors/_gather_string_annotation_names.py
```diff
@@ -35,14 +35,48 @@ class GatherNamesFromStringAnnotationsVisitor(ContextAwareVisitor):
...
+        self._typing_functions: Set[str] = (
+            typing_functions
+            if typing_functions is not None
+            else FUNCS_CONSIDERED_AS_STRING_ANNOTATIONS
+        )
+        self._annotation_stack: List[cst.CSTNode] = []
         #: The set of names collected from string literals.
         self.names: Set[str] = set()
 
-    @m.call_if_inside(ANNOTATION_MATCHER)
```
jimmylai, line 44: The defined name ANNOTATION_MATCHER is now unused and can be removed.

libcst/codemod/visitors/_gather_string_annotation_names.py
```diff
@@ -35,14 +35,48 @@ class GatherNamesFromStringAnnotationsVisitor(ContextAwareVisitor):
...
+    def __init__(
+        self, context: CodemodContext, typing_functions: Optional[Set[str]] = None
+    ) -> None:
         super().__init__(context)
 
+        self._typing_functions: Set[str] = (
+            typing_functions
+            if typing_functions is not None
+            else FUNCS_CONSIDERED_AS_STRING_ANNOTATIONS
+        )
```
jimmylai, line 47: Why not just make it as default value?
```suggestion
    def __init__(
        self, context: CodemodContext, typing_functions: Optional[Set[str]] = FUNCS_CONSIDERED_AS_STRING_ANNOTATIONS
    ) -> None:
        super().__init__(context)

        self._typing_functions: Set[str] = typing_functions
```

libcst/codemod/visitors/_gather_unused_imports.py
```diff
@@ -37,9 +36,18 @@ class GatherUnusedImportsVisitor(ContextAwareVisitor):
...
+        self,
+        context: CodemodContext,
+        ignored_modules: Optional[Set[str]] = None,
+        typing_functions: Optional[Set[str]] = None,
+    ) -> None:
         super().__init__(context)
 
+        self._ignored_modules: Set[str] = (
+            ignored_modules if ignored_modules is not None else {"__future__"}
+        )
```
jimmylai, line 49:
```suggestion
    def __init__(
        self,
        context: CodemodContext,
        ignored_modules: Optional[Set[str]] = {"__future__"},
        typing_functions: Optional[Set[str]] = None,
    ) -> None:
        super().__init__(context)
        self._ignored_modules: Set[str] = ignored_modules
```

libcst/codemod/visitors/_gather_string_annotation_names.py
zsol: Mutable values are not good as default values: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

libcst/codemod/visitors/_gather_string_annotation_names.py
```diff
@@ -35,14 +35,48 @@ class GatherNamesFromStringAnnotationsVisitor(ContextAwareVisitor):
...
+
+    def visit_ConcatenatedString(self, node: cst.ConcatenatedString) -> bool:
+        if self._annotation_stack != []:
+            self.handle_any_string(node)
+        return False
+
+    def visit_SimpleString(self, node: cst.SimpleString) -> bool:
+        if self._annotation_stack != []:
+            self.handle_any_string(node)
+        return False
```
Kronuz, line 70: Why not simply check the stack as as boolean:

```suggestion
    def leave_Call(self, original_node: cst.Call) -> None:
        if self._annotation_stack and self._annotation_stack[-1] == original_node:
            self._annotation_stack.pop()

    def visit_ConcatenatedString(self, node: cst.ConcatenatedString) -> bool:
        if self._annotation_stack:
            self.handle_any_string(node)
        return False

    def visit_SimpleString(self, node: cst.SimpleString) -> bool:
        if self._annotation_stack:
            self.handle_any_string(node)
        return False
```

libcst/codemod/visitors/_gather_string_annotation_names.py
jimmylai: I think it's a concern when the parameter is mutated in the function which may be a surprise to the users. In our case, it's used as read only parameter which should be fine.

zsol: If you feel strongly, I can change it; I'm not so sure I'd catch this if a couple months later I make a change that mutates the parameter

jimmylai: How about let type checker to make sure it's not mutated? That way it's both simple and safe.
```suggestion
    def __init__(
        self, context: CodemodContext, typing_functions: Collection[str] = FUNCS_CONSIDERED_AS_STRING_ANNOTATIONS
    ) -> None:
        super().__init__(context)
        self._typing_functions: Collection[str] = typing_functions
```

zsol: That's a great idea, I haven't considered it 👍 Also TIL that `Collection` is immutable

libcst/codemod/visitors/_gather_unused_imports.py
zsol: I'll change these to `Collection`s too tomorrow, once I'm at my computer, and merge this PR afterwards.

libcst/codemod/visitors/_gather_string_annotation_names.py
jimmylai: Yes, we actually run an automated codemod continuously to refine `Dict` as `Mapping` and `Set` as `Collection` to keep the type annotations more precise.

PR #370, opened 2020-08-12
Handle more cases for __all__

Kronuz:
## Summary

This flags exports of undefined names and safely handles more cases for `__all__`.

## Test Plan

It should now handle:

```py
import os
__all__, _ = ["os"], []
```

### DIFF
--- a/libcst/codemod/visitors/_gather_exports.py
+++ b/libcst/codemod/visitors/_gather_exports.py
@@ -3,13 +3,12 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 #
-from typing import Set
+from typing import Set, Union
 
-import libcst
-import libcst.matchers as m
+import libcst as cst
 from libcst.codemod._context import CodemodContext
 from libcst.codemod._visitor import ContextAwareVisitor
-from libcst.helpers import ensure_type, get_full_name_for_node
+from libcst.helpers import get_full_name_for_node
 
 
 class GatherExportsVisitor(ContextAwareVisitor):
@@ -32,7 +31,8 @@ class GatherExportsVisitor(ContextAwareVisitor):
 
     def __init__(self, context: CodemodContext) -> None:
         super().__init__(context)
-        # Track any re-exported objects in an __all__ reference
+        # Track any re-exported objects in an __all__ reference and whether
+        # they're defined or not
         self.explicit_exported_objects: Set[str] = set()
 
         # Presumably at some point in the future it would be useful to grab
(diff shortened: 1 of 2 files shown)


libcst/codemod/visitors/_gather_exports.py
```diff
@@ -43,68 +46,86 @@ def __init__(self, context: CodemodContext) -> None:
...
+    def _handle_string_export(
+        self, node: Union[cst.SimpleString, cst.ConcatenatedString]
+    ) -> None:
+        if self._in_assigned_export:
+            name = node.evaluated_value
+            if name is None:
+                return
+            scope = self.get_metadata(cst.metadata.ScopeProvider, node, None)
+            if scope:
+                defined = name in scope.assignments
```
zsol, line 130: This will be `False` in a scenario like
```
from .submodule import *
__all__ = ["SomethingFromSubmodule"]
```

zsol: The reason I bring this up is because star imports are more common in cases when `__all__` is defined, which is precisely the case this visitor is useful in.

Kronuz: That's too bad ...maybe we want to just not set those to `False` if there are star imports... there is value in knowing exported symbols are undefined in the other cases

libcst/codemod/visitors/_gather_exports.py
```diff
@@ -22,17 +21,18 @@ class GatherExportsVisitor(ContextAwareVisitor):
     After visiting a module the following attributes will be populated:
 
      explicit_exported_objects
-      A sequence of strings representing objects that the module exports
-      directly. Note that when ``__all__`` is absent, this attribute does not
-      store default exported objects by name.
+      A mapping of strings to boolean representing objects that the module exports
+      directly and whether they're defined. Note that when ``__all__`` is absent,
+      this attribute does not store default exported objects by name.
```
jimmylai, line 26: `explicit_exported_objects` is still a `Set[str]` instead of a mapping after this change.

libcst/codemod/visitors/_gather_exports.py
```diff
@@ -43,68 +43,87 @@ def __init__(self, context: CodemodContext) -> None:
...
+
+    def leave_Tuple(self, original_node: cst.Tuple) -> None:
+        self._is_assigned_export.discard(original_node)
+        self._in_assigned_export.discard(original_node)
+
+    def visit_Set(self, node: cst.Set) -> bool:
+        if node in self._is_assigned_export:
+            self._in_assigned_export.add(node)
+            return True
+        return False
```
jimmylai, line 108: This line is not covered by tests.
It can be easily covered by having a test case like

```
__all__, _ = ["bar", "baz"], ["biz"], {"foo"}
```

PR #372, opened 2020-08-13
Annotation access

Kronuz:
## Summary

Access references with this type come from an annotation.

## Test Plan

Run tests.

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -63,11 +63,14 @@ class Access:
     #: assignment.
     scope: "Scope"
 
+    is_annotation: bool
+
     __assignments: Set["BaseAssignment"]
 
-    def __init__(self, node: cst.Name, scope: "Scope") -> None:
+    def __init__(self, node: cst.Name, scope: "Scope", is_annotation: bool) -> None:
         self.node = node
         self.scope = scope
+        self.is_annotation = is_annotation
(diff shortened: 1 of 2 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -85,6 +88,10 @@ def record_assignments(self, assignments: Set["BaseAssignment"]) -> None:
         self.__assignments |= assignments
 
 
+class AnnotationAccess(Access):
+    pass
+
+
```
jimmylai, line 94: AnnotationAccess is not used and can be removed.

libcst/metadata/scope_provider.py
```diff
@@ -699,13 +707,19 @@ def visit_Call(self, node: cst.Call) -> Optional[bool]:
...
+        self.__in_annotation.discard(original_node)
+
     def visit_Name(self, node: cst.Name) -> Optional[bool]:
         # not all Name have ExpressionContext
         context = self.provider.get_metadata(ExpressionContextProvider, node, None)
         if context == ExpressionContext.STORE:
             self.scope.record_assignment(node.value, node)
         elif context in (ExpressionContext.LOAD, ExpressionContext.DEL):
-            access = Access(node, self.scope)
+            access = Access(node, self.scope, is_annotation=bool(self.__in_annotation))
```
jimmylai, line 722: So clever!

PR #373, opened 2020-08-14
Handle string annotations in ScopeProvider

Kronuz:
## Summary

This adds support for string annotations to ScopeProvider

## Test Plan

Run tests

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -699,9 +699,20 @@ class ScopeVisitor(cst.CSTVisitor):
 
     def visit_Call(self, node: cst.Call) -> Optional[bool]:
         self.__top_level_attribute_stack.append(None)
+        if any(
+            qn.name == "typing.TypeVar"
+            for qn in self.scope.get_qualified_names_for(node)
+        ):
+            node.func.visit(self)
+            self.__in_annotation.add(node)
+            for arg in node.args[1:]:
+                arg.visit(self)
+            return False
+        return True
 
     def leave_Call(self, original_node: cst.Call) -> None:
         self.__top_level_attribute_stack.pop()
+        self.__in_annotation.discard(original_node)
 
     def visit_Annotation(self, node: cst.Annotation) -> Optional[bool]:
         self.__in_annotation.add(node)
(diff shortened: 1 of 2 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -699,22 +700,66 @@ def visit_Attribute(self, node: cst.Attribute) -> Optional[bool]:
...
+    def visit_Subscript(self, node: cst.Subscript) -> Optional[bool]:
+        if any(
+            qn.name == "typing.Literal"
+            for qn in self.scope.get_qualified_names_for(node.value)
+        ):
+            node.value.visit(self)
+            self.__in_literal.add(node)
+            for s in node.slice:
+                s.visit(self)
+            return False
```
jimmylai, line 752: Can `Literal` contains a name reference as string?
https://docs.python.org/3/library/typing.html#typing.Literal
It doesn't seem like it.

libcst/metadata/scope_provider.py
```diff
@@ -699,22 +700,66 @@ def visit_Attribute(self, node: cst.Attribute) -> Optional[bool]:
...
+            qn.name == "typing.TypeVar"
+            for qn in self.scope.get_qualified_names_for(node)
+        ):
+            node.func.visit(self)
+            for arg in node.args[:1]:
+                arg.visit(self)
+            self.__in_annotation.add(node)
+            for arg in node.args[1:]:
+                arg.visit(self)
+            return False
```
jimmylai, line 713: Why do we visit the `args` twice?
The first parameter of TypeVar is just a name of the defined type.
We should only analyze the other parameters (`args[1:]`).

libcst/metadata/scope_provider.py
Kronuz: Yes, we want to ignore the strings inside a `Literal`, so we ignore those by tracking if we're in a `Literal` using `__in_literal`.

jimmylai: NVM, figured it out.

libcst/metadata/scope_provider.py
Kronuz: The first argument of `typing.TypeVar` is the name, and is always a non-typed string, the following arguments are restriction types either as positional arguments or as `bound=`.

libcst/metadata/scope_provider.py
Kronuz:
```py
import foo
def function(param: Literal["foo", "bar"]) -> None:
    pass
```

If we don't ignore the strings inside `Literal`, in the above example the imported `foo` would have a wrong access attributed to it.

libcst/metadata/scope_provider.py
Kronuz: Oh, got it! yes... we can avoid analyzing the first argument, you're right!

libcst/metadata/scope_provider.py
```diff
@@ -699,22 +699,58 @@ def visit_Attribute(self, node: cst.Attribute) -> Optional[bool]:
...
+    def _handle_string_annotation(
+        self, node: Union[cst.SimpleString, cst.ConcatenatedString]
+    ) -> None:
+        if self.__in_annotation:
+            value = node.evaluated_value
+            if value:
+                mod = cst.parse_module(value)
+                mod.visit(self)
+
+    def visit_Subscript(self, node: cst.Subscript) -> Optional[bool]:
```
jimmylai, line 740: Return False at the end?

libcst/metadata/tests/test_scope_provider.py
```diff
@@ -1013,22 +1013,56 @@ def g():
...
+
+        assignment = list(scope["B"])[0]
+        self.assertIsInstance(assignment, Assignment)
+        self.assertEqual(len(assignment.references), 1)
+        references = list(assignment.references)
+        self.assertTrue(references[0].is_annotation)
+
+        assignment = list(scope["C"])[0]
+        self.assertIsInstance(assignment, Assignment)
+        self.assertEqual(len(assignment.references), 0)
```
jimmylai, line 1048: We should test that the `Access` of C has `is_annotation` == `True`.

jimmylai: Nvm, the C in Literal is not an Access.

Kronuz: There won't be `Access` of `C`, because `"C"` a string literal.

libcst/metadata/scope_provider.py
Kronuz: Which function?

As of `_handle_string_annotation()`? it's because `visit_SimpleString()` and `visit_ConcatenatedString()` are the ones using it, and they do return `False`.

And `visit_Subscript()` returns `False` only when it's a `Literal[...]`, other cases (such as `Union[...]`) should continue to analyze things further.

jimmylai: I mean `visit_Subscript`. That makes sense, let's return True to make it explicit?

libcst/metadata/scope_provider.py
```diff
@@ -699,22 +699,59 @@ def visit_Attribute(self, node: cst.Attribute) -> Optional[bool]:
...
+        return False
+
+    def _handle_string_annotation(
+        self, node: Union[cst.SimpleString, cst.ConcatenatedString]
+    ) -> None:
+        if self.__in_annotation:
+            value = node.evaluated_value
+            if value:
+                mod = cst.parse_module(value)
+                mod.visit(self)
```
zsol, line 738: There's a subtle problem with this approach that I only just discovered. Running `self` on another module will mean we'll collect CST nodes that don't appear in the original CST. That means anyone using these collected nodes will not be able to meaningfully use them (for example to look up other metadata for them, like position). I'll open a separate issue.

PR #380, opened 2020-08-25
Upgrade dev tools (Black/Flake8/isort) and read install requirements from requirements.txt

jimmylai:
## Summary
Upgrade dev tools (Black/Flake8/isort) and fix new violations.
Read install requirements from requirements.txt
So we don't forget update them and they'll always in sync.

## Test Plan
Existing build jobs.


setup.py
```diff
@@ -46,12 +46,7 @@
...
-    install_requires=[
-        "dataclasses; python_version < '3.7'",
-        "typing_extensions >= 3.7.4.2",
-        "typing_inspect >= 0.4.0",
-        "pyyaml >= 5.2",
-    ],
+    install_requires=[dep.strip() for dep in open("requirements.txt").readlines()],
     extras_require={
         "dev": [
             "black",
```
zsol, line 52: we could do the same for these

jimmylai: Done.

zsol:
Seems like tons of stuff are failing after the `requirements-dev.txt` change. I wonder if that file has the right versions for things

PR #392, opened 2020-09-23
Removing an import alias with a trailing standalone comment should preserve the comment

zsol:
## Summary
When removing import aliases, comments often get mangled (see test cases for details). This PR fixes comment handling.

## Test Plan
Added test case


jimmylai (approved):
LGTM

Some of the logic can probably be built as reusable helpers. Especially moving the comments in `ParenthesizedWhitespace` is also pretty common in other codemod use case like remove an argument from a call.

PR #396, opened 2020-09-28
Use correct type for AugAssign and AnnAssign target

### LINKED ISSUE #391: Incorrect typing of {Ann|Aug}Assign.target
The typing of `{Ann,Aug}Assign.target` is currently `BaseExpression` where it seems like it should probably `AssignTarget` or minimally `BaseAssignTargetExpression`. 

```Python
# Correctly errors
libcst.parse_module('x+1 = 1').validate_types_deep()

# Does not error
libcst.parse_module('x+1 : int = 1').validate_types_deep()
libcst.parse_module('x+1 += 1').validate_types_deep()
```

cdonovick:
## Summary
Closes #391 

Changes to `libcst/matchers/__init__.py` performed by `tox -e codegen`

## Test Plan
I couldn't find any relevant tests to extend but would be happy to write one.

### DIFF
--- a/libcst/_nodes/tests/test_assign.py
+++ b/libcst/_nodes/tests/test_assign.py
@@ -119,6 +119,28 @@ class AssignTest(CSTNodeTest):
     def test_invalid(self, **kwargs: Any) -> None:
         self.assert_invalid(**kwargs)
 
+    @data_provider(
+        (
+            {
+                "get_node": (
+                    lambda: cst.Assign(
+                        targets=[
+                            cst.BinaryOperation(
+                                left=cst.Name("x"),
+                                operator=cst.Add(),
+                                right=cst.Integer("1"),
+                            ),
+                        ],
+                        value=cst.Name("y"),
+                    )
+                ),
+                "expected_re": "Expected an instance of .*statement.AssignTarget.*",
+            },
+        )
+    )
+    def test_invalid_types(self, **kwargs: Any) -> None:
+        self.assert_invalid_types(**kwargs)
+
 
 class AnnAssignTest(CSTNodeTest):
     @data_provider(
@@ -284,6 +306,30 @@ class AnnAssignTest(CSTNodeTest):
     def test_invalid(self, **kwargs: Any) -> None:
         self.assert_invalid(**kwargs)
 
+    @data_provider(
+        (
+            {
+                "get_node": (
+                    lambda: cst.AnnAssign(
+                        target=cst.BinaryOperation(
+                            left=cst.Name("x"),
+                            operator=cst.Add(),
+                            right=cst.Integer("1"),
+                        ),
+                        annotation=cst.Annotation(cst.Name("int")),
+                        equal=cst.AssignEqual(),
+                        value=cst.Name("y"),
+                    )
+                ),
+                "expected_re": (
+                    "Expected an instance of .*BaseAssignTargetExpression.*"
+                ),
+            },
+        )
+    )
+    def test_invalid_types(self, **kwargs: Any) -> None:
+        self.assert_invalid_types(**kwargs)
+
 
 class AugAssignTest(CSTNodeTest):
     @data_provider(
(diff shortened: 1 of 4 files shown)


jimmylai:
This change looks great. It makes LibCST parsing more correctly just like AST.

Can you add invalid test cases to `AnnAssignTest ` and `AugAssignTest `?

https://github.com/Instagram/LibCST/blob/c023fa7c4caff3fd2b3946080f9a58b539b10363/libcst/_nodes/tests/test_assign.py#L269-L285

jimmylai (changes requested):
Other than missing test cases, LGTM.

cdonovick:
The suggested pattern does not quite work as `*Assign._validate` doesn't check types just existence/length of arguments.  I tried using:
 ```Python
"get_node": (
    # x + 1 : int = y 
    lambda: cst.AnnAssign(...).validate_types_shallow()
)
```
But this generates a `TypeError` instead of a `CSTValidationError`.

I will add a new test methods (`test_invalid_types` / `assert_invalid_types`) that will call `get_node().validate_types_shallow()` similar to the to `test_invalid` / `assert_invalid`  pattern.

**Comment by cdonovick** [comment]
Hmmm not sure what pyre is complaining about.  I tried to reproduce locally but I get tons of errors on master so I must have something configured wrong.

File: libcst/_nodes/tests/test_assign.py [hunk 1]
```diff
@@ -119,6 +119,28 @@ def test_valid(self, **kwargs: Any) -> None:
     def test_invalid(self, **kwargs: Any) -> None:
         self.assert_invalid(**kwargs)
 
+    @data_provider(
+        (
+            {
+                "get_node": (
+                    lambda: cst.Assign(
+                        targets=[
```
zsol, line 127:
```suggestion
                        # pyre-fixme: Incompatible parameter type [6]
                        targets=[
```

libcst/_nodes/tests/test_assign.py
```diff
@@ -284,6 +306,30 @@ def test_valid(self, **kwargs: Any) -> None:
     def test_invalid(self, **kwargs: Any) -> None:
         self.assert_invalid(**kwargs)
 
+    @data_provider(
+        (
+            {
+                "get_node": (
+                    lambda: cst.AnnAssign(
+                        target=cst.BinaryOperation(
```
zsol, line 314:
```suggestion
                        # pyre-fixme: Incompatible parameter type [6]
                        target=cst.BinaryOperation(
```

libcst/_nodes/tests/test_assign.py
```diff
@@ -362,3 +408,26 @@ class AugAssignTest(CSTNodeTest):
     )
     def test_valid(self, **kwargs: Any) -> None:
         self.validate_node(**kwargs)
+
+    @data_provider(
+        (
+            {
+                "get_node": (
+                    lambda: cst.AugAssign(
+                        target=cst.BinaryOperation(
```
zsol, line 417:
```suggestion
                        # pyre-fixme: Incompatible parameter type [6]
                        target=cst.BinaryOperation(
```

zsol (commented):
These should silence the pyre errors (since we're intentionally constructing these types incorrectly)

libcst/_nodes/tests/test_assign.py
jimmylai: > Hmmm not sure what pyre is complaining about. I tried to reproduce locally but I get tons of errors on master so I must have something configured wrong.
> 
> Edit:
> Possibly my intentional construction of ill-typed nodes? Is there a way to get pyre to skip something?

Pyre did the right thing to complain about the target node are not of `BaseAssignTargetExpression`.
Yes, we can ignore the pyre errors in the test cases as @zsol suggested.
Instead of `pyre-fixme`, use `pyre-ignore` is better because we don't plan to fix them.

jimmylai (approved):
@cdonovick Looks good now. Thanks for contributing!

PR #397, opened 2020-09-29
fix RemoveImportsVisitor crash when ImportAlias is inserted without comma

zsol:
## Summary

The comment preserving logic introduced in #392 assumed that in an `ImportFrom` node, `ImportAlias`es have a `comma` property (except for the last one). That's only true if the `ImportFrom` node is parsed from actual source, but isn't necessarily true if it's constructed manually.

This PR adds a defensive check for the existence of a proper `Comma` node, and in case it doesn't exist (because it's a `MaybeSentinel`), the logic now just moves over the `Comma` node from the alias that's about to be removed.

## Test Plan

Added a test case that adds a new `ImportAlias` node manually (with `comma=MaybeSentinel.DEFAULT`), then removes the next alias.

### DIFF
--- a/libcst/codemod/visitors/_remove_imports.py
+++ b/libcst/codemod/visitors/_remove_imports.py
@@ -374,12 +374,19 @@ class RemoveImportsVisitor(ContextAwareTransformer):
                 if len(names_to_keep) != 0:
                     # there is a previous import alias
                     prev = names_to_keep[-1]
-                    names_to_keep[-1] = prev.with_deep_changes(
-                        whitespace_after=_merge_whitespace_after(
-                            prev.comma.whitespace_after,
-                            comma.whitespace_after,
+                    if isinstance(prev.comma, cst.Comma):
+                        prev = prev.with_deep_changes(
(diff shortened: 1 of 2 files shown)


libcst/codemod/visitors/_remove_imports.py
```diff
@@ -374,12 +374,19 @@ def _process_importfrom_aliases(
...
-                        whitespace_after=_merge_whitespace_after(
-                            prev.comma.whitespace_after,
-                            comma.whitespace_after,
+                    if isinstance(prev.comma, cst.Comma):
+                        prev = prev.with_deep_changes(
+                            whitespace_after=_merge_whitespace_after(
+                                prev.comma.whitespace_after,
+                                comma.whitespace_after,
+                            )
                         )
```
jimmylai, line 383: This is not tested. Can you add a test case?

https://codecov.io/gh/Instagram/LibCST/src/9d76eadaa3b5a52cadeb75960b587b446a0f508f/libcst/codemod/visitors/_remove_imports.py

jimmylai (approved):
Other than the comment, LGTM.

PR #401, opened 2020-10-09
Support string annotations for type aliases

Kronuz:
## Summary

Add support for string annotations in type aliases.

## Test Plan

Run tests.

This will support things like:

```py
from typing import Dict, Union
from video import VideoType
VideoDict = Dict[str, Union[str, int, "VideoType"]]
```

And get a properly referenced `VideoType`.


jimmylai (approved):
Other than the formatting error. LGTM.
The error can be fixed by running `tox -e autofix`.

PR #402, opened 2020-10-12
Codemod CLI multiprocessing simplification

akx:
## Summary

This PR refactors some of the codemod CLI internals to:

* use a simple "execution configuration" dataclass instead of passing a number of kwargs around
* use the default `multiprocessing.Pool()` instead of managing processes and queues by hand
* simplify the special-casing of a single file being processed by `parallel_exec_transform_with_prettyprint` to use a synchronous stub "multiprocessing" Pool instead

These changes pave the way to surfacing `_execute_transform()` (née `_parallel_exec_process_stub()`) as a public API, which would be useful for writing wrapper tools (like `django-codemod`, via which I came across LibCST in https://github.com/browniebroke/django-codemod/pull/233) that could opt-in to doing either parallel or serial codemod processing as they like.

## Test Plan

This is an internal change; no tests need editing.

### DIFF
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -15,10 +15,9 @@ import sys
 import time
 import traceback
 from dataclasses import dataclass, replace
-from multiprocessing import Process, Queue, cpu_count
+from multiprocessing import Pool, cpu_count
 from pathlib import Path
-from queue import Empty
-from typing import AnyStr, Dict, List, Optional, Sequence, Set, Union, cast
+from typing import AnyStr, List, Optional, Sequence, Tuple, Union, cast
 
 from libcst import PartialParserConfig, parse_module
 from libcst.codemod._codemod import Codemod
@@ -382,20 +381,6 @@ def _execute_transform(  # noqa: C901
         )
 
 
-def _parallel_exec_process_stub(
-    result_queue: "Queue[ParallelExecResult]",
-    transformer: Codemod,
-    filename: str,
-    config: ExecutionConfig,
-) -> None:
-    result = _execute_transform(
-        transformer=transformer,
-        filename=filename,
-        config=config,
-    )
-    result_queue.put(result)
-
-
 class Progress:
     ERASE_CURRENT_LINE: str = "\r\033[2K"
 
@@ -528,6 +513,18 @@ class ParallelTransformResult:
     skips: int
 
 
+# Unfortunate wrapper required since there is no `istarmap_unordered`...
+def _execute_transform_wrap(
+    args: Tuple[Codemod, str, ExecutionConfig],
+) -> ExecutionResult:
+    transformer, filename, config = args
+    return _execute_transform(
+        transformer=transformer,
+        filename=filename,
+        config=config,
+    )
+
+
 def parallel_exec_transform_with_prettyprint(  # noqa: C901
     transform: Codemod,
     files: Sequence[str],
@@ -610,10 +607,6 @@ def parallel_exec_transform_with_prettyprint(  # noqa: C901
         )
     print("Executing codemod...", file=sys.stderr)
 
-    # We place results in this queue inside _parallel_exec_process_stub
-    # so that we can control when things get printed to the console.
-    queue = Queue()
-
     config = ExecutionConfig(
         repo_root=repo_root,
         unified_diff=unified_diff,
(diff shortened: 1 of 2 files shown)


libcst/codemod/_cli.py
```diff
@@ -277,91 +277,76 @@ def _parallel_exec_process_stub(  # noqa: C901
...
         )
 
         # Run the transform, bail if we failed or if we aren't formatting code
         try:
             input_tree = parse_module(
                 oldcode,
                 config=(
-                    PartialParserConfig(python_version=python_version)
-                    if python_version is not None
+                    PartialParserConfig(python_version=str(config.python_version))
```
zsol, line 288: Could the `str()` be dropped here?

libcst/codemod/_cli.py
```diff
@@ -373,35 +358,41 @@ def _parallel_exec_process_stub(  # noqa: C901
...
+            transform_result=TransformFailure(
+                error=ex,
+                traceback_str=traceback.format_exc(),
+                warning_messages=transformer.context.warnings,
+            ),
         )
 
 
+# Unfortunate wrapper required since there is no `istarmap_unordered`...
+def _execute_transform_wrap(
```
zsol, line 385: nit: I would probably put this near the callsite

libcst/codemod/_cli.py
```diff
@@ -373,35 +358,41 @@ def _parallel_exec_process_stub(  # noqa: C901
...
+    args: Tuple[Codemod, str, ExecutionConfig],
+) -> ExecutionResult:
+    transformer, filename, config = args
+    return _execute_transform(
+        transformer=transformer,
+        filename=filename,
+        config=config,
+    )
```
zsol, line 393: isn't this equivalent to the following?
```
lambda args: _execute_transform(*args)
```

libcst/codemod/_cli.py
```diff
@@ -534,6 +525,21 @@ class ParallelTransformResult:
     skips: int
 
 
+class DummyPool:
+    def __init__(self, processes: Optional[int] = None) -> None:
+        self.processes = processes
+
+    def imap_unordered(self, func, iterable, chunksize=None):  # type: ignore
```
zsol, line 532:
```suggestion
    def imap_unordered(self, func: Callable[[ArgT], RetT], iterable: Iterable[ArgT], chunksize: Optional[int]=None) -> Generator[RetT, None, None]:  
```

where
```
RetT = TypeVar("RetT")
ArgT = TypeVar("ArgT")
```

libcst/codemod/_cli.py
```diff
@@ -534,6 +525,21 @@ class ParallelTransformResult:
...
+        for args in iterable:
+            yield func(args)
+
+    def __enter__(self) -> "DummyPool":
+        return self
+
+    def __exit__(self, *args) -> None:  # type: ignore
```
zsol, line 539:
```suggestion
    def __exit__(
        self,
        exc_type: Optional[Type[Exception]],
        exc: Optional[Exception],
        tb: Optional[TracebackType],
    ) -> None:
```

libcst/codemod/_cli.py
```diff
@@ -616,171 +622,68 @@ def parallel_exec_transform_with_prettyprint(  # noqa: C901
...
-                joinable_processes.add(process)
-
-                # Start it, bookkeep that we did
-                process.start()
-                processes_started += 1
-            except KeyboardInterrupt:
-                interrupted = True
-                continue
 
+    with pool_impl(processes=jobs) as p:  # type: ignore
```
zsol, line 658: Hmm what is this type ignore suppressing?

zsol (approved):
Sorry reviewing took so long. This looks good to me, I have a bunch of minor points before merging though.

Have you tested the case when you kill the cli during processing with `Ctrl-c`?

libcst/codemod/_cli.py
akx: Currently yes, but in the case someone happens to reorder arguments or so, things would break in unexpected ways. (Explicit is better than implicit and all that.)

I'll change it to a dict so we can use `**kwargs`.

libcst/codemod/_cli.py
akx:
```
`Union[typing.Callable(Pool)[[Named(processes, Optional[int], default), Named(initializer, Optional[typing.Callable[..., typing.Any]], default), Named(initargs, typing.Iterable[typing.Any], default), Named(maxtasksperchild, Optional[int], default)], multiprocessing.pool.Pool], typing.Callable(type.__call__)[[Variable(typing.Any), Keywords(typing.Any)], typing.Any]]` is not a function.
```

libcst/codemod/_cli.py
akx: Unfortunately not – Pyre would complain:

```
libcst/codemod/_cli.py:289:40 Incompatible parameter type [6]: Expected `Union[libcst._parser.types.config.AutoConfig, str]` for 1st parameter `python_version` to call `libcst._parser.types.config.PartialParserConfig.__init__` but got `Optional[str]`.
```

akx:
> Have you tested the case when you kill the cli during processing with `Ctrl-c`?

Empirically, yeah – seems to work just as the old version did 😊 

Thanks for the review work!

akx:
@zsol The CI errors in lint seem to be unrelated to any changes made in this PR 😕

zsol:
Thanks for the contribution!

PR #406, opened 2020-10-21
fix is_annotation for types used in classdef base and assign value

luciawlli:
## Summary
fix is_annotation for types used in classdef base and assign value
discussion
## Test Plan
tox -e py37
tox -e autofix

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -647,7 +647,7 @@ class ScopeVisitor(cst.CSTVisitor):
             Union[cst.Call, cst.Annotation, cst.Subscript]
         ] = set()
         self.__in_ignored_subscript: Set[cst.Subscript] = set()
-        self.__ignore_annotation: bool = False
+        self.__ignore_annotation: int = 0
 
     @contextmanager
     def _new_scope(
@@ -706,6 +706,7 @@ class ScopeVisitor(cst.CSTVisitor):
         qnames = self.scope.get_qualified_names_for(node)
         if any(qn.name in {"typing.NewType", "typing.TypeVar"} for qn in qnames):
             node.func.visit(self)
+            self.__in_annotation.add(node)
             for arg in node.args[1:]:
                 arg.visit(self)
             return False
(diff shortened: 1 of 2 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -716,11 +716,18 @@ def leave_Call(self, original_node: cst.Call) -> None:
...
 
     def visit_Annotation(self, node: cst.Annotation) -> Optional[bool]:
-        self.__in_annotation.add(node)
+        if not self.__ignore_annotation:
+            self.__in_annotation.add(node)
 
     def leave_Annotation(self, original_node: cst.Annotation) -> None:
         self.__in_annotation.discard(original_node)
 
+    def visit_Assign_value(self, original_node: cst.Assign) -> None:
```
zsol, line 725:
```suggestion
    def visit_Assign_value(self, node: cst.Assign) -> None:
```

libcst/metadata/scope_provider.py
```diff
@@ -817,6 +827,12 @@ def visit_ClassDef(self, node: cst.ClassDef) -> Optional[bool]:
                 statement.visit(self)
         return False
 
+    def visit_ClassDef_bases(self, node: cst.ClassDef) -> None:
+        self.__ignore_annotation = True
+
+    def leave_ClassDef_bases(self, node: cst.ClassDef) -> None:
```
zsol, line 833:
```suggestion
    def leave_ClassDef_bases(self, original_node: cst.ClassDef) -> None:
```

libcst/metadata/scope_provider.py
```diff
@@ -647,6 +647,7 @@ def __init__(self, provider: "ScopeProvider") -> None:
             Union[cst.Call, cst.Annotation, cst.Subscript]
         ] = set()
         self.__in_ignored_subscript: Set[cst.Subscript] = set()
+        self.__ignore_annotation: bool = False
```
zsol, line 650: This should probably be a stack or a counter instead of a simple bool to keep track of nesting. Let's add a nested test case while we're at it.

libcst/metadata/scope_provider.py
```diff
@@ -740,7 +747,10 @@ def _handle_string_annotation(
 
     def visit_Subscript(self, node: cst.Subscript) -> Optional[bool]:
         qnames = self.scope.get_qualified_names_for(node.value)
-        if any(qn.name.startswith(("typing.", "typing_extensions.")) for qn in qnames):
+        if (
+            any(qn.name.startswith(("typing.", "typing_extensions.")) for qn in qnames)
+            and not self.__ignore_annotation
+        ):
```
zsol, line 753: We still want to do the work in `_handle_string_annotation` for these cases, with the only difference being `Access` nodes should have `is_annotation=False`.

zsol (changes requested):

libcst/metadata/scope_provider.py
```diff
@@ -721,6 +722,12 @@ def visit_Annotation(self, node: cst.Annotation) -> Optional[bool]:
     def leave_Annotation(self, original_node: cst.Annotation) -> None:
         self.__in_annotation.discard(original_node)
 
+    def visit_Assign_value(self, node: cst.Assign) -> None:
+        self.__ignore_annotation += 1
+
+    def leave_Assign_value(self, node: cst.Assign) -> None:
+        self.__ignore_annotation -= 1
+
```
jimmylai, line 730: Handle Assign differently doesn't seem right.
We should only consider the is_annotation for access in Annotation node (`__in_annotation` already takes case that).

jimmylai (commented):
The logic in `visit_Call` also needs to be updated.
The Aug of TypeVar and NewType are not actually Annotation.

zsol (approved):

PR #411, opened 2020-11-03
[scope] Visit concatenated f-strings

### LINKED ISSUE #410: ScopeProvider misses accesses within f-strings nested in concatenated strings
Simplest repro is for a program like
```
from a import b
f'{b}' 'hello'
```

Code that demonstrates the problem:

```
m = cst.MetadataWrapper(cst.parse_module("from a import b; f'{b}' 'b'"))                        
scopes = list(set(filter(None, m.resolve(cst.metadata.ScopeProvider).values())))                
list(scopes[0].accesses)                   
```

Yields `[]`.
If the string is a standalone `f'{b}'` (without implicit string concatenation), then the access is correctly recorded.

zsol:
## Summary

This PR makes sure scope analysis visits nodes under a `ConcatenatedString`. This is important because `ConcatenatedString`s may contain `FormattedString`s which can contain further expressions. We only need to go deeper if the concatenated string is not part of a type annotation though (those can't contain f-strings).

Fixes #410.

## Test Plan

Added unit test.

### DIFF
--- a/libcst/metadata/tests/test_scope_provider.py
+++ b/libcst/metadata/tests/test_scope_provider.py
@@ -111,6 +111,25 @@ class ScopeProviderTest(UnitTest):
         wrapper = MetadataWrapper(cst.parse_module("def a():\n    from b import c\n\n"))
         wrapper.visit(DependentVisitor())
 
+    def test_fstring_accesses(self) -> None:
+        m, scopes = get_scope_metadata_provider(
+            """
+            from a import b
+            f"{b}" "hello"
+            """
+        )
+        global_scope = scopes[m]
+        self.assertIsInstance(global_scope, GlobalScope)
+        global_accesses = list(global_scope.accesses)
+        self.assertEqual(len(global_accesses), 1)
+        import_node = ensure_type(
+            ensure_type(m.body[0], cst.SimpleStatementLine).body[0], cst.ImportFrom
+        )
+        b_referent = list(global_accesses[0].referents)[0]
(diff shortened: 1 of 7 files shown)


libcst/metadata/tests/test_scope_provider.py
```diff
@@ -111,6 +111,25 @@ def fn_def():
...
+        self.assertIsInstance(global_scope, GlobalScope)
+        global_accesses = list(global_scope.accesses)
+        self.assertEqual(len(global_accesses), 1)
+        import_node = ensure_type(
+            ensure_type(m.body[0], cst.SimpleStatementLine).body[0], cst.ImportFrom
+        )
+        b_referent = list(global_accesses[0].referents)[0]
+        self.assertIsInstance(b_referent, Assignment)
+        assert isinstance(b_referent, Assignment)  # for the typechecker's eyes
+        self.assertEqual(b_referent.node, import_node)
```
jimmylai, line 131: nit: can we do this instead? We usually don't use assert because it's a noop in optimized mode.

```
if isinstance(b_referent, Assignment):
    self.assertEqual(b_referent.node, import_node)
```

jimmylai (approved):
LGTM with a minor comment.

libcst/metadata/tests/test_scope_provider.py
zsol: sure, but it doesn't matter much because there's a
```
self.assertIsInstance(b_referent, Assignment)
```
right above it, so if the assertion fails, no further code will be executed

PR #412, opened 2020-11-03
Pin linters

zsol:
## Summary
The recent Fixit releases contained new linters and behavior which caused master to be broken, causing all pull requests to be red. This PR pins `black`, `fixit`, and `isort` to a specific revision to make sure this doesn't happen again, and fixes lint violations so master can be green again

## Test Plan

`tox -e lint -r`


jimmylai (approved):
Thanks for fixing this.

PR #413, opened 2020-11-04
[scope] keep track of assignment/access ordering

### LINKED ISSUE #326: [scope analysis] References attached to the wrong scope
Consider this example:
```
from a import b
class X:
    @b
    def b(self):
        pass
```
In this case `@b` refers to the `b` imported in the global scope, but `ScopeProvider` attaches the access to the method definition instead. This means the imported `b` has 0 references (and as a consequence, gets cleaned up by `RemoveImportsVisitor`).

zsol:
## Summary
This PR makes sure references are never attached to assignments that are declared later than the access (but still in the same scope). It does so by keeping track of the relative ordering of accesses to assignments, through the inclusion of an `__index` member in both. The idea is to increment the index every time an assignment "happens" in a scope. This is a bit tricky because for `a = b`, the right hand side cannot refer to variables on the left side even if the left side is visited first.

Fixes #326 

## Test Plan

Added extensive test case for a variable shadowing in a class/global scope scenario.

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -68,6 +68,7 @@ class Access:
     is_type_hint: bool
 
     __assignments: Set["BaseAssignment"]
+    __index: int
 
     def __init__(
         self, node: cst.Name, scope: "Scope", is_annotation: bool, is_type_hint: bool
@@ -77,6 +78,7 @@ class Access:
         self.is_annotation = is_annotation
         self.is_type_hint = is_type_hint
         self.__assignments = set()
+        self.__index = scope._assignment_count
 
     def __hash__(self) -> int:
         return id(self)
@@ -86,11 +88,25 @@ class Access:
         """Return all assignments of the access."""
         return self.__assignments
 
-    def record_assignment(self, assignment: "BaseAssignment") -> None:
-        self.__assignments.add(assignment)
+    @property
+    def _index(self) -> int:
+        return self.__index
 
-    def record_assignments(self, assignments: Set["BaseAssignment"]) -> None:
-        self.__assignments |= assignments
+    def record_assignment(self, assignment: "BaseAssignment") -> None:
+        if assignment.scope != self.scope or assignment._index < self.__index:
+            self.__assignments.add(assignment)
+
+    def record_assignments(self, name: str) -> None:
+        assignments = self.scope[name]
+        # filter out assignments that happened later than this access
+        previous_assignments = {
+            assignment
+            for assignment in self.scope[name]
+            if assignment.scope != self.scope or assignment._index < self.__index
+        }
+        if previous_assignments == set() and assignments != set():
+            previous_assignments = self.scope.parent[name]
+        self.__assignments |= previous_assignments
 
 
 class BaseAssignment(abc.ABC):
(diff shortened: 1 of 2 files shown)


libcst/metadata/tests/test_scope_provider.py
```diff
@@ -1310,3 +1310,51 @@ def test_gen_dotted_names(self) -> None:
             )
         }
         self.assertEqual(names, {"a.b.c", "a.b", "a"})
+
+    def test_ordering(self) -> None:
```
jimmylai, line 1314: Can you add another test case that the Assignment is defined after the access in a different scope.
E.g.
```
def f():
    print(ab)  # this access should refer to the below assign

ab = 1
```

```
def f(ab):
    print(ab)  # this access should refer to the above assign

ab = 1
```

zsol: Yep, will do

libcst/metadata/scope_provider.py
```diff
@@ -915,6 +979,33 @@ def _visit_comp_alike(
...
+
+    def leave_NamedExpr(self, original_node: cst.NamedExpr) -> None:
+        self._leave_assignment_alike()
+
+    def leave_Parameters(self, original_node: cst.Parameters) -> None:
+        self._leave_assignment_alike()
+
+    def leave_WithItem(self, original_node: cst.WithItem) -> None:
+        self._leave_assignment_alike()
+
```
jimmylai, line 1008: The current way to increment assignment count require us to manual identify each leave_Node and call `self._leave_assignment_alike()` which make the code less easy to maintain and easier to miss some nodes.
Another simpler way to increment the assignment count is to simply provide a `get_and_increment_assignment_count()` method which automatically increment the index when it's called. In this way the index will be incremented whenever we create an assignment and access the assignment_count which is a simpler implementation and easier to make it right.

libcst/metadata/scope_provider.py
```diff
@@ -109,10 +125,20 @@ def __init__(self, name: str, scope: "Scope") -> None:
...
+        later_accesses = {
+            access
+            for access in accesses
+            if access.scope != self.scope or self._index < access._index
+        }
+        self.__accesses |= later_accesses
+        earlier_accesses = accesses - later_accesses
+        if earlier_accesses != set() and self.scope.parent != self.scope:
+            for shadowed_assignment in self.scope.parent[self.name]:
+                shadowed_assignment.record_accesses(earlier_accesses)
```
jimmylai, line 141: Why do we need this?

jimmylai: NVM, I got it. It's to handle the accesses not part of later access. Add them to parent assignments make sense.

jimmylai (approved):
Other than the comments, LGTM.

libcst/metadata/scope_provider.py
zsol: This is what I started out with, but unfortunately it's not correct: `Assignment`s are created as we visit nodes, so the ordering is dependent on visitation order. For a statement like
```
b = b
```
We want the right hand side to not be able to refer to the left hand side. This means we need to increase the counter _after_ processing this statement, not just after visiting the left hand side.

libcst/metadata/scope_provider.py
zsol: Exactly!

zsol: I used to have a comment here, it might've got lost during rebases. I'll add it back

libcst/metadata/scope_provider.py
jimmylai: Got it. I'm a bit worry about the increasing complexity of ScopeProvider class implementation.
How about move all the assignment counter state to another Visitor class and let ScopeProvider class inherit from it? That can isolate the index counter logic and keep it simple.

libcst/metadata/scope_provider.py
```diff
@@ -86,11 +88,25 @@ def referents(self) -> Collection["BaseAssignment"]:
...
+
+    def record_assignments(self, name: str) -> None:
+        assignments = self.scope[name]
+        # filter out assignments that happened later than this access
+        previous_assignments = {
+            assignment
+            for assignment in self.scope[name]
+            if assignment.scope != self.scope or assignment._index < self.__index
+        }
+        if previous_assignments == set() and assignments != set():
```
thatch, line 107: why not truthiness checks?

PR #417, opened 2020-11-11
Correct handling of walrus operator in function args

### LINKED ISSUE #416: Function call with walrus loses colon in roundtrip
This input

```
f(x:=1)
```

is roundtripped as

```
f(x=1)
```

due to a parse error.

I'm fairly certain the problem is the grammar on https://github.com/Instagram/LibCST/blob/master/libcst/_parser/conversions/expression.py#L1430 not matching namedexpr_test nor doing anything but AssignEqual in the else case.  I tried to alter the grammar there to use namedexpr_test, but that caused ambiguity.

Going the other route to fix, we'll need another case to fix because this is the first positional arg, not a kwarg.

thatch:
## Summary

Previous behavior treated it as identical to equal, making a kwarg; it should
instead be a positional arg.  Includes several tests to make sure that
whitespace handling is correct.

Fixes #416

## Test Plan

Includes new tests that were failing before.  I had some trouble ensuring the whitespace ended up in the right spot during development, so this includes a few probably unnecessary tests out of a general distrust that I did it right.

### DIFF
--- a/libcst/_parser/conversions/expression.py
+++ b/libcst/_parser/conversions/expression.py
@@ -1441,8 +1441,16 @@ def convert_arg_assign_comp_for(
         elt, for_in = children
         return Arg(value=GeneratorExp(elt.value, for_in, lpar=(), rpar=()))
     else:
-        # "key = value" assignment argument
         lhs, equal, rhs = children
+        # "key := value" assignment; positional
+        if equal.string == ":=":
+            val = convert_namedexpr_test(config, children)
+            if not isinstance(val, WithLeadingWhitespace):
+                raise Exception(
+                    f"convert_namedexpr_test returned {val!r}, not WithLeadingWhitespace"
+                )
+            return Arg(value=val.value)
+        # "key = value" assignment; keyword argument
         return Arg(
             keyword=lhs.value,
             equal=AssignEqual(
--- a/libcst/_nodes/tests/test_namedexpr.py
+++ b/libcst/_nodes/tests/test_namedexpr.py
@@ -101,6 +101,71 @@ class NamedExprTest(CSTNodeTest):
                 "parser": _parse_statement_force_38,
                 "expected_position": None,
             },
+            # Function args
+            {
+                "node": cst.Call(
+                    func=cst.Name(value="f"),
+                    args=[
+                        cst.Arg(
+                            value=cst.NamedExpr(
+                                target=cst.Name(value="y"),
+                                value=cst.Integer(value="1"),
+                                whitespace_before_walrus=cst.SimpleWhitespace(""),
+                                whitespace_after_walrus=cst.SimpleWhitespace(""),
+                            )
+                        ),
+                    ],
+                ),
+                "code": "f(y:=1)",
+                "parser": _parse_expression_force_38,
+                "expected_position": None,
+            },
+            # Whitespace handling on args is fragile
+            {
+                "node": cst.Call(
+                    func=cst.Name(value="f"),
+                    args=[
+                        cst.Arg(
+                            value=cst.Name(value="x"),
+                            comma=cst.Comma(
+                                whitespace_after=cst.SimpleWhitespace("  ")
+                            ),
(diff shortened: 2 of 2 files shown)


jreese:
Looks good to me, and I verified locally that it resolves the issue from usort.

Before fix:
```
(.venv) jreese@garbodor ~/workspace/usort main± » cat foo.py
foo.func(value := "some string literal")

(.venv) jreese@garbodor ~/workspace/usort main± » usort diff foo.py
--- a/foo.py
+++ b/foo.py
@@ -1,2 +1,2 @@
-foo.func(value := "some string literal")
+foo.func(value = "some string literal")
```

With fix:
```
(.venv) jreese@garbodor ~/workspace/usort main± » pip install -U git+https://github.com/thatch/libcst@issue416
Collecting git+https://github.com/thatch/libcst@issue416
  Cloning https://github.com/thatch/libcst (to revision issue416) to /private/var/folders/_b/6gpzshq548d4d_4bk1m5sq280000gn/T/pip-req-build-n3b3irzn
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Requirement already satisfied, skipping upgrade: typing-inspect>=0.4.0 in ./.venv/lib/python3.8/site-packages (from libcst==0.3.13) (0.6.0)
Requirement already satisfied, skipping upgrade: typing-extensions>=3.7.4.2 in ./.venv/lib/python3.8/site-packages (from libcst==0.3.13) (3.7.4.3)
Requirement already satisfied, skipping upgrade: pyyaml>=5.2 in ./.venv/lib/python3.8/site-packages (from libcst==0.3.13) (5.3.1)
Requirement already satisfied, skipping upgrade: mypy-extensions>=0.3.0 in ./.venv/lib/python3.8/site-packages (from typing-inspect>=0.4.0->libcst==0.3.13) (0.4.3)
Building wheels for collected packages: libcst
  Building wheel for libcst (PEP 517) ... done
  Created wheel for libcst: filename=libcst-0.3.13-py3-none-any.whl size=502635 sha256=9dd3fcfe16688992f2ff6bb0ddd8235f15fb595134768bbdb9aeb5660fcee259
  Stored in directory: /private/var/folders/_b/6gpzshq548d4d_4bk1m5sq280000gn/T/pip-ephem-wheel-cache-i73lgx32/wheels/cc/af/27/cece6ddd4a6a62f85a0ea3964b00c4fd3d176216786b578d03
Successfully built libcst
Installing collected packages: libcst
  Attempting uninstall: libcst
    Found existing installation: libcst 0.3.13
    Uninstalling libcst-0.3.13:
      Successfully uninstalled libcst-0.3.13
Successfully installed libcst-0.3.13
WARNING: You are using pip version 20.2.1; however, version 20.2.4 is available.
You should consider upgrading via the '/Users/jreese/workspace/usort/.venv/bin/python -m pip install --upgrade pip' command.

(.venv) jreese@garbodor ~/workspace/usort main± » usort diff foo.py
```

jreese (approved):

libcst/_parser/conversions/expression.py
```diff
@@ -1441,8 +1441,13 @@ def convert_arg_assign_comp_for(
         elt, for_in = children
         return Arg(value=GeneratorExp(elt.value, for_in, lpar=(), rpar=()))
     else:
-        # "key = value" assignment argument
         lhs, equal, rhs = children
+        # "key := value" assignment; positional
+        if equal.string == ":=":
+            val = convert_namedexpr_test(config, children)
+            assert isinstance(val, WithLeadingWhitespace)
```
jimmylai, line 1448: We don't use assert since it's noop in optimized mode. Let's just use if condition and raise when it returns unexpected result.
```suggestion
            if isinstance(val, WithLeadingWhitespace):
                return Arg(value=val.value)
            else:
                raise Exception("...")
```

jimmylai (approved):
Other than the comment, LGTM.

PR #418, opened 2020-11-12
Handle string type references in cast()

Kronuz:
## Summary
Handle `cast("Type", object)` so `Type` is properly referenced in the scope.

## Test Plan
Run tests

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -719,8 +719,8 @@ class ScopeVisitor(cst.CSTVisitor):
         if "typing.cast" in qnames:
             node.func.visit(self)
             self.__in_type_hint.add(node)
-            for arg in node.args[:1]:
-                arg.visit(self)
+            if len(node.args) > 0:
+                node.args[0].visit(self)
             return False
         return True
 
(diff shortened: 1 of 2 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -709,13 +709,19 @@ def visit_Attribute(self, node: cst.Attribute) -> Optional[bool]:
...
             self.__in_type_hint.add(node)
             for arg in node.args[1:]:
                 arg.visit(self)
             return False
+        if "typing.cast" in qnames:
+            node.func.visit(self)
+            self.__in_type_hint.add(node)
+            for arg in node.args[:1]:
+                arg.visit(self)
+            return False
```
zsol, line 724: I had to read this 3 times before realizing what's the difference between the two `if` branches.
```suggestion
        if "typing.NewType" in qnames or "typing.TypeVar" in qnames:
            node.func.visit(self)
            self.__in_type_hint.add(node)
            for arg in node.args[1:]:
                arg.visit(self)
            return False
        if "typing.cast" in qnames:
            node.func.visit(self)
            self.__in_type_hint.add(node)
            if len(node.args) > 0:
                node.args[0].visit(self)
            return False
```

zsol (approved):

PR #422, opened 2020-11-20
Fix referencing of remaining objects in cast()

Kronuz:
## Summary

Casted objects weren't being referenced. i.e., in:

```
from typing import cast
from module import SomeType, SomeObject

casted = cast("SomeType", SomeObject)
```

`SomeObject` ended up not being referenced.

## Test Plan

Run tests


jimmylai (approved):
LGTM, thanks for fixing the issue!

PR #423, opened 2020-11-20
fix assignment/access ordering in comprehensions

zsol:
## Summary
This change fixes attaching accesses to the right assignment in list comprehensions, so in
```
a = 1
[a for a in [1,2,3]]
```
The `a` at the beginning of the list comprehension is now attached to the `a` declared after `for` instead of the declaration on the first line.
## Test Plan
Added unit tests

### DIFF
--- a/libcst/metadata/tests/test_scope_provider.py
+++ b/libcst/metadata/tests/test_scope_provider.py
@@ -1531,3 +1531,28 @@ class ScopeProviderTest(UnitTest):
         a_global_assignment = list(scopes[a_global]["a"])[0]
         a_global_refs = list(a_global_assignment.references)
         self.assertEqual(a_global_refs, [])
+
+    def test_ordering_comprehension_confusing(self) -> None:
+        m, scopes = get_scope_metadata_provider(
+            """
+            def f(a):
+                [a for a in a]
+            a = 1
+            """
+        )
+        f = cst.ensure_type(m.body[0], cst.FunctionDef)
+        a_param = f.params.params[0].name
+        a_param_assignment = list(scopes[a_param]["a"])[0]
+        a_param_refs = list(a_param_assignment.references)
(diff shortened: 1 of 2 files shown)


libcst/metadata/tests/test_scope_provider.py
```diff
@@ -1454,3 +1454,80 @@ def f(a):
...
+                [b for a in [] for b in a]
+                [a for a in [] for a in []]
+            a = 1
+            """
+        )
+        f = cst.ensure_type(m.body[0], cst.FunctionDef)
+        a_param = f.params.params[0].name
+        a_param_assignment = list(scopes[a_param]["a"])[0]
+        a_param_refs = list(a_param_assignment.references)
+        self.assertEqual(a_param_refs, [])
```
jimmylai, line 1472: Doesn't the param `a` have references in the comprehensions (the last `a` appear in the comprehension)?
[a for a in [] for b in `a`]
[b for a in [] for b in `a`]

zsol: no, that last `a` refers to the binding in the first `for`, so it's equivalent to:
```
def f(a_1):
    [a_2 for a_2 in [] for b_1 in a_2]
    [b_2 for a_3 in [] for b_2 in a_3]
    [a_5 for a_4 in [] for a_5 in []]
```

jimmylai: Got it. I didn't aware the nested for loop comprehension.
`[a for a in a]` is an example that I'm curious about whether it'll work fine.
The 3rd `a` should be an access of the function parameter `a`.
The first `a` is an Access of the Assignment of the 2nd `a`.
Can you also add this as a test case?

jimmylai (approved):
Other than the comment, LGTM.

PR #426, opened 2020-12-03
Gathering exports in augmented assignment statements

Kronuz:
## Summary
Add support for gathering exports in `__all__` augmented assignments, such as:

```
import os
__all__ = []
__all__ += ["os"]
```

## Test Plan
Run existing tests

### DIFF
--- a/libcst/codemod/visitors/_gather_exports.py
+++ b/libcst/codemod/visitors/_gather_exports.py
@@ -6,6 +6,7 @@
 from typing import Set, Union
 
 import libcst as cst
+import libcst.matchers as m
 from libcst.codemod._context import CodemodContext
 from libcst.codemod._visitor import ContextAwareVisitor
 from libcst.helpers import get_full_name_for_node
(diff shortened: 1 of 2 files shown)


libcst/codemod/visitors/_gather_exports.py
```diff
@@ -53,6 +53,14 @@ def visit_AnnAssign(self, node: cst.AnnAssign) -> bool:
                 return True
         return False
 
+    def visit_AugAssign(self, node: cst.AugAssign) -> bool:
+        target_name = get_full_name_for_node(node.target)
+        if target_name == "__all__":
+            if isinstance(node.value, (cst.List, cst.Tuple, cst.Set)):
+                self._is_assigned_export.add(node.value)
+                return True
+        return False
```
jimmylai, line 62: We should also verify the operator is `+=` to be specific.
There is a type error that can be fixed by the suggested change.
```suggestion
    def visit_AugAssign(self, node: cst.AugAssign) -> bool:
        target_name = get_full_name_for_node(node.target)
        if target_name == "__all__":
            value = node.value
            if isinstance(value, (cst.List, cst.Tuple, cst.Set)):
                self._is_assigned_export.add(value)
                return True
        return False
```

jimmylai (changes requested):

libcst/codemod/visitors/_gather_exports.py
```diff
@@ -53,6 +53,17 @@ def visit_AnnAssign(self, node: cst.AnnAssign) -> bool:
                 return True
         return False
 
+    def visit_AugAssign(self, node: cst.AugAssign) -> bool:
+        target_name = get_full_name_for_node(node.target)
+        if target_name == "__all__":
+            value = node.value
+            if isinstance(value, (cst.List, cst.Tuple)) and isinstance(
+                node.operator, cst.AddAssign
+            ):
```
jimmylai, line 62: Those two if statements can be replaced as a matcher.
```suggestion
            if m.matches(node, m.AugAssign(target=m.Name("__all__"), operator=m.AddAssign, value=m.List() | m.Tuple())):
```

jimmylai (approved):
LGTM, use a matcher can keep the code simpler.

PR #428, opened 2020-12-08
Fix for not parsing subscripts such as cast()["from"]

Kronuz:
## Summary

When handling subscripts, `cast()`, and potentially others such as `TypeVar()` and `NewType()`, were being treated as potential string type hints, resulting in snippets like the following failing with a `ParserSyntaxError`:

```py
cast(Dict[str, str], {})["from"]
```

## Test Plan

Run tests


jimmylai (approved):
LGTM.
@Kronuz thanks for fixing this quickly.

PR #430, opened 2020-12-14
Handle scope ordering in For statements

zsol:
## Summary
When visiting `for ... in ... :` statements, the scope provider was incrementing the assignment counter _after_ the entire `For` statement was visited. This meant that all accesses to `For.target` in `For.body` were potentially attached to the wrong scope. The only exception to this was when the scope of the `For` statement was in the global scope, in which case the fallback logic accidentally saved us (because for the global scope, `scope.parent == scope`).

This PR fixes the bug by customizing the way `ScopeProvider` visits `For` statements, and increments the assignment counter directly after visiting `For.target`.

## Test Plan

Added a test case.

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -45,7 +45,6 @@ _ASSIGNMENT_LIKE_NODES = (
     cst.AugAssign,
     cst.ClassDef,
     cst.CompFor,
-    cst.For,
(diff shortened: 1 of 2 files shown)


pcottle:
Excited to see this land 🙌🏻

jimmylai (approved):
LGTM

PR #431, opened 2020-12-14
Don't allow out of order accesses in the global scope

zsol:
## Summary

While investigating #430 I realized that scope provider might attach accesses to assignments AFTER the access in the global scope. This PR fixes that bug.

## Test Plan

Added test case and verified it fails before this PR.


jimmylai (approved):
LGTM

PR #433, opened 2020-12-15
Walrus operator's left hand side now has STORE expression context

zsol:
## Summary

This PR makes scope provider assign a `STORE` expression context to the left hand side of a walrus operator. So, in an `x := y` expression, `x` should have `ExpressionContext.STORE` as its expression context metadata.

## Test Plan

Added unit tests.

### DIFF
--- a/libcst/metadata/tests/test_expression_context_provider.py
+++ b/libcst/metadata/tests/test_expression_context_provider.py
@@ -4,7 +4,6 @@
 # LICENSE file in the root directory of this source tree.
 
 
-import sys
 from textwrap import dedent
 from typing import Dict, Optional, cast
 
--- a/libcst/metadata/tests/test_expression_context_provider.py
+++ b/libcst/metadata/tests/test_expression_context_provider.py
@@ -411,3 +411,19 @@ class ExpressionContextProviderTest(UnitTest):
                 },
             )
         )
+
+    def test_walrus(self) -> None:
+        code = """
+        if x := y:
+            pass
+        """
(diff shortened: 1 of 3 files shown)


libcst/metadata/tests/test_expression_context_provider.py
```diff
@@ -411,3 +412,21 @@ def test_function(self) -> None:
...
         )
+
+    def test_walrus(self) -> None:
+        if sys.version_info < (3, 8):
+            self.skipTest("This python version does not support :=")
+        code = """
+        if x := y:
+            pass
+        """
+        wrapper = MetadataWrapper(parse_module(dedent(code)))
```
thatch, line 423: The version can be decoupled; if you want, this will let the test run everywhere:

```
parse_module(
  dedent(code),
  config=cst.PartialParserConfig(python_version="3.8"),
)
```

libcst/metadata/tests/test_scope_provider.py
```diff
@@ -1609,6 +1610,27 @@ def test_no_out_of_order_references_in_global_scope(self) -> None:
                     ),
                 )
 
+    def test_walrus_accesses(self) -> None:
+        if sys.version_info < (3, 8):
```
thatch, line 1614: Same thing here?

zsol: This one is not as easy, I'd have to plumb the parser config through some helper functions, so I thought it's fine to keep it as is

thatch (approved):
Change is pretty good.

Issue #435, opened 2020-12-21
`QualifiedNameProvider` matcher unusable with Windows process pools


Zac-HD:
I have a MVP for https://github.com/HypothesisWorks/hypothesis/issues/2705, but discovered that on Windows it can only format a single file at a time.  The underlying problem is that `multiprocessing` on Windows works via `spawn` (vs `fork` on Linux), and lambdas cannot be pickled but are practically required to use matchers.  Related to #294 in that running tests on Windows would presumably detect this too?

```python
import libcst as cst
import libcst.matchers as m
from libcst.codemod import VisitorBasedCodemodCommand

def match_qualname(name):
    return m.MatchMetadataIfTrue(
        cst.metadata.QualifiedNameProvider,
        lambda qualnames: all(n.name == name for n in qualnames),  # <-- here's the problem
    )

class HypothesisFixComplexMinMagnitude(VisitorBasedCodemodCommand):
    DESCRIPTION: str = "Upgrade from deprecated or inefficient uses of Hypothesis."
    METADATA_DEPENDENCIES = (cst.metadata.QualifiedNameProvider,)

    @m.call_if_inside(
        m.Call(metadata=match_qualname("hypothesis.strategies.complex_numbers"),
               args=[m.Arg(value=m.Name("None"), keyword=m.Name("min_magnitude"))])
    )
    def leave_Arg(self, original_node: cst.Arg, updated_node: cst.Arg) -> cst.Arg:
        if m.matches(updated_node.keyword, m.Name("min_magnitude")):
            return updated_node.with_changes(value=cst.Integer("0"))
        return updated_node

# From https://github.com/Zac-HD/hypothesis/commit/5002c516332f5692be18d3f551157e694400c517
# aka  https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:codemods
```
```python
# Save two files with this code, try to codemod them, and you'll get the error.
from hypothesis.strategies import complex_numbers

complex_numbers(min_magnitude=None)
```

It's impractical to write a top-level function for each of the fifty-ish qualnames we want to handle, and unfortunately `functools.partial` has the same problem as lambdas.

The obvious workaround is to give up some performance and run in a single process on Windows.  Unfortunately *that doesn't work either* because `--jobs=1` still opens a process pool...

thatch:
I don't quite follow what the "all" is doing, but a callable object may be picklable enough to find the _next_ windows bug...

```py
class QualnameMatcher:
[...]

PR #436, opened 2020-12-21
Work around Windows problem by using dummy pool for `jobs=1`

Zac-HD:
Per #435, any codemod which is implemented with a `lambda` cannot be run on more than one file at time.  That's because `pickle` limitations mean we can't use a process pool, and so this patch ensures that we *don't* use a process pool for `jobs=1` and thus enables a "just run serially" workaround.

This is also something of a performance optimisation, since starting a single process to do all the work is likely to take longer than just doing it in the current process.


zsol (approved):
Seems reasonable. Thx for the fix!

PR #440, opened 2021-01-02
refactor: optimize number of parallel processes

sk-:
## Summary
Instead of always launching the specified number of jobs or cpu_count, we should take into account how many files need to be processed and how many files are delivered at once to each process.

Note that this is similar to what fixit does, although they don't specify a chunk size.


jimmylai (approved):
LGTM.
This fix will reduce the number of unused process if the input are small number of files. It probably won't show significant impact on performance.

PR #445, opened 2021-01-07
ApplyTypeAnnotations: apply parameter annotations even if no return type is declared

MaggieMoss:
## Summary

Caught this bug while running codemods, the result was that stubs for functions with parameter annotations, but without return annotations would not be annotated.

## Test Plan

`python -m unittest -v libcst.codemod.visitors.tests.test_apply_type_annotations`


zsol (approved):
Looks straightforward, thanks!

PR #453, opened 2021-01-26
Extra unused imports added in other files issue

### LINKED ISSUE #447: Extra unused imports added in other files
I'm the author of [django-codemod](https://github.com/browniebroke/django-codemod), which uses libCST under the hood. I'm not using libCST in the usual way, so maybe there is something wrong in how I do things, but I recently noticed an issue with libCST 0.3.14 and above (https://github.com/browniebroke/django-codemod/issues/302).

Basically, when some of my context aware transformers are executed, they seem to 'leak' imports into other files. One thing that is special is that the affected transformers are making use of the built-in `AddImportsVisitor` to add new imports:

https://github.com/browniebroke/django-codemod/blob/1b785b7f50f9d96a17d56ae7c8b3d76b6f218390/django_codemod/visitors/base.py#L128-L133

I wrote [a small test case](https://github.com/browniebroke/django-codemod/blob/main/tests/test_e2e.sh) for django-codemod, it passes with libCST v0.3.13 but [breaks with v0.3.14 and above](https://github.com/browniebroke/django-codemod/pull/304). In the test case, we expect to have only `models.py` file modified, but in the failure, we can see that `views.py` and `__init__.py` are also modified, they have some unused imports being added. 

I wanted to reach out in case someone has an idea on what might cause this bug. Failing that, I'll try to build a more minimal test case without my library in the equation.

browniebroke:
## Summary

Fixes #447 

As I suspected, `git bisect` points at #402. 

## Test Plan

- [x] Write a broad CLI test reproducing the problem
- [x] Track down the bug and fix it
- [ ] Rewrite the test as a non-regression test for the root cause?

### DIFF
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -279,6 +279,7 @@ def _execute_transform(  # noqa: C901
             transformer.context,
             filename=filename,
             full_module_name=_calculate_module(config.repo_root, filename),
+            scratch={},
         )
 
         # Run the transform, bail if we failed or if we aren't formatting code
--- a/libcst/tests/test_e2e.py
+++ b/libcst/tests/test_e2e.py
@@ -1,0 +1,76 @@
+import contextlib
+import os
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from unittest import TestCase
+
+from libcst import Name, matchers as m
+from libcst.codemod import (
+    CodemodContext,
+    VisitorBasedCodemodCommand,
+    gather_files,
+    parallel_exec_transform_with_prettyprint,
+)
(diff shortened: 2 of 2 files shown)


browniebroke:
Not sure how to proceed with the test actually. The code with the bug is qualified as a "somewhat gross hack" in a nearby comment, and is in a middle of a pretty long function for which I couldn't locate any existing tests specific to it.

Let me know what I should do with the test I wrote.

zsol:
I think the test is fine like it is. Thank you for spotting and fixing the bug!

zsol (approved):

Vayu:
@browniebroke that broke MagicArgsCodemodCommand passing cmd-line arguments to child transforms

PR #454, opened 2021-02-04
Update README.rst

cdonovick:
Just spent a couple hours trying to figure out why I couldn't get pyre to work.  Only to discover that having my venv inside the libcst dir was breaking pyre.


jimmylai (approved):
Thanks for updating the README and that makes it more clear about the hidden assumption: the venv path cannot be in working directory to avoid Pyre checks all source code installed in the venv.

PR #455, opened 2021-02-05
Add FlattenSentinel

### LINKED ISSUE #263: Inserting statements into blocks

cdonovick:
## Summary
Adds a `FlattenSentinel` as proposed by @sk- in #269 to resolve #263.

The main idea is to allow transformers to return multiple statements to replace a single statement.

## Test Plan
Wrote basic tests to ensure both `BaseSmallStatement` and `BaseStatement`  could be replaced with a `FlattenSentinel`. 

## Notes
Does not currently type check but I want to make sure this design is acceptable before I try to push `FlattenSentinel` basically everywhere.


cdonovick:
Another possible option which might make typing a bit more sane would be to add types:
```Python3
class SmallStatementSeq(BaseSmallStatement):
    nodes: Sequence[BaseSmallStatement]

class StatementSeq(BaseStatement):
    nodes: Sequence[BaseStatement]   
```

These types would simply exist to allow transformers to return multiple nodes instead of a single node.

cdonovick:
I think the `*Seq` types approach is in some ways not as nice as they would continue to live in the AST until they were serialized, but the two approach could be potentially combined with something like:

```Python3
class _FlattenSentinel:
    nodes: Sequence[CSTNode]

class SmallStatementSeq(BaseSmallStatement, _FlattenSentinel):
    nodes: Sequence[BaseSmallStatement]

class StatementSeq(BaseStatement, _FlattenSentinel):
    nodes: Sequence[BaseStatement]
```
And have the transformer machinery flatten eagerly.

jimmylai:
FlattenSentinel can solve the case of replacing a node (e.g. a statement) with multiple nodes. We still need other helpers to make other use case easier, e.g. moving/adding statements to an another location, and that can be another PR.

FlattenSentinel idea looks good to me. @cdonovick please go ahead to complete this PR by fixing the type error and add detailed document in
https://github.com/Instagram/LibCST/blob/6731aa5d2940ce82dd221963b64a496955bfd07a/docs/source/visitors.rst

cdonovick:
Probably won't get the docs done today but I will try to get them done tomorrow.

cdonovick:
@jimmylai should be all set when you have a chance to review.

cdonovick:
@jimmylai Any chance this can be merged?

jimmylai (approved):
Sorry for the late review. The change looks great!

PR #458, opened 2021-02-17
Issue 341

### LINKED ISSUE #341: [BUG] Generated Code different than AST
When doing replacements in a node, it's easy to end up with a tree, which won't get correctly converted to source code. The issue is that `libcst` is not adding the necessary parentheses.

For example, the code below
```py
import libcst.matchers as m
import libcst

module = libcst.parse_module('x * y')

matcher = m.Name(value='x')
new_module = m.replace(module, matcher, libcst.parse_expression("1 + 2"))
print(new_module.code)
```

will output 
```py
1 + 2 * y
```

whereas the expected code is:

```py
(1 + 2) * y
```

I think it's reasonable to expect that a `module` is equivalent to `parse_module(module.code)`.

cdonovick:
## Summary
resolves #341 

Adds `_PrecedenceNode` as a helper class to automatically parenthesize child nodes when necessary

## Test Plan
Adds basic unit tests for implicit parens and extends fuzz testing.  

## Notes
I had originally tried to allow  `MaybeSentinel` in `{l,r}par` attrs but this proved difficulty to get right.  It would require numerous changes to the parser to ensure proper construction of `_BaseParenthesizableNode` as they are often explicitly constructed with empty `lpar` and `rpar` attributes.

### DIFF
--- a/libcst/_nodes/expression.py
+++ b/libcst/_nodes/expression.py
@@ -9,7 +9,7 @@ from abc import ABC, abstractmethod
 from ast import literal_eval
 from contextlib import contextmanager
 from dataclasses import dataclass, field
-from enum import Enum, auto, IntEnum
+from enum import Enum, IntEnum, auto
 from tokenize import (
     Floatnumber as FLOATNUMBER_RE,
     Imagnumber as IMAGNUMBER_RE,
@@ -259,11 +259,9 @@ class _BaseParenthesizedNode(CSTNode, ABC):
         if len(self.lpar) != len(self.rpar):
             raise CSTValidationError("Cannot have unbalanced parens.")
 
-
     @property
     def _is_parenthesized(self) -> bool:
         return len(self.lpar) > 0 and len(self.rpar) > 0
-
 
     @contextmanager
     def _parenthesize(self, state: CodegenState) -> Generator[None, None, None]:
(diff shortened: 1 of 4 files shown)


notEvil:
Isn't this pretty important? Afaik any transformer might produce logically different code. Is there a simple "workaround" I don't see?

*thank you for libcst!*

cdonovick:
@notEvil I have a quick a dirty work around in the linked issue. https://github.com/Instagram/LibCST/issues/341#issuecomment-716900549

You could probably do something more clever by incorporating the logic I use here in the transformer.

notEvil:
> @notEvil I have a quick a dirty work around in the linked issue. [#341 (comment)](https://github.com/Instagram/LibCST/issues/341#issuecomment-716900549)

Thanks, I've seen it but don't want to introduce parentheses.

> You could probably do something more clever by incorporating the logic I use here in the transformer.

I just did, except for Power which probably takes me another hour to figure out (because transformer act at child level). Still think this PR should gain more attention.

kiri11:
Looks amazing! Prevents tricky issues.

zsol:
I'm happy to review this if someone's willing to resurrect and rebase

notEvil:
I've just published my recent work on parentheses and precedence at https://gitlab.com/notEvil/normcst. The tests generate ~7.5k combinations of expressions/statements/match patterns and ensure that `normcst.ParenthesisTransformer` produces valid and minimal results. I will use it in my own transformations for a while before pushing it to PyPI and would be happy for anyone who helps by giving it a try.

Issue #460, opened 2021-02-26
libcst.metadata.Scope.get_qualified_names_for doesn't handle relative imports


lpetre:
This also applies to QualifiedNameProvider. As far as I can tell, these types don't have the necessary context (ie full module name) to resolve relative imports.

libcst.helpers.get_absolute_module_for_import could resolve the proper name if we had access to a full module name for the scope.

Working:
```
>>> wrapper = cst.MetadataWrapper(cst.parse_module("from lib import a\na()"))
>>> scopes = wrapper.resolve(cst.metadata.ScopeProvider)
>>> scope_of_module = scopes[wrapper.module]
>>> a_call = wrapper.module.body[1].body[0].value
>>> scope_of_module.get_qualified_names_for(a_call)
{QualifiedName(name='lib.a', source=<QualifiedNameSource.IMPORT: 1>)}

```
Broken:
```
>>> wrapper = cst.MetadataWrapper(cst.parse_module("from . import a\na()"))
>>> scopes = wrapper.resolve(cst.metadata.ScopeProvider)
>>> scope_of_module = scopes[wrapper.module]
>>> a_call = wrapper.module.body[1].body[0].value
>>> scope_of_module.get_qualified_names_for(a_call)
{QualifiedName(name='a', source=<QualifiedNameSource.IMPORT: 1>)}
```

I believe this TODO is necessary to implement (somehow): https://github.com/Instagram/LibCST/blob/master/libcst/metadata/scope_provider.py#L298

PR #461, opened 2021-02-26
Update Scope.get_qualified_names_for to resolve relative imports

lpetre:
## Summary
This is an attempt to fix issue #460.

Unfortunately I'm not sure how to expose this fix to `libcst.metadata.QualifiedNameProvider`, which would be very useful in my opinion.  As far as I can tell, metadata providers can not be parameterized, nor do they have access to the `libcst.MetadataWrapper`. It might be an idea to add an optional `full_module_name`  as a field on the `libcst.MetadataWrapper` so that we can do this relative import resolution automatically.

## Test Plan

```
>>> import libcst as cst
>>> wrapper = cst.MetadataWrapper(cst.parse_module("from . import a\na()"))
>>> scopes = wrapper.resolve(cst.metadata.ScopeProvider)
>>> scope_of_module = scopes[wrapper.module]
>>> a_call = wrapper.module.body[1].body[0].value
>>> scope_of_module.get_qualified_names_for(a_call, "x.y.z")
{QualifiedName(name='x.y.a', source=<QualifiedNameSource.IMPORT: 1>)}
```

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -28,7 +28,7 @@ import libcst as cst
 from libcst import ensure_type
 from libcst._add_slots import add_slots
 from libcst._metadata_dependent import MetadataDependent
-from libcst.helpers import get_full_name_for_node
+from libcst.helpers import get_full_name_for_node, get_absolute_module_for_import
 from libcst.metadata.base_provider import BatchableMetadataProvider
 from libcst.metadata.expression_context_provider import (
     ExpressionContext,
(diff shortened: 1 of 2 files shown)


jimmylai:
I like the idea of supporting relative import but not the implementation.
Add the new parameter to get_qualified_names_for makes it fragile, it delegate the responsibility of preparing full module name to users and it returns different results when it's called with different input. (When operating on a syntax tree, we expect all qualified name should be related to a repository or a path.)
The path of current module can be used to compute the needed prefix for relative imports and we can just compute it in a metadata provider.
There is a gen_cache API to allow a MetadataProvider initialized with required data. We could implement a subclass of ScopeProvider/QualifiedNameProvider that has a gen_cache to support it.
https://github.com/Instagram/LibCST/blob/6731aa5d2940ce82dd221963b64a496955bfd07a/libcst/metadata/base_provider.py#L61
TypeInferenceProvider is an example provider that has gen_cache implemented.
https://libcst.readthedocs.io/en/latest/metadata.html#type-inference-metadata

lpetre:
Closing per comments and better solution in #465

PR #464, opened 2021-03-11
Split QualifiedNameProvider out from libcst.metadata.scope_provider

zsol:
## Summary
The provider is still available from `libcst.metadata`. I'm planning to add `FullyQualifiedNameProvider` to this file too.
## Test Plan
New and existing tests.


libcst/metadata/tests/test_name_provider.py
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
```
jimmylai, line 2: There is a test_qualified_name_provider.py which is duplicated with this  test_name_provider.py
We should consolidate and provide a better file naming.
https://github.com/Instagram/LibCST/blob/6731aa5d2940ce82dd221963b64a496955bfd07a/libcst/metadata/tests/test_qualified_name_provider.py

jimmylai (approved):

Issue #462, opened 2021-03-10
Incorrectly parsing escape char in raw string (eg: `\{`)


rodrigozhou:
LibCST v0.3.17 is raising an exception when parsing this valid code:

```
a = "'"
print(fr"\{a}")
```

Stack trace:
```
Traceback (most recent call last):
  File ".../libcst/_parser/base_parser.py", line 152, in _add_token
    plan = stack[-1].dfa.transitions[transition]
KeyError: ReservedString(})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".../libcst/tool.py", line 832, in <module>
    main(os.environ.get("LIBCST_TOOL_COMMAND_NAME", "libcst.tool"), sys.argv[1:])
  File ".../libcst/tool.py", line 827, in main
    return lookup.get(args.action or None, _invalid_command)(proc_name, command_args)
  File ".../libcst/tool.py", line 272, in _print_tree_impl
    tree = parse_module(
  File ".../libcst/_parser/entrypoints.py", line 71, in parse_module
    result = _parse(
  File ".../libcst/_parser/entrypoints.py", line 51, in _parse
    result = parser.parse()
  File ".../libcst/_parser/base_parser.py", line 111, in parse
    self._add_token(token)
  File ".../libcst/_parser/base_parser.py", line 187, in _add_token
    raise ParserSyntaxError(
libcst._exceptions.ParserSyntaxError: Syntax Error @ 29:17.
Incomplete input. Encountered '}', but expected 'FSTRING_END', 'FSTRING_STRING', or '{'.
```

My guess is LibCST is escaping `"\{"`, but it wasn't supposed to escape since it's a raw string.

thatch:
Bisect says I introduced this in 02fc4401bc3313f0db109d86d1b84c6b35abba5a which probably means it's broken in parso as well.  See https://github.com/Instagram/LibCST/pull/466

zsol:
This should be fixed by #466.

PR #466, opened 2021-03-13
Improve handling of raw fstrings (#462)

thatch:
## Summary

Fixes the reported issue by reverting to the regex used before 02fc4401bc3313f0db109d86d1b84c6b35abba5a when raw.

I am not 100% sure checking just top of stack is enough, however.  I wish there was a conformance suite for f-strings.

## Test Plan

Includes new test case

### DIFF
--- a/libcst/_nodes/tests/test_atom.py
+++ b/libcst/_nodes/tests/test_atom.py
@@ -683,6 +683,20 @@ class AtomTest(CSTNodeTest):
                 "parser": parse_expression,
                 "expected_position": None,
             },
+            {
+                "node": cst.FormattedString(
+                    parts=(
+                        cst.FormattedStringText("\\"),
+                        cst.FormattedStringExpression(
+                            cst.Name(value="a"),
+                        ),
+                    ),
+                    start='fr"',
+                ),
+                "code": 'fr"\\{a}"',
+                "parser": parse_expression,
+                "expected_position": None,
+            },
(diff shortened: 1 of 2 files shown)


bgw:
It doesn't seem like this problem is unique to raw f-strings? It seems that `{` and `}` characters are processed before escape characters are:

```
>>> a = "test"
>>> f"\{a}"
'\\test'
```

> I wish there was a conformance suite for f-strings.

[rant]

CPython's algorithm for this is much simpler than ours (and therefore, much less bug-prone). The decision to split f-strings in the tokenizer is one of my regrets in LibCST's design.

CPython doesn't try to split f-strings in the tokenizer. Instead, it detects f-strings when building the AST and then recursively calls the tokenizer and parser. This moves complexity out of the (already way too complex) tokenizer. We could do the same, but we didn't, and at this point, fixing it is harder than just patching these bugs when they come up.

It might make sense to change approaches when we inevitably have to rewrite much of the parser to support 3.10's soft-keywords (e.g. using a PEG grammar).

I might try to tackle this as part of my Rust porting attempts (#452). My new Rust-based tokenizer implementation supports enabling/disabling the f-string splitting behavior with a feature flag.

[/rant]

zsol:
I'm not super familiar with this part but if there's no other reviews by Monday I'll accept and merge

zsol (approved):

PR #471, opened 2021-03-31
Add Github Actions and derive version from git tags

lpetre:
## Summary
Two changes here:
 - I've ported the circleci config to github actions. I propose this as a replacement to circleci
 - I've added setuptools_scm and configured it to generate `libcst\_version.py`
 
To improve our workflow around libcst releases, I'd like to propose that we publish pre-release versions of the package. 

More context about publishing here: https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#publishing-the-distribution-to-pypi-and-testpypi

And pre-release versions here: https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning

Relating that to this change:
 - setuptools_scm makes the pre-release versioning trivial
 - We can add ~10 lines of github action yaml to handle the publishing to PyPi

## TODOs:
 - I will follow up with a commit that handles the pypi publishing
 - I will delete the circleci files
 - If this proposal is acceptable, we would need to configure some secrets in Github (ie codecov and pypi tokens)
 - Add testing on `windows-latest`. I discovered that the test suite doesn't pass on windows, which I will fix in another PR.

## Test Plan
See https://github.com/lpetre/LibCST/actions/runs/704265869

### DIFF
--- a/libcst/__init__.py
+++ b/libcst/__init__.py
@@ -191,6 +191,12 @@ from libcst._parser.types.config import (
 )
 from libcst._removal_sentinel import RemovalSentinel, RemoveFromParent
 from libcst._visitors import CSTNodeT, CSTTransformer, CSTVisitor, CSTVisitorT
+
+
+try:
+    from libcst._version import version as LIBCST_VERSION
+except ImportError:
+    LIBCST_VERSION = "unknown"
 from libcst.helpers import (  # from libcst import ensure_type is deprecated, will be removed in 0.4.0
     ensure_type,
 )
@@ -200,11 +206,6 @@ from libcst.metadata.base_provider import (
     VisitorMetadataProvider,
 )
 from libcst.metadata.wrapper import MetadataWrapper
-
-try:
-    from libcst._version import version as LIBCST_VERSION
-except ImportError:
-    LIBCST_VERSION = "unknown"
 
 
 __all__ = [
--- a/libcst/__init__.py
+++ b/libcst/__init__.py
@@ -190,7 +190,6 @@ from libcst._parser.types.config import (
     PartialParserConfig,
 )
 from libcst._removal_sentinel import RemovalSentinel, RemoveFromParent
-from libcst._version import LIBCST_VERSION
 from libcst._visitors import CSTNodeT, CSTTransformer, CSTVisitor, CSTVisitorT
 from libcst.helpers import (  # from libcst import ensure_type is deprecated, will be removed in 0.4.0
     ensure_type,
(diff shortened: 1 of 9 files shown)


tox.ini
```diff
@@ -47,40 +43,38 @@ commands =
...
     codecov
 
+[testenv:pyre]
+usedevelop=True
+allowlist_externals=
+    cp
+commands =
+    cp .github/workflows/.pyre_configuration .
+    pyre --version
+    pyre check
```
zsol, line 53: This is important to run:
```
python libcst/tests/test_pyre_integration.py
```
circle config has this but github now misses

zsol (changes requested):
I like this direction, but let's make sure the CI setup is healthy before changing the release process.

libcst/__init__.py
```diff
@@ -190,8 +190,13 @@
...
 )
 from libcst._removal_sentinel import RemovalSentinel, RemoveFromParent
-from libcst._version import LIBCST_VERSION
 from libcst._visitors import CSTNodeT, CSTTransformer, CSTVisitor, CSTVisitorT
+
+
+try:
+    from libcst._version import version as LIBCST_VERSION
+except ImportError:
+    LIBCST_VERSION = "unknown"
```
zsol, line 199: when can this condition trigger?

thatch:
If someone adds it to pythonpath without using setup.py develop or similar.

Tim

tox.ini
lpetre: I've added this and now the integration test runs.

libcst/__init__.py
lpetre: I think @thatch's comment is meant to be a reply here. 

https://github.com/Instagram/LibCST/pull/471#issuecomment-895529475

> If someone adds it to pythonpath without using setup.py develop or similar.

lpetre:
Note: I have a follow up PR w/ windows testing (and a windows fix).

tox.ini
```diff
@@ -1,5 +1,11 @@
 [tox]
-envlist = py36, py37, py38, py39, lint, docs
+envlist = py36, py37, py38, lint, docs
+
+[gh-actions]
+python =
+    3.6: py36
+    3.7: py37
+    3.8: py38
```
jimmylai, line 8: Why not `3.9`?

Move to GH actions LGTM.

zsol: Currently circleci doesn't cover 3.9. I'd prefer adding 3.9 in a separate PR, this is already complex enough

zsol (approved):
LGTM. if you resolve the conflict I can merge

tox.ini
lpetre: I'm happy to follow up w/ 3.9 support in a separate PR

lpetre:
> LGTM. if you resolve the conflict I can merge

Resolved!

tox.ini
lpetre: 3.9 run here: https://github.com/lpetre/LibCST/actions/runs/1116879480

Issue #472, opened 2021-03-31
type_inference_provider does not properly clean up child processes on timeout


lisroach:
If the subprocess call to in `type_inference_provider` times out the child processes do not get immediately cleaned up:

```
ERROR:1:1: X002 raise TimeoutExpired(
ERROR:1:1: X002 subprocess.TimeoutExpired: Command 'pyre --noninteractive query "types(path='python/fixit/flake8/test.py')"' timed out after 0 seconds
ERROR:1:1: X002 .../lib/python3.8/subprocess.py:942: ResourceWarning: subprocess 881287 is still running
ERROR:1:1: X002 _warn("subprocess %s is still running" % self.pid,
ERROR:1:1: X002 ResourceWarning: Enable tracemalloc to get the object allocation traceback
ERROR:1:1: X002 sys:1: ResourceWarning: unclosed file <_io.BufferedReader name=3>
ERROR:1:1: X002 sys:1: ResourceWarning: unclosed file <_io.BufferedReader name=5>
```

When running over a large directory a new process is spawned per path so this can become problematic for scaling.

I believe the processes don't get killed by the `TimeoutExpired`  being raised because `shell=True` is in use (https://github.com/Instagram/LibCST/blob/27ef2bf10ee71d61e658a94f97dcae1d7d94bf39/libcst/metadata/type_inference_provider.py#L106).
 StackOverflow suggests it's because the blocking process is a child of the shell process (https://stackoverflow.com/questions/48763362/python-subprocess-kill-with-timeout).

I'm trying to get a repro of this externally but haven't done so yet.

I plan to work on this, but am opening the issue for discussion and tracking.

zsol:
Easiest way to repro is probably to add a fake `pyre` wrapper on your path that sleeps for a minute before delegating to real pyre :)

lisroach:
@zsol that is a great idea, I got something working before I noticed it though:

<test_libcst.py>
```
from pathlib import Path
from libcst.metadata.type_inference_provider import TypeInferenceProvider

cache = TypeInferenceProvider.gen_cache(Path("."), ["test.py"], 0)
```

And running with:
`python3 -Wdefault test_libcst.py`

Triggers the warning like:
```
(libcst-env) lisroach@lisroach-mbp libCST % python3 -Wdefault test_libcst.py
Traceback (most recent call last):
  File "test_libcst.py", line 4, in <module>
    cache = TypeInferenceProvider.gen_cache(Path("."), ["test.py"], 0)
[...]

PR #474, opened 2021-04-01
Fix leaking processes from TypeInferenceProvider

lisroach:
## Summary
https://github.com/Instagram/LibCST/issues/472

Switches to use `subprocess.run` instead of `Popen.communicate` which will automatically kill child process for us. Also removed the usage of `shell=True` because it is unnecessary and can be a [security concern](https://docs.python.org/3/library/subprocess.html#security-considerations). 
## Test Plan
Updated existing unit tests. 

Tested manually (with `python3 -Wdefault test_libcst.py` as referenced in https://github.com/Instagram/LibCST/issues/472) to ensure the ResourceWarning went away and Zombie processes are not spawned.

`tox -e py37`


zsol:
Something's wrong with they pyre integration, you can manually check with `python libcst/tests/test_pyre_integration.py` in a venv that has pyre setup.

zsol (approved):
Nice, thanks!

PR #477, opened 2021-04-15
Return more specific QNames for assignments

zsol:
## Summary

When `scope.get_qualified_names_for()` is called with a node that's an `Assignment`, return the qualified name for that node instead of all the assignments for the same name.

For example:
```
if False:
  def f(): pass  # {"f"}
else:
  from a import f  # {"a.f"}

f()  # {"f", "a.f"}
```

Before this PR, all three nodes had the same set of qnames returned (`{"f", "a.f"}`).

## Test Plan

Will add tests before merging.

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -529,7 +529,7 @@ class Scope(abc.ABC):
                     names = _NameUtil.find_qualified_name_for_non_import(
                         assignment, full_name
                     )
-                if assignment_node is node:
+                if not isinstance(node, str) and _is_assignment(node, assignment_node):
                     return names
                 else:
                     results |= names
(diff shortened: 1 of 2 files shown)


jimmylai (approved):
I like this change which provides more accurate and useful names.
Please also add test cases and the document.

zsol:
While working on this I discovered that `QualifiedNameProvider` didn't provide names for import statements, so I added this in the commits above. This caused a completely unrelated `RenameCommand` to break because it assumes `Name` nodes inside imports don't have a qualified name (for example [here](https://github.com/Instagram/LibCST/blob/master/libcst/codemod/commands/rename.py#L235-L238)).

I'll try and fix the codemod, but may decide it's more trouble than worth and just leave the original behavior for imports:

```
if False:
  def f(): pass  # {}
else:
  from a import f  # {"a.f"}

f()  # {"f", "a.f"}
```

PR #480, opened 2021-04-27
Fix formatting for link to QualifiedName class in docs

browniebroke:
## Summary

While checking out the cods for the recently added `FullyQualifiedNameProvider`, I noticed that one link to a class wasn't properly handled by Sphinx:

image

I think the idea was to add an `s` at the end of the class name since we're talking about multiple of them, but I couldn't find how to make the `s` stick, so I've reworded a bit.

Alternative solutions I considered, with the reason why I didn't go for them:

- Specify a title with the "s" (`` :class:`QualifiedNames <QualifiedName>` ``): altering the class name might confuse people
- Add a space between the class name and the "s": formatting looks off
- Remove extra markup/formatting: seems like a step back vs having links to the class docs.

## Test Plan

Check the docs output

image


zsol (approved):
Thanks for the fix!

PR #483, opened 2021-05-01
[scope-provider] tie accesses from string annotation to the string node

### LINKED ISSUE #381: ScopeProvider sometimes returns nodes that don't appear in the CST
Consider this example code:
```
def f() -> "None":
  pass
```

Scope analysis will return an `Access` for a node like `Name(value="None")`, but there's no such CST node in the code. This is a problem because users of LibCST would expect to query metadata (for example the position) of nodes returned by scope analysis.

Here's some code to demonstrate the problem:

```
import libcst as cst
m = cst.metadata.MetadataWrapper(cst.parse_module("def f() -> \"None\": pass"))
pos = m.resolve(cst.metadata.PositionProvider)
global_scope = list(filter(None, m.resolve(cst.metadata.ScopeProvider).values()))[0]
alien_node = list(global_scope.accesses['None'])[0].node
pos[alien_node]  # BOOM
```

This produces an exception like the following:
```
KeyError: Name(
    value='None',
    lpar=[],
    rpar=[],
)
```

zsol:
## Summary
Instead of referencing `Name` nodes that are outside of the original CST, this PR makes `Access`es refer to the containing `BaseString` node (technically only `SimpleString | ConcatenatedString`).

Closes #381 

## Test Plan

Added test cases.

TODO before merge:

- [x] update docs about the type of `Access.node`

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -74,9 +74,10 @@ class Access:
 
     #: The node of the access. A name is an access when the expression context is
     #: :attr:`ExpressionContext.LOAD`. This is usually the name node representing the
-    #: access, except for dotted imports, when it might be the attribute that
-    #: represents the most specific part of the imported symbol.
-    node: Union[cst.Name, cst.Attribute]
+    #: access, except for: 1) dotted imports, when it might be the attribute that
+    #: represents the most specific part of the imported symbol; and 2) string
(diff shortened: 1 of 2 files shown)


zsol:
cc @lpetre

jimmylai (approved):
Looks great. It makes more sense to refer to the CSTNode to make further tree positioning easier.
The added DeferredAccess also improved readability.

Issue #486, opened 2021-05-20
Support PEP-614 - relaxed grammar restrictions on decorators


Zac-HD:
Following [PEP-614](https://www.python.org/dev/peps/pep-0614/), the follow code is valid in Python 3.9 and later:

```python
@(lambda f: f)  # or other expressions, e.g. `@buttons[0].clicked.connect`
def valid_on_py39():
    pass
```

but `libcst.parse_module()` raises

```python-traceback
Traceback (most recent call last):
  ...
libcst._exceptions.ParserSyntaxError: Syntax Error @ 1:2.
Incomplete input. Encountered '(', but expected 'NAME'.

@(lambda f: f)
 ^
```

This syntax isn't in common use *yet*, but [`shed`](https://pypi.org/project/shed/) depends on LibCST for `--refactor` mode and I'd love to support it :smiley:

Zac-HD:
Closed by #926 and fixed in version 0.4.10 🎉

PR #489, opened 2021-05-23
Improve decorator parsing for 3.9

thatch:
## Summary

As pointed out in #486, decorators are allowed to do more things.

## Test Plan

Ran the tests.

Why draft?

[ ] unrelated test fixes for 3.9
[ ] had to disable a bunch of validation
[ ] not sure the types are right for Decorator.decorator anymore
[ ] only one new test


thatch:
type_enforce change is a fix for #305

PR #492, opened 2021-06-22
Fix typo

zhammer:
Caught this while quoting your docs in a blog post, which I'll share once published :)

## Summary

Typo: extra word in a sentence.

## Test Plan

Just docs, no testing needed.


zsol (approved):
Thanks! :)

Issue #494, opened 2021-07-02
Codemodding fails on windows due to error calling black with env={}


Josverl:
# The issue: 

- when running any codemod on windows 10, the codemod itself is run successfully, 
- but the black formatting cannot be run due to `env={}` arguments in the code
- Note : this is also not reported as an warning or error, but as a success (`Finished codemodding 1 files!`)

## simple example of the issue

```
(.venv) PS C:\develop\MyPython\LibCST> black.exe
No Path provided. Nothing to do 😴
(.venv) PS C:\develop\MyPython\LibCST> python -m libcst.tool codemod convert_format_to_fstring.ConvertFormatStringCommand .\stubbertest\testfile.py
Calculating full-repo metadata...
Executing codemod...
Codemodding C:\develop\MyPython\LibCST\stubbertest\testfile.py
Traceback (most recent call last):
  File "C:\develop\MyPython\LibCST\libcst\codemod\_cli.py", line 327, in _execute_transform
    newcode = invoke_formatter(config.formatter_args, newcode)
  File "C:\develop\MyPython\LibCST\libcst\codemod\_cli.py", line 56, in invoke_formatter
    subprocess.check_output(
  File "C:\Users\josverl\AppData\Local\Programs\Python\Python39\lib\subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "C:\Users\josverl\AppData\Local\Programs\Python\Python39\lib\subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "C:\Users\josverl\AppData\Local\Programs\Python\Python39\lib\subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Users\josverl\AppData\Local\Programs\Python\Python39\lib\subprocess.py", line 1420, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
OSError: [WinError 87] The parameter is incorrect

Failed to codemod C:\develop\MyPython\LibCST\stubbertest\testfile.py

Finished codemodding 1 files!
 - Transformed 0 files successfully.
 - Skipped 0 files.
 - Failed to codemod 1 files.
 - 0 warnings were generated.
(.venv) PS C:\develop\MyPython\LibCST>

```
# Minimal repo 

``` python
import subprocess

code = """
print("hello {}".format("Jos"))
"""

[...]

PR #495, opened 2021-07-02
fix codemodding on windows 

Josverl:
## Summary
running codemodding fails due to passing an empty environment `env={}` rather than `env=None` according to the Python spec.
`Note If specified, env must provide any variables required for the program to execute. On Windows, in order to run a side-by-side assembly the specified env must include a valid SystemRoot.
`
by specifying env=None things work smoothly on Windows 10

fixes: #494 

## Test Plan
I have only performed limited testing on Windows, but this should be validated across all supported platforms.
My assumption that this is already part of the CI/CD tests

### DIFF
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -55,7 +55,6 @@ def invoke_formatter(formatter_args: Sequence[str], code: AnyStr) -> AnyStr:
         AnyStr,
         subprocess.check_output(
             formatter_args,
-            env=None,
             input=code,
             universal_newlines=not work_with_bytes,
             encoding=None if work_with_bytes else "utf-8",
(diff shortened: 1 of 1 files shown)


libcst/codemod/_cli.py
```diff
@@ -55,7 +55,7 @@ def invoke_formatter(formatter_args: Sequence[str], code: AnyStr) -> AnyStr:
         AnyStr,
         subprocess.check_output(
             formatter_args,
-            env={},
+            env=None,
```
zsol, line 58: just removing this parameter has the same effect as passing in `None`, so let's remove it

Josverl: sure 👌🏻

zsol (approved):

zsol:
Thx for contributing! :)

bgw:
After looking at the pre-OSS blame for this, it seems like the use of `env={}` in the first place was to work around an internal issue with an Instagram script runner. It's unclear if that issue has been resolved (though it probably has). Curious Facebook employees can look at D14541659.

zsol:
Looks like that particular error has been resolved, but I see a different one today. In any case, we should fix this issue internally

PR #502, opened 2021-08-04
Add tests verifying TypeAnnotationsVisitor current behaviors

stroxler:
No changes to production code, just adding tests to make the existing
logic self-documented and regression-proof.

## Test that LibCST TypeAnnotationsVisitor handles string annotations

This is an important property for certain use cases, so it
makes sense to verify it in tests so that we can safely
depend on it.

At present, the reason we want to be able to rely on this is:
- at the moment, imports added by infer can make pysa traces
  hard to understand, because the line numbers are off
- if we add the ability to use fully-qualified string annotations
  for the stubs from infer, then we can do so without adding
  any import lines and pyre will understand the types.

## Test various cases of merging imports

The goal here is to have LibCST's current TypeAnnotationsVisitor
behavior - which handles some types of imports and not others -
be well-documented in a test.

The reason is that import handling bugs are one of the main problems
preventing pyre infer from being 100% reliable, and to think about
how to improve it (change LibCST, change infer, change both?) we need
to have a clear understanding of what currently works, isolated in a
small test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ build/
 .coverage
 .hypothesis/
 .pyre_configuration
+.python-version
--- a/libcst/codemod/visitors/tests/test_apply_type_annotations.py
+++ b/libcst/codemod/visitors/tests/test_apply_type_annotations.py
@@ -22,6 +22,76 @@ class TestApplyAnnotationsVisitor(CodemodTest):
         (
             (
                 """
+                from __future__ import annotations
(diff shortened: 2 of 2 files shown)


.python-version
```diff
@@ -0,0 +1 @@
+libcst
```
zsol, line 1: this looks accidental

zsol (approved):

Issue #467, opened 2021-03-21
Python 3.9: codegen creates invalid file


bnavigator:
Trying to package libcst  0.3.17 for Python 3.9 (after fixing typing_inspect) I run into the following test failure:

```
[   44s] ======================================================================
[   44s] FAIL: test_codegen_clean_matcher_classes (libcst.codegen.tests.test_codegen_clean.TestCodegenClean)
[   44s] Verifies that codegen of matcher classes would not result in a
[   44s] ----------------------------------------------------------------------
[   44s] Traceback (most recent call last):
[   44s]   File "/home/abuild/rpmbuild/BUILD/libcst-0.3.16/libcst/codegen/tests/test_codegen_clean.py", line 81, in test_codegen_clean_matcher_classes
[   44s]     self.assertTrue(
[   44s] AssertionError: False is not true : libcst.matchers.__init__ needs new codegen!
[   44s] 
[   44s] ----------------------------------------------------------------------
[   44s] Ran 1788 tests in 34.980s
```

`python3.9 -m libcst.codegen.generate matchers` creates a vastly different `libcst/matchers/__init__.py` from python3.8, specifically one that creates
```
[   27s] ======================================================================
[   27s] ERROR: libcst.codegen.tests.test_codegen_clean (unittest.loader._FailedTest)
[   27s] ----------------------------------------------------------------------
[   27s] ImportError: Failed to import test module: libcst.codegen.tests.test_codegen_clean
[   27s] Traceback (most recent call last):
[   27s]   File "/usr/lib64/python3.9/unittest/loader.py", line 436, in _find_test_path
[   27s]     module = self._get_module_from_name(name)
[   27s]   File "/usr/lib64/python3.9/unittest/loader.py", line 377, in _get_module_from_name
[   27s]     __import__(name)
[   27s]   File "/home/abuild/rpmbuild/BUILD/libcst-0.3.17/libcst/codegen/tests/test_codegen_clean.py", line 12, in <module>
[   27s]     from libcst.codegen.generate import clean_generated_code, format_file
[   27s]   File "/home/abuild/rpmbuild/BUILD/libcst-0.3.17/libcst/codegen/generate.py", line 21, in <module>
[   27s]     from libcst.codegen.transforms import (
[   27s]   File "/home/abuild/rpmbuild/BUILD/libcst-0.3.17/libcst/codegen/transforms.py", line 18, in <module>
[   27s]     import libcst.matchers as m
[...]

PR #506, opened 2021-08-10
Adding python3.9 to the CI

lpetre:
## Summary

In my previous PR there was some discussion that we weren't running 3.9 in the CI, this adds that.

I've had to update the matchers codegen or else the tests fail. Please advise!

## Test Plan

See GH Actions run


lpetre:
Something needs to change here, the codegen generates different code for >=3.9 and <3.9, so the tests are going to fail somewhere.

As I see it:
 - either we fix the codegen to generate identical code for both cases
 - or we need to only run the codegen tests on the same version which is responsible for generating the code

lpetre:
This is blocked by #467

zsol:
I commented on the linked issue with how I think this could be solved. ;)

zsol (approved):

PR #508, opened 2021-08-11
Change codegen to treat typing.Union[Foo, NoneType] and typing.Optional[Foo] as the same

lpetre:
## Summary

This is a fix for #467. As per [this comment](https://github.com/Instagram/LibCST/issues/467#issuecomment-896329824), the type of optional fields in dataclasses has changed in python 3.9.

This change makes the codegen rewrite the old type to the new type, so it works the same in all versions.

## Test Plan

Codegen tests


libcst/codegen/gen_matcher_classes.py
```diff
@@ -396,8 +399,8 @@ def _get_clean_type_and_aliases(
 
     # First, get the type as a parseable expression.
     typestr = repr(typeobj)
-    if typestr.startswith("<class '") and typestr.endswith("'>"):
-        typestr = typestr[8:-2]
+    typestr = re.sub(CLASS_RE, r"\1", typestr)
+    typestr = re.sub(OPTIONAL_RE, r"typing.Optional[\1]", typestr)
```
zsol, line 403: This won't work for `Union[A, B, NoneType]` or `Union[A, None]`. While this might not cause an issue today, I don't want to be on the debugging end if it pops up in the future :)

lpetre: Why don't it work for `Union[A, B, NoneType]`, my regex rejects that.

I'm not sure I understand the difference between `Union[A, NoneType]` and `Union[A, None]`

zsol: Hm, fair enough. apparently `Option[Union[A, B]]` always gets flattened to `Union[A, B, NoneType]` by dataclasses, so it doesn't even matter.

zsol (approved):

PR #511, opened 2021-08-11
Split up some larger CI/tox steps into separate jobs

lpetre:
## Summary

This takes the runtime of the CI from ~8 minutes to <5 mins.

There is some unfortunate duplication, and I wish `needs` could depend on one matrix configuration.

Otherwise I'm very happy with this change.

This also enables me to ship windows testing, as I don't rely on tox anymore.

## Test Plan
See https://github.com/lpetre/LibCST/actions/runs/1121624876

### DIFF
--- a/.github/workflows/.pyre_configuration.template
+++ b/.github/workflows/.pyre_configuration.template
@@ -3,7 +3,7 @@
         "."
     ],
     "search_path": [
-        "stubs", "~/cache/tox/pyre/lib/python3.9/site-packages/"
+        "stubs", "{{ .python_site_packages }}"
     ],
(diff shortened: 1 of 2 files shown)


zsol (approved):
this is great! thx

lpetre:
> this is great! thx

👍 I was able to move the copy/pasta into a local action file, phew!

lpetre:
> > this is great! thx
> 
> 👍 I was able to move the copy/pasta into a local action file, phew!

Actually that didn't work at all. The composite run actions doesn't support `uses` or `if` which makes it worthless for my usecase.

PR #513, opened 2021-08-17
Fix black formatting failure on tokenize.pyi

stroxler:
## Summary

I noticed this file was mal-formatted when using `black .` to
autoformat a different branch. Let's keep the autoformat clean.

## Test Plan

```
python -m unittest
```
All tests are clean except for skipped tests / expected failures


stroxler:
Closing - this PR is correct in that black doesn't like our stubs file, but if I autoformat it then isort rejects the change.

Seems like a configuration issue, I can look into whether there's a way to make them agree

stroxler:
Reopening with a fix that disables isort on .pyi files

As discussed in the summary we could also try to raise this issue upstream - the isort team seems to think they closed this bug but their fix isn't working for me

stroxler:
On second thought, leaving closed, I think https://github.com/Instagram/LibCST/pull/515 is a better approach

PR #515, opened 2021-08-18
Use `profile = "black"` for isort config. This changes the blank lines after imports.

stroxler:
## Summary

I'm still trying to make black and isort agree with one another. This is the
recommended approach - use `profile = "black"` for isort.

If we do this, it looks like isort forces one line after imports                
everywhere (not just .pyi files but also .py files) and black is                
okay with it.                                                                   
                                                                                
I think this indicates that black will leave 1 or 2 lines as-is                 
in `.py` files - it accepts either but only accepts 1 line in                   
`.pyi` files. Whereas as far as I can tell isort only allows us                 
to use the same rule in both, which means it has to be 1 line                   
for black compatibility on `.pyi` files.  

## Test Plan

```
# first, make both tools happy
black .
isort .

# then, check that the format is stable
black .   # no change
isort .    # no change
```


jreese:
I would personally prefer switching LibCST to use [µfmt](https://ufmt.readthedocs.io), which combines black and [µsort](https://usort.readthedocs.io), an alternative to isort that was built on top of LibCST. This gives a single, atomic command that both sorts imports and formats sources in a consistent and stable manner, with minimal configuration.

stroxler:
@zsol thoughts on using µsort instead?

zsol:
Yeah I'm with @jreese on this one

PR #523, opened 2021-09-07
Fix pyre command for type inference provider

rodrigozhou:
## Summary

Pyre command for type inference provider has one of its params quoted and causing the Pyre to fail to parse the command line.

## Test Plan

Added unit test to generate the cache and compare with the expected result.

### DIFF
--- a/libcst/metadata/tests/test_type_inference_provider.py
+++ b/libcst/metadata/tests/test_type_inference_provider.py
@@ -9,6 +9,7 @@ import os
 import subprocess
 import sys
 from pathlib import Path
+from unittest import skipIf
 
 import libcst as cst
 from libcst import MetadataWrapper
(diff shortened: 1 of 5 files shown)


zsol:
Looks like the new test is failing because `pyre` is not started. Might need to wrap the test code in a context manager or decorator that runs `pyre start --no-watchman` and tears it down afterwards (`pyre stop`)

zsol:
cc @stroxler if the quotes are causing a problem, how are you not running into it?

rodrigozhou:
@zsol Updated to start and stop Pyre in test setUp and tearDown functions.

stroxler:
My changes bumped the pyre version which did seem to trigger a pyre start somewhere because I had to add the `--no-watchman` flag, but it sounds like we probably didn't start pyre pointed at the test code (my guess is that we were starting it up pointed at LibCST itself)

The error we're getting makes me think we lack a `.pyre_configuration` json file in the test fixture.

Assuming I'm correct about this being the problem, we could fix that by manually adding one, or probably (according to https://pyre-check.org/docs/getting-started#setting-up-a-project) by running `pyre init` prior to `pyre start`.

zsol (approved):
lets go

PR #538, opened 2021-10-28
Support relative imports in ATAV qualifier handling

stroxler:
## Summary

Based on diff review of https://github.com/Instagram/LibCST/pull/536,
I investigated relative import handling and realized that with minor
changes we can now handle them correctly.

Relative imports aren't likely in code coming from an automated
tool, but they could happen in hand-written stubs if anyone tries
to use this codemod tool to merge stubs with code.

## Test Plan

Added a new test:
```
.............................................
Ran 45 tests in 2.195s

OK

```


zsol (approved):
lgtm

PR #539, opened 2021-10-28
Do not add imports if we added no type info in ATAV

stroxler:
## Summary

In ApplyTypeAnnotationsVisitor, there are edge cases where we
might have changed the module imports even though we never wound
up applying any type annotations.

This will become even more common if we support adding
`from __future__ import annotations`, which I would like to do
soon.

To handle this, we can simply return the original tree from
`transform_module_impl` (discarding any changes from either
`self` or `AddImportsVisitor`) whenever there are no changes
in `self.annotation_counts`.

## Test Plan

I updated the no-annotations-changed test to reflect this:
```
> python -m unittest libcst.codemod.visitors.tests.test_apply_type_annotations.TestApplyAnnotationsVisitor
...............................................
Ran 47 tests in 2.312s

OK
```

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -266,7 +266,7 @@ class AnnotationCounts:
     return_annotations: int = 0
     classes_added: int = 0
 
-    def applied_changes(self):
+    def any_changes(self):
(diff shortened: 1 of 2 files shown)


stroxler:
Closing, because #540 included this commit

PR #540, opened 2021-10-28
Use `from __future__ import annotations` in ATAV

stroxler:
## Summary

Add a new argument to `ApplyTypeAnnotationsVisitor` that will trigger it
to add a `from __future__ import annotations` import to the top of a module.

This is a flag because it won't work on python 3.6 (in that case, users should
go in and manually quote types as needed or use quoted type names in the
stub file to begin with).

But for python 3.7+ it is a powerful tool to make adding
type annotations safer - malformed types (including types that are actually
valid for typecheckers like `os.PathLike[str]`) won't lead to runtime errors
and it's safe to use types from the current module anywhere without quoting them.

NOTE: This change should be merged only after we've merged
https://github.com/Instagram/LibCST/pull/539
to keep a clean history.

## Test Plan

```
.................................................
Ran 49 tests in 2.415s

OK
```

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -266,7 +266,7 @@ class AnnotationCounts:
     return_annotations: int = 0
     classes_added: int = 0
 
-    def applied_changes(self):
+    def any_changes(self):
         return (
             self.global_annotations
             + self.attribute_annotations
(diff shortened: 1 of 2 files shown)


libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -266,7 +266,7 @@ class AnnotationCounts:
     return_annotations: int = 0
     classes_added: int = 0
 
-    def applied_changes(self):
+    def any_changes(self):
```
shannonzhu, line 269: super nit - `changes_applied` maybe what the original name is going for?

shannonzhu (approved):
Looks good to me!

PR #542, opened 2021-11-01
Fix typing errors on ATAV

stroxler:
## Summary

We have type hints on LibCST but pyre doesn't seem to be fully hooked up.

As a result, the code was fully annotated, and was type-error-free at
runtime, but had multiple places where pyre complained either due to
type refinement not working or type errors coming from branches of
the code that weren't reachable but couldn't be statically ruled out.

## Test Plan

I ran unit tests, and vendored the changes into the pyre project to see that they
typecheck.


thatch (approved):
LG

PR #545, opened 2021-11-08
Don't gather metadata if the wrapper already contains it

lpetre:
## Summary
The MetadataWrapper is meant to resolve METADATA_DEPENDENCIES and prevent a single provider from being invoked multiple times. 

This logic was flawed if the wrapper already contained a provider. 

## Test Plan
python3 -m unittest libcst.metadata.tests.test_metadata_wrapper.MetadataWrapperTest

### DIFF
--- a/libcst/metadata/wrapper.py
+++ b/libcst/metadata/wrapper.py
@@ -74,7 +74,7 @@ def _resolve_impl(
     as well as their dependencies.
     """
     completed = set(wrapper._metadata.keys())
-    remaining = _gather_providers(providers, set()) - completed
+    remaining = _gather_providers(set(providers), set()) - completed
 
     while len(remaining) > 0:
         batchable = set()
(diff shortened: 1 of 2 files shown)


amyreese (approved):

libcst/metadata/wrapper.py
```diff
@@ -73,10 +73,9 @@ def _resolve_impl(
     Updates the _metadata map on wrapper with metadata from the given providers
     as well as their dependencies.
     """
-    providers = set(providers) - set(wrapper._metadata.keys())
-    remaining = _gather_providers(providers, set())
+    completed = set(wrapper._metadata.keys())
+    remaining = _gather_providers(providers, set()) - completed
```
jreese, line 77: Does this need to do `set(providers)` here to match previous behavior?

Issue #491, opened 2021-06-21
ScopeProvider fails when string type annotation is unparseable


zsol:
Example:
```
src = 'a: "{hello{foo}}"'
cst.MetadataWrapper(cst.parse_module(src)).resolve(cst.metadata.ScopeProvider)
```

zsol:
ScopeProvider should fail gracefully in these cases. It's probably enough to catch the exception from the internal `parse_expression` call during `_handle_string_annotation`.

srisatya1197:
@zsol will work on this issue.

zsol:
Great, thank you!

srisatya1197:
@zsol Are there any contributing guidleines, Like do I have to install any packages or tools?

zsol:
Check out [the readme](https://github.com/Instagram/LibCST/blob/master/README.rst#development) and [CONTRIBUTING.md](https://github.com/Instagram/LibCST/blob/master/CONTRIBUTING.md)

srisatya1197:
Thank you so much.

zzl0:
@zsol I was trying to find some "good first issue", it seems this issue has already been fixed by @lpetre .

zsol:
That's right, thanks for the heads up :)

PR #546, opened 2021-11-10
Stop parsing string annotations when no longer in a typing call

lpetre:
## Summary

```
# this should parse a string annotation
TypeVar("T", "int")

# this should not
TypeVar("T", func("int"))
```

To correct handle these nested function calls I've turned the type hint tracking in the ScopeVisitor into a stack rather than a set.

## Test Plan

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -789,10 +789,8 @@ class ScopeVisitor(cst.CSTVisitor):
         self.scope: Scope = GlobalScope()
         self.__deferred_accesses: List[DeferredAccess] = []
         self.__top_level_attribute_stack: List[Optional[cst.Attribute]] = [None]
-        self.__in_annotation: Set[
-            Union[cst.Call, cst.Annotation, cst.Subscript]
-        ] = set()
-        self.__in_type_hint: Set[Union[cst.Call, cst.Annotation, cst.Subscript]] = set()
+        self.__in_annotation_stack: List[bool] = [False]
+        self.__in_type_hint_stack: List[bool] = [False]
         self.__in_ignored_subscript: Set[cst.Subscript] = set()
         self.__last_string_annotation: Optional[cst.BaseString] = None
         self.__ignore_annotation: int = 0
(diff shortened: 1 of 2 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -789,10 +789,8 @@ def __init__(self, provider: "ScopeProvider") -> None:
         self.scope: Scope = GlobalScope()
         self.__deferred_accesses: List[DeferredAccess] = []
         self.__top_level_attribute_stack: List[Optional[cst.Attribute]] = [None]
-        self.__in_annotation: Set[
-            Union[cst.Call, cst.Annotation, cst.Subscript]
-        ] = set()
-        self.__in_type_hint: Set[Union[cst.Call, cst.Annotation, cst.Subscript]] = set()
+        self.__in_annotation_stack: List[bool] = [False]
+        self.__in_type_hint_stack: List[bool] = [False]
         self.__in_ignored_subscript: Set[cst.Subscript] = set()
```
lpetre, line 794: I actually think I can remove `__in_ignored_subscript` and implement it by pushing False into `__in_annotation_stack`

lpetre:
I didn't make the suggested change in #491 because this test is expecting that to fail for typing calls:
https://github.com/Instagram/LibCST/blob/main/libcst/metadata/tests/test_scope_provider.py#L1713-L1768

libcst/metadata/scope_provider.py
```diff
@@ -851,33 +849,37 @@ def visit_Attribute(self, node: cst.Attribute) -> Optional[bool]:
...
     def visit_Call(self, node: cst.Call) -> Optional[bool]:
         self.__top_level_attribute_stack.append(None)
+        self.__in_type_hint_stack.append(False)
+        self.__in_annotation_stack.append(False)
         qnames = {qn.name for qn in self.scope.get_qualified_names_for(node)}
         if "typing.NewType" in qnames or "typing.TypeVar" in qnames:
             node.func.visit(self)
-            self.__in_type_hint.add(node)
+            self.__in_type_hint_stack.pop()
+            self.__in_type_hint_stack.append(True)
```
zsol, line 858: This is almost 2x faster
```suggestion
            self.__in_type_hint_stack[-1] = True
```

zsol (approved):

zsol:
I'd be happy to remove those offending tests if that'd make things easier. I think they're just there to document existing behavior.

libcst/metadata/scope_provider.py
```diff
@@ -851,33 +849,36 @@ def visit_Attribute(self, node: cst.Attribute) -> Optional[bool]:
 
     def visit_Call(self, node: cst.Call) -> Optional[bool]:
         self.__top_level_attribute_stack.append(None)
+        self.__in_type_hint_stack.append(False)
+        self.__in_annotation_stack.append(False)
```
lpetre, line 853: Appending to `__in_annotation_stack` here was an error

PR #548, opened 2021-11-17
Swallow parsing errors in string annotations.

lpetre:
## Summary
This is the same behavior as cPython.

I've also rewritten the test that was relying on this exception to check where type parsing was happening

## Test Plan

```
(libcstvenv) lpetre@lpetre-mbp LibCST % python -m unittest libcst.metadata.tests.test_scope_provider.ScopeProviderTest          
................................................s..
Ran 51 tests in 2.834s

OK (skipped=1)
(libcstvenv) lpetre@lpetre-mbp LibCST % 
```


lpetre:
Fixes #491

zsol (approved):
Thank you!

PR #549, opened 2021-11-17
Pass absolute path to codemod test

lpetre:
## Summary
This test is breaking in #548 

## Test Plan

 python -m unittest libcst.codemod.tests.test_codemod_cli


zsol (approved):
Maybe skip this test on windows if python < 3.8?

PR #551, opened 2021-11-18
disable setuptools local scheme when env var is set

zsol:
## Summary
When `LIBCST_NO_LOCAL_SCHEME` is set while packaging, the local scheme will be removed from the version number (i.e. `0.3.22.dev16` instead of `0.3.22.dev16+g5c05001.d20211118`).
## Test Plan

```
env LIBCST_NO_LOCAL_SCHEME=1 python setup.py sdist
ls dist/libcst-0.3.22.dev16.tar.gz
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,6 +182,7 @@ jobs:
       if: steps.cache.outputs.cache-hit != 'true'
       run: exit 1
     - name: Disable scmtools local scheme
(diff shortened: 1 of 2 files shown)


lpetre:
Can you set the env var in the github action where we publish?

lpetre (approved):
LGTM. Fix the pyre error please

PR #554, opened 2021-11-18
Pin accesses to import alias node

### LINKED ISSUE #485: Scope provider should pin accesses to the ImportAlias node
In the following:
```
from a import b, c
b()
```
The access to `b` has a single `Assignment` as the referent and this assigment's `.node` is the entire `ImportFrom` statement. This is both inaccurate and problematic because:
1. client code has to do additional work to figure out which imported symbol the access is actually referring to
2. when the access is coming from a string annotation it's unreasonably hard to figure out which symbol is being referred to

```
from a import b, c
from typing import List, Optional
x: "Optional[List[b]]"
```
There are three accesses in that string annotation.

I propose the assignment generated from import aliases should be attached to the import alias itself instead of the whole import statement.

Unfortunately this will break a lot of code assuming the current behavior.

giomeg:
## Summary
Add ImportAssignment class extending Assignment to record assignments for import statements.

## Test Plan
python -m unittest libcst.codemod.visitors.tests.test_gather_unused_imports

Closes #485


lpetre (approved):
This looks great to me, thanks for working on this.

libcst/metadata/scope_provider.py
```diff
@@ -416,9 +446,12 @@ def _contains_in_self_or_parent(self, name: str) -> bool:
         """Overridden by ClassScope to hide it's assignments from child scopes."""
         return name in self
 
-    def _record_assignment_as_parent(self, name: str, node: cst.CSTNode) -> None:
+    def _add_assignment(self, assignment: "BaseAssignment") -> None:
+        self._assignments[assignment.name].add(assignment)
+
+    def _add_assignment_as_parent(self, assignment):
```
lpetre, line 452: Please add typing here

zsol:
lgtm

libcst/metadata/scope_provider.py
```diff
@@ -401,10 +418,23 @@ def __init__(self, parent: "Scope") -> None:
         self._assignment_count = 0
 
     def record_assignment(self, name: str, node: cst.CSTNode) -> None:
-        self._assignments[name].add(
+        self._add_assignment(
             Assignment(name=name, scope=self, node=node, index=self._assignment_count)
```
lpetre, line 422: Since scope is set here, the old behavior of overriding `record_assignment` no longer changes which scope is recorded in the Assignment instance

libcst/metadata/scope_provider.py
```diff
@@ -416,9 +446,12 @@ def _contains_in_self_or_parent(self, name: str) -> bool:
         """Overridden by ClassScope to hide it's assignments from child scopes."""
         return name in self
 
-    def _record_assignment_as_parent(self, name: str, node: cst.CSTNode) -> None:
+    def _add_assignment(self, assignment: "BaseAssignment") -> None:
+        self._assignments[assignment.name].add(assignment)
```
lpetre, line 450:
```suggestion
        assignment.scope = self
        self._assignments[assignment.name].add(assignment)
```

libcst/metadata/scope_provider.py
giomeg: Thanks Luke, this was very frustrating!

PR #556, opened 2021-11-19
Correct and simplify logic of recording assignments

giomeg:
## Summary
Correctly records scope assignments and simplifies some logic of resolving target scope.

## Test Plan

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -418,22 +418,32 @@ class Scope(abc.ABC):
         self._assignment_count = 0
 
     def record_assignment(self, name: str, node: cst.CSTNode) -> None:
-        self._add_assignment(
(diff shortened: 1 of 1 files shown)


lpetre (approved):
Thanks for the quick fix. Lets continue talking about the need for `_find_assignment_target_parent`

libcst/metadata/scope_provider.py
```diff
@@ -446,14 +456,6 @@ def _contains_in_self_or_parent(self, name: str) -> bool:
         """Overridden by ClassScope to hide it's assignments from child scopes."""
         return name in self
 
-    def _add_assignment(self, assignment: "BaseAssignment") -> None:
-        assignment.scope = self
```
lpetre, line 450: @zsol FYI the bug was here, we were changing assignment.scope but not assignment._index

PR #557, opened 2021-11-19
Move find_qualified_names_for in the Assignment class.

giomeg:
## Summary
Move _NameUtil.find_qualified_name_for ... method inside Assignment classes.
For non imports we move this implementation inside Assignment.
For Imports we move this implementation inside ImportAssignment. 

## Test Plan

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -128,6 +128,23 @@ class Access:
         self.__assignments |= previous_assignments
 
 
+class QualifiedNameSource(Enum):
+    IMPORT = auto()
+    BUILTIN = auto()
+    LOCAL = auto()
+
+
+@add_slots
+@dataclass(frozen=True)
+class QualifiedName:
+    #: Qualified name, e.g. ``a.b.c`` or ``fn.<locals>.var``.
+    name: str
+
+    #: Source of the name, either :attr:`QualifiedNameSource.IMPORT`, :attr:`QualifiedNameSource.BUILTIN`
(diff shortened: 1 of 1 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -128,6 +128,23 @@ def record_assignments(self, name: str) -> None:
         self.__assignments |= previous_assignments
 
 
+class QualifiedNameSource(Enum):
```
lpetre, line 131: Do you need to move these?

libcst/metadata/scope_provider.py
```diff
@@ -175,6 +192,13 @@ def _index(self) -> int:
         """Return an integer that represents the order of assignments in `scope`"""
         return -1
 
+    def get_qualified_names_for(
+        self, get_qualified_names_for: str
+    ) -> Set[QualifiedName]:
+        raise NotImplementedError(
```
lpetre, line 198: Use abc.abstractmethod

lpetre (approved):
A few nits, please address them before landing

libcst/metadata/scope_provider.py
giomeg: If I don't do this I am getting:

```
NameError: name 'QualifiedName' is not defined
```

libcst/metadata/scope_provider.py
giomeg: I can't do that either because of:

```
Can't instantiate abstract class BuiltinAssignment with abstract methods get_qualified_names_for
```

PR #275, opened 2020-03-27
Support most Python 2 syntax back to 2.3

thatch:
## Summary

This now parses a good amount of real-world `setup.py` code, and I've validated back to 2.3 (the earliest version I have handy).  It passes most of [python-grammar-changes](https://github.com/thatch/python-grammar-changes) with only a few remaining failures:

```
                                                  23 24 25 26 27 30 31 32 33 34 35 36 37 38
examples/py27-0024-as-keyword.py                  o. o. o  .. .. .. .. .  .. .  .. .. .. .. 
examples/py30-0004-unpacking-args1.py             o. o. o  o. o. .. .. .  .. .  .. .. .. .. 
examples/py30-0013-kwargs-after-star.py           .o .o .  oo oo oo oo o  oo o  oo oo oo oo 
examples/py30-0018-except-as.py                   .o .o .  oo oo oo oo o  oo o  oo oo oo oo 
examples/py30-0023-relative-imports1.py           .o .o o  oo oo oo oo o  oo o  oo oo oo oo 
examples/py30-0023-relative-imports2.py           .o .o o  oo oo oo oo o  oo o  oo oo oo oo 
```

Both `relative-imports` and `except-as` are important.  `unpacking-args` is a feature I haven't seen used since about 2004.  The other two are for additional strictness to reject bad code, but are not necessary for my immediate needs.

Integer validation is reasonably strict at parse time, and handles both the octal changes and `L` suffix change.

```
                                                  23 24 25 26 27 30 31 32 33 34 35 36 37 38
examples/py30-0030-int-literals1.py               oo oo o  oo oo .. .. .  .. .  .. .. .. .. 
examples/py30-0030-int-literals2.py               oo oo o  oo oo .. .. .  .. .  .. .. .. .. 
examples/py30-0030-int-literals3.py               oo oo o  oo oo .. .. .  .. .  .. .. .. .. 
examples/py30-0030-int-literals4.py               .. .. .  oo oo oo oo o  oo o  oo oo oo oo 
examples/py30-0030-int-literals5.py               .. .. .  .. .. .. .. .  .. .  .. oo oo oo 
Legend:
 green header means will test with libcst

 first result is python, second [optional] result is libcst
 o   parses
 .   does not parse
```

New classes have `Py2`-prefixed names, to make it a conscious decision for people so they don't accidentally using them when constructing trees by hand.

## Test Plan

```
$ ../libcst-venv/bin/pyre
 ƛ No type errors found
$ tox -e py36
...
[...]

PR #566, opened 2021-12-17
Implement a Python PEG parser in Rust

zsol:
## Summary

This massive PR implements an alternative Python parser that will allow LibCST to parse Python 3.10's new grammar features. The parser is implemented in Rust, but it's turned off by default through the `LIBCST_PARSER_TYPE` environment variable. Set it to `native` to enable. The PR also enables new CI steps that test just the Rust parser, as well as steps that produce binary wheels for a variety of CPython versions and platforms.

Note: this PR aims to be roughly feature-equivalent to the main branch, so it doesn't include new 3.10 syntax features. That will be addressed as a follow-up PR.

The new parser is implemented in the `native/` directory, and is organized into two rust crates: `libcst_derive` contains some macros to facilitate various features of CST nodes, and `libcst` contains the `parser` itself (including the Python grammar), a `tokenizer` implementation by @bgw, and a very basic representation of CST `nodes`. Parsing is done by
1. **tokenizing** the input utf-8 string (bytes are not supported at the Rust layer, they are converted to utf-8 strings by the python wrapper)
2. running the **PEG parser** on the tokenized input, which also captures certain anchor tokens in the resulting syntax tree
3. using the anchor tokens to **inflate** the syntax tree into a proper CST

These steps are wrapped into a high-level `parse_module` API [here](https://github.com/zsol/LibCST/blob/parser/native/libcst/src/lib.rs#L43), along with `parse_statement` and `parse_expression` functions which all just accept the input string and an optional encoding.

These Rust functions are exposed to Python [here](https://github.com/zsol/LibCST/blob/parser/native/libcst/src/py.rs) using the excellent [PyO3](https://pyo3.rs/) library, plus an `IntoPy` trait which is mostly implemented via a macro in `libcst_derive`.

## Test Plan

Apart from Rust unit tests and roundtrip tests, the new implementation is also tested by running the pure python tests on them (by setting `LIBCST_PARSER_TYPE=native` before running `python -m unittest`). All of these are part of CI steps.

## Known missing features

Apart from the soon-to-come 3.10 syntactic features mentioned above, here's an incomplete list of the features the Rust implementation lacks compared to the pure Python parser:

* No support for limiting the parser to a certain Python version. The parser will accept input that's valid in any of Python 3.9, 3.8, 3.7, or 3.6 at the moment.
* The native parser API only accepts utf-8 encoded strings. Support for bytes is provided in the wrapping python layer.
* In some places the old parser returned a tuple of CSTNodes, the new parser returns a list of them (see adjusted unit tests)


bgw:
Oh my god. The mad lad, @zsol actually did it. Congratulations!

I don't think I can reasonably read through all this code, but I might look through it a bit later. None of that should prevent you from merging though.

zsol:
Thanks! I'd still appreciate any review or thoughts for even parts of the code, because I'm 100% sure there are some questionable decisions in there that would benefit from ... questioning :) Can incorporate any changes in future PRs.

stroxler:
Wow, there's a lot here! I'll also try to read through parts of it.

My view is that if tests pass we can probably go ahead and merge it even if there are questionable bits as long as the CST data itself is reasonable, nothing prevents us from revisiting the code later and getting 3.10 PEG support landed is pretty time-sensitive.

Before I try to dive in, a few high level questions:

(1) are there any particular parts of the code I should focus on for review? I'm coming off of making some changes to the cpython parser for PEP 677 so I can try to focus on that, but there's a big difference between making one change, even a nontrivial one, versus trying to take in the whole thing at once!

(2) are there any changes you'd like to make on top of this? It might help me build context to not just read the code but make a small modification or two if there's something you already have in mind.

(3) for code that can parse using either the old or new parser, is the produced CST the same? My impression is yes given your description but just want to double-check. In my mind even if there are messy bits, if this is true we should err on the side of merging fast, especially since it's off by default.

zsol:
> are there any particular parts of the code I should focus on for review? I'm coming off of making some changes to the cpython parser for PEP 677 so I can try to focus on that, but there's a big difference between making one change, even a nontrivial one, versus trying to take in the whole thing at once!

In that case, I'd recommend `grammar.rs` which implements step 2 from the PR description. Specifically I'm a bit worried about accepting syntax that shouldn't be accepted.

> are there any changes you'd like to make on top of this? It might help me build context to not just read the code but make a small modification or two if there's something you already have in mind.

Yeah, I'd like to add match statements myself, but you could try extending the grammar to support parenthesized context managers (https://bugs.python.org/issue12782)

> for code that can parse using either the old or new parser, is the produced CST the same?

Yes! The only case where this isn't true is the tuple-vs-list problem I mention in the PR description. PyO3 automatically converts rust's `List<_>` into python `List[...]`, but in some cases the pure python parser produces a tuple.

libcst/_nodes/tests/test_import.py
```diff
@@ -397,9 +397,11 @@ class ImportFromCreateTest(CSTNodeTest):
                         cst.ImportAlias(cst.Name("bar"), comma=cst.Comma()),
                         cst.ImportAlias(cst.Name("baz"), comma=cst.Comma()),
                     ),
+                    lpar=cst.LeftParen(),
+                    rpar=cst.RightParen(),
                 ),
-                "code": "from foo import bar,baz,",
-                "expected_position": CodeRange((1, 0), (1, 23)),
+                "code": "from foo import (bar,baz,)",
```
thatch, line 403: Curious if this is intentional -- the old syntax w/o parens is valid.  Does the new parser reject it?

libcst/_nodes/tests/test_atom.py
```diff
@@ -1120,6 +1121,8 @@ def test_invalid(self, **kwargs: Any) -> None:
         )
     )
     def test_versions(self, **kwargs: Any) -> None:
+        if is_native() and not kwargs.get("expect_success", True):
+            self.skipTest("parse errors are disabled for native parser")
```
thatch, line 1125: Do you have a plan for making them work?  This is an important part of the API.

libcst/_parser/detect_config.py
```diff
@@ -114,6 +114,23 @@ def _detect_future_imports(tokens: Iterable[Token]) -> FrozenSet[str]:
     return frozenset(future_imports)
```
thatch, line 114: I'm not 100% sure this has good tests.  Does this avoid having to tokenize twice?

thatch:
Can you expand more on

> No support for limiting the parser to a certain Python version

Are you intending on adding that before a non-pre release?  I put a decent chunk of effort in supporting old versions and even have a draft going back to 2.3 that I'd be happy to rework if version-dependent support existed. See https://github.com/Instagram/LibCST/pull/275.

libcst/_nodes/tests/test_atom.py
zsol: Yeah it's possible to make it work, but my priority is to get 3.10 working first. The old parser won't be removed until this feature is supported by the new one.

libcst/_parser/detect_config.py
zsol: Hmm, I don't think this code is ever getting run in native mode. Will probably have to be reimplemented on the rust side in the future. As far as I can tell this is only being used to figure out the exact grammar version to be used.

stroxler:
@thatch my impression of what @zsol was saying about python versions is just that the parser won't throw errors if you use Python 3.9 syntax in a 3.6 codebase.

In my mind this isn't a huge problem, generally there are other ways to get pretty fast feedback about syntax errors, and it's not critical that LibCST deal with it... if there's an exception to that I'd say it's actually on the *output* of codemods, not the input, but presumably that's not the parser anyway

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+
+        rule strings() -> String<'a>
+            = s:(str:tok(STRING, "STRING") t:&_ {(make_string(str), t)}
+                / str:fstring() t:&_ {(String::Formatted(str), t)})+ {
+                make_strings(s)
+            }
+
+        rule tuple() -> Expression<'a>
+            = lpar:lpar() first:star_named_expression() &lit(",")
+                rest:(c:comma() e:star_named_expression() {(c, e)})*
```
stroxler, line 773: [nit / curiosity] The official grammar uses `star_named_expression ', ' [star_named_expressions]` here.

This code is needed to handle empty tuples... I'm actually having trouble figuring out where the official grammar does that.

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+            }
+
+        rule try_stmt() -> Try<'a>
+            = kw:lit("try") lit(":") b:block() f:finally_block() {
+                make_try(kw, b, vec![], None, Some(f))
+            }
+            / kw:lit("try") lit(":") b:block() ex:except_block()+ el:else_block()?
+                f:finally_block()? {
+                    make_try(kw, b, ex, el, f)
+            }
```
stroxler, line 796: [compat] for 3.11 we'll need except_star_block handling. Makes sense not to have yet

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+        rule function_def() -> FunctionDef<'a>
+            = d:decorators() f:function_def_raw() {f.with_decorators(d)}
+            / function_def_raw()
+
+        rule decorators() -> Vec<Decorator<'a>>
+            = (at:lit("@") e:named_expression() nl:tok(NL, "NEWLINE") {
+                make_decorator(at, e, nl)
+                } )+
+
+        rule _returns() -> Annotation<'a>
```
stroxler, line 835: [nit] most of the time we have helper rules after, this one is before.

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+                c:lit(":") b:block() el:else_block()? {
+                    make_for(None, f, t, i, it, c, b, el)
+            }
+            / asy:tok(Async, "ASYNC") f:lit("for") t:star_targets() i:lit("in")
+                it:star_expressions()
+                c:lit(":") b:block() el:else_block()? {
+                    make_for(Some(asy), f, t, i, it, c, b, el)
+            }
+
+        rule with_stmt() -> With<'a>
```
stroxler, line 973: [compat] need to add parenthesized with statements, which also allow trailing commas (later)

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+                col:lit(":") b:block() {
+                    make_with(None, kw, comma_separate(items.0, items.1, None), col, b)
+            }
+            / asy:tok(Async, "ASYNC") kw:lit("with") items:separated(<with_item()>, <comma()>)
+                col:lit(":") b:block() {
+                    make_with(Some(asy), kw, comma_separate(items.0, items.1, None), col, b)
+            }
+
+        rule with_item() -> WithItem<'a>
+            = e:expression() a:lit("as") t:star_target() &(lit(",") / lit(":")) {
```
stroxler, line 984: [compat] needs to include `)` in the peek ahead once we do parenthesized with

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+            / e:expression() {
+                make_with_item(e, None, None)
+            }
+
+        rule named_expression() -> Expression<'a>
+            = a:name() op:lit(":=") b:expression() {
+                Expression::NamedExpr(make_named_expr(a, op, b))
+            }
+            / e:expression() !lit(":=") { e }
+
```
stroxler, line 996: [compat] this is the start of another section where our ordering diverges wildly from the official grammar

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+        rule annotated_rhs() -> Expression<'a>
+            = yield_expr() / star_expressions()
+
+        rule global_stmt() -> Global<'a>
+            = kw:lit("global") init:(n:name() c:comma() {(n, c)})* last:name() {
```
stroxler, line 1001: [nit] it sems like in most cases where the official grammar has `','.something+`, we've been able to use a similar form involving comma-separated stuff. Does that not apply here?

I think what we have is equivalent but the grammar diverges

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+                make_global(kw, init, last)
+            }
+
+        rule nonlocal_stmt() -> Nonlocal<'a>
+            = kw:lit("nonlocal") init:(n:name() c:comma() {(n, c)})* last:name() {
```
stroxler, line 1006: [compat] (same as for global)

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+
+        rule yield_stmt() -> Expression<'a>
+            = yield_expr()
+
+        rule assert_stmt() -> Assert<'a>
+            = kw:lit("assert") test:expression() rest:(c:comma() msg:expression() {(c, msg)})? {
+                make_assert(kw, test, rest)
+            }
+
+        rule del_stmt() -> Del<'a>
```
stroxler, line 1018: [nit / curiosity] this diverges from the grammar because we have separate rules for del_target and del_targets

Is that a historical thing, e.g. we wrap one in a tuple and not the other so it's easier to have two handler actions?

bgw:
> the parser won't throw errors if you use Python 3.9 syntax in a 3.6 codebase

How does this work for bits of Python grammar that were changed in backwards-incompatible ways? In older versions of Python, it was valid to use `async` and `await` as variable names. In 3.7+, it's no longer valid. However, if you enable the `async`/`await` tokenizer hacks on 3.7+ code, you might run into issues with `async` generator expressions (https://bugs.python.org/issue31708).

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+                make_import_from_as_names(items.0, items.1)
+            }
+
+        rule import_from_as_name() -> ImportAlias<'a>
+            = n:name() asname:(kw:lit("as") z:name() {(kw, z)})? {
+                make_import_alias(NameOrAttribute::N(n), asname)
+            }
+
+        rule dotted_as_names() -> Vec<ImportAlias<'a>>
+            = init:(d:dotted_as_name() c:comma() {d.with_comma(c)})*
```
stroxler, line 1063: [nit / curiosity] another case where grammar just uses a `','.dotted_as_name+` and we don't, it seems like we could just use a comma separated rule here

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+                last:dotted_as_name() {
+                    concat(init, vec![last])
+            }
+
+        rule dotted_as_name() -> ImportAlias<'a>
+            = n:dotted_name() asname:(kw:lit("as") z:name() {(kw, z)})? {
+                make_import_alias(n, asname)
+            }
+
+        rule dotted_name() -> NameOrAttribute<'a>
```
stroxler, line 1073: [compat] this rule diverges pretty heavily from the latest grammar which uses a recursive rule instead.

I'm pretty certain they are equivalent

stroxler (approved):
I read the grammar itself (not the entire grammar.rs file, just up to line 1180 or so) end-to-end and compared to the main grammar at
https://docs.python.org/3.11/reference/grammar.html
as well as, where necessary, the full cpython PEG at
https://github.com/python/cpython/blob/main/Grammar/python.gram

Overall it looks good to me, especially in light of having test coverage.

I flagged a number of things:

- [substantive] I think I only found 2 or 3 things that might be wrong, and
   none of them seem like they ought to block merge of an off-by-default
   new parser - I'd say address them if you want or we can merge and follow
   up later

- [nit / curiousity] a few places the rules seemed awkward, or slightly out of
   sync with the python official grammar in ways that appeared unnecessary.
   I definitely may be missing something though, marked these as [nit], [curiosity]
   or [both] and it's probably fine to merge without addressing (although if there
   are reasons I'd appreciate an explanation since I'd learn something!)

- [compat] I've marked rules that are not in a great order or otherwise have
   cosmetic issues relative to the official grammar as [compat]. Feel free to
   ignore these, they shouldn't block land but I do think I might follow up
   later with a reorder because that would make checking them against
   one another easier going forward

stroxler:
> How does this work for bits of Python grammar that were changed in backwards-incompatible ways? 

Not sure - I was mostly describing my understanding of @zsol's comment about supporting all versions, but it does appear to me that the new parser is using `async` / `await` as dedicated tokens.

I wonder if a solution is to only use the new parser, when we roll it out as default, only for 3.7+? The main motivation to move to the PEG is to properly support newer python versions. If we want to support 3.6 properly with the native parser we could always put the time into that later.

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+}
+
+impl std::fmt::Display for ParseLoc {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.start_pos.fmt(f)
+    }
+}
+
+impl<'a> Parse for TokVec<'a> {
+    type PositionRepr = ParseLoc;
```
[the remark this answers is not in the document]
zsol: It's part of the trait, you need to specify an associated type that will be used in the peg implementation: https://docs.rs/peg/0.7.0/peg/trait.Parse.html

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+            / &lit("global") s:global_stmt() {SmallStatement::Global(s)}
+            / &lit("nonlocal") s:nonlocal_stmt() {SmallStatement::Nonlocal(s)}
+
+        rule assignment() -> SmallStatement<'a>
+            = a:name() col:lit(":") ann:expression()
+                rhs:(eq:lit("=") d:annotated_rhs() {(eq, d)})? {
+                    SmallStatement::AnnAssign(
+                        make_ann_assignment(AssignTargetExpression::Name(a), col, ann, rhs))
+            }
+            // TODO: there's an extra '(' single_target ')' clause here in upstream
```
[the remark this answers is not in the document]
zsol: I was surprised at this as well, but apparently it doesn't break any tests. I wasn't able to figure out what this is used for in CPython. Maybe an optimization?

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+
+        rule star_named_expression() -> Element<'a>
+            = star:lit("*") e:bitwise_or() {
+                Element::Starred(make_starred_element(star, expr_to_element(e)))
+            }
+            / e:named_expression() { expr_to_element(e) }
+
+        #[cache]
+        rule expression() -> Expression<'a>
+            = _conditional_expression()
```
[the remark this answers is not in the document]
zsol: I pulled out `_conditional_expression` because it's also used for parsing f-strings. The official grammar doesn't include a rule for f-strings, they're parsed in a fairly ad-hoc way. Arguably this implementation is "more correct" for some definition of correct.

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+            }
+            / comparison()
+
+        #[cache]
+        rule comparison() -> Expression<'a>
+            = a:bitwise_or() b:compare_op_bitwise_or_pair()+ { make_comparison(a, b) }
+            / bitwise_or()
+
+        #[cache]
+        rule compare_op_bitwise_or_pair() -> (CompOp<'a>, Expression<'a>)
```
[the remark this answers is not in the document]
zsol: I'll add a note here. This is to avoid bloating the parser cache.

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+        rule kvpair() -> (Expression<'a>, TokenRef<'a>, Expression<'a>)
+            = k:expression() colon:lit(":") v:expression() { (k, colon, v) }
+
+        rule genexp() -> GeneratorExp<'a>
+            = lpar:lpar() g:_bare_genexp() rpar:rpar() {
+                g.with_parens(lpar, rpar)
+            }
+
+        rule _bare_genexp() -> GeneratorExp<'a>
+            = elt:named_expression() comp:for_if_clauses() {
```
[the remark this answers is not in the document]
zsol: Same reason as `_conditional_expression` above. It's used for f-string parsing.

native/libcst/src/parser/grammar.rs
zsol: The official grammar has the luxury of discarding all those separators. We need to refer to them later, which is why this is a bit more complex. Eventually I cooked up the `separated` and `separated_trailer` helper meta-rules, but
1. I'm not sure they improve readability
2. I was too lazy to go back and refactor every occurrence of this pattern. This is tech debt TODO

native/libcst/src/parser/grammar.rs
zsol: I forget why this was necessary. Maybe it wasn't? Not sure.

native/libcst/src/parser/grammar.rs
zsol: Yeah this change is because I needed to be 100% backwards compatible with the pure python parser, which produces a tuple in one case and not the other.

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+        rule _comp_if() -> CompIf<'a>
+            = kw:lit("if") cond:disjunction() {
+                make_comp_if(kw, cond)
+            }
+
+        rule arguments() -> Vec<Arg<'a>>
+            = a:args() trail:comma()? &lit(")") {add_arguments_trailing_comma(a, trail)}
+
+        rule args() -> Vec<Arg<'a>>
+            = first:_posarg()
```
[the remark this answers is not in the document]
zsol: Yes, this is just for convenience.

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+            / slices:separated_trailer(<slice()>, <comma()>) {
+                make_slices(slices.0, slices.1, slices.2)
+            }
+
+        rule slice() -> BaseSlice<'a>
+            = l:expression()? col:lit(":") u:expression()?
+                rest:(c:lit(":") s:expression()? {(c, s)})? {
+                    make_slice(l, col, u, rest)
+            }
+            / v:expression() { make_index(v) }
```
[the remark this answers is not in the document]
zsol: Yeah that's a 3.10 thing. FWIW I accidentally checked in the grammar I based the parser on :) https://github.com/zsol/LibCST/blob/parser/native/libcst/Grammar

native/libcst/src/parser/grammar.rs
```diff
@@ -0,0 +1,2930 @@
...
+//     }
+// }
+
+parser! {
+    pub grammar python<'a>(input: &'a str) for TokVec<'a> {
+        pub rule file(encoding: Option<&str>) -> Module<'a>
+            = traced(<_file(encoding.unwrap_or_else(|| "utf-8"))>)
+
+        pub rule expression_input() -> Expression<'a>
+            = traced(<e:star_expressions() tok(NL, "NEWLINE") tok(EndMarker, "EOF") {e}>)
```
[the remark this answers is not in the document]
zsol: This is to preserve LibCST behavior: https://github.com/Instagram/LibCST/blob/45c0d96c06a8f20e330e6afd671ff7145a3bab8e/libcst/_parser/conversions/expression.py#L157

native/libcst/src/parser/grammar.rs
zsol: I remember being confused about this, but I don't remember if I found how this works in CPython. Maybe I just gave up and implemented it **MY WAY** :stuck_out_tongue_closed_eyes:

jreese:
> No support for limiting the parser to a certain Python version. The parser will accept input that's valid in any of Python 3.9, 3.8, 3.7, or 3.6 at the moment.

@zsol can you confirm if you intend to add support for this functionality before the Rust parser becomes enabled by default? ie, have separate grammar for each major release of CPython, and ensure strict parsing when given a target version? 

There is a lot of work that Tim had in-flight for supporting older releases of CPython, potentially back to 2.x versions, that I don't want to lose, as well as the functionality in guaranteeing correctness and safety when parsing/modifying/dumping trees for production code. Eg, when writing codemods, we want to know that if we ask to parse with 3.7 code, that we're not getting trees/nodes/grammar that are only available in 3.8 or 3.9, and vice versa.

zsol:
Yeah this feature won't be lost one way or another. Having said that, I
don't think supporting python 2 grammars is very valuable at this point, as
I commented on Tim's issue recently

PR #568, opened 2021-12-23
Implement PEP-634 - Match statement

zsol:
This PR implements the match statement in LibCST.

- [x] Python CST nodes to represent the new syntax
- [x] Code generation from Python CST nodes
- [x] Validation for Python CST nodes
- [x] Documentation for Python CST Nodes
- [x] Extend the Rust grammar to parse new syntax
- [x] Rust CST nodes to represent the new syntax
- [x] Code generation from Rust CST nodes
- [x] Round trip tests
- [x] Unit tests
- [x] Generate new matchers and types


stroxler:
@zsol did you want me to review this as well?

zsol:
Sure, if you have time!

tushar-deepsource:
Hey! First of all, amazing work, pulling this off in such a short time.

I was testing this branch on my own test cases, and I've found a really peculiar bug:

```python
# pylint: for some reason, having a comment starting with `pylint: ` is important for this bug to show up

match x:
    case y.z:
        pass
```

This is what happens when I try to parse this:

```python
[...]
E           libcst._exceptions.ParserSyntaxError: Syntax Error @ 4:14.
E           parser error: error at 4:13: expected one of (, .
E           
E               case y.z:
E                        ^

venv3/lib/python3.10/site-packages/libcst/_parser/entrypoints.py:55: ParserSyntaxError
```

If I remove the comment on top, or if i change the `pylint: ` part to anything else, the error goes away.

zsol:
Hey @tushar-deepsource thanks for trying this out! Can confirm I can repro the problem, but AFAICT the issue is with the `y.z` case matcher, not the pylint comment. The error doesn't go away even if I tinker (or completely remove) the `pylint:` comment. I'll fix this

tushar-deepsource:
That seems to have fixed every edge case that I have in my codebase. Just one thing remains: something to do with trailing indent whitespace:

```pycon
Python 3.10.0
>>> import libcst
>>> code = '''\
... if x == type(y):
...     pass
...     ''' # This indent whitespace is important
>>> libcst.parse_statement(code)
  File ".../libcst/_parser/entrypoints.py", line 138, in parse_statement
    result = _parse(
  File ".../libcst/_parser/entrypoints.py", line 55, in _parse
    return parse(source_str)
libcst._exceptions.ParserSyntaxError: Syntax Error @ 3:5.
parser error: error at 3:4: expected one of (, *, +, -, ..., AWAIT, DEDENT, False, NAME, NUMBER, None, True, [, break, continue, lambda, match, not, pass, ~

    pass
       ^
>>>
```

native/libcst/tests/fixtures/malicious_match.py
```diff
@@ -0,0 +1,36 @@
...
+    case  Cls  (  0  ,  1  ,  x  =  0  ,  y  =  2  )  :  pass
+    case [x] as y: pass
+    case  [x]   as  y    :  pass
+    case (True)as x:pass
+    case Foo:pass
+    case (Foo):pass
+    case (  Foo   ) :  pass
+    case [  ( Foo  )   , ]: pass
+    case Foo|Bar|Baz : pass
+    case Foo  |    Bar |  (  Baz):  pass
```
stroxler, line 36: I think we're missing an exampleof bare open sequence pattern here, e.g.
```python
case x, y, *more: pass
```

stroxler: Do we have an example of a value_pattern rule match here?

It seems like the bug @tushar-deepsource flagged involved a value pattern with an attribute `y.z`, but I don't see that case.

native/libcst/src/parser/grammar.rs
```diff
@@ -2991,3 +3214,272 @@ fn make_named_expr<'a>(name: Name<'a>, tok: TokenRef<'a>, expr: Expression<'a>)
...
+    pat_comma: Option<Comma<'a>>,
+    mut kwds: Vec<MatchKeywordElement<'a>>,
+    kwd_comma: Option<Comma<'a>>,
+    rpar_tok: TokenRef<'a>,
+) -> MatchPattern<'a> {
+    if let Some(c) = pat_comma {
+        if let Some(el) = patterns.pop() {
+            patterns.push(el.with_comma(c));
+        }
+        // TODO: else raise error
```
stroxler, line 3450: do we need to handle these TODOs before merging? There are several of them both above and below this one

stroxler (approved):
Overall LGTM

I have a few nits on ordering that you could ignore, and I think one of the test fixtures could use more cases.

libcst/__init__.py
```diff
@@ -163,6 +163,23 @@
     ImportAlias,
     ImportFrom,
     IndentedBlock,
+    Match,
+    MatchCase,
+    MatchAs,
+    MatchClass,
```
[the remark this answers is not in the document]
zsol: Just between us: I mostly used the `ast` module as a reference instead of Python.asdl, the latter is not typed strongly enough for my taste

libcst/_nodes/statement.py
```diff
@@ -2419,3 +2432,811 @@ def _codegen_impl(
...
+    leading_lines: Sequence[EmptyLine] = ()
+
+    #: Whitespace between the ``match`` keyword and the subject.
+    whitespace_after_match: SimpleWhitespace = SimpleWhitespace.field(" ")
+
+    #: Whitespace after the subject but before the colon.
+    whitespace_before_colon: SimpleWhitespace = SimpleWhitespace.field("")
+
+    #: Any optional trailing comment and the final ``NEWLINE`` at the end of the line.
+    whitespace_after_colon: TrailingWhitespace = TrailingWhitespace.field()
```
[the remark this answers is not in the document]
zsol: It's a different class. In `TrailingWhitespace` all fields have a default value, in `SimpleWhitespace` they don't.

native/libcst/src/parser/grammar.rs
zsol: Not really. There's no good way of erroring out the parsing apart from `panic!()` at the moment, and these cases shouldn't be happening unless there's a bug in the grammar.

PR #569, opened 2021-12-23
Fix CI publish to test.pypi.org

zsol:
This should fix the failing pypi publish steps. TLDR there's a misfeature in `actions/checkout@v2` that fails to pull down tags. See https://github.com/actions/checkout/issues/290


lpetre (approved):
LGTM, thanks

PR #570, opened 2021-12-23
Fix pypi upload CI step

zsol:
This will get the built wheel to not have a local scheme, so they can be uploaded to test.pypi.org, then the upload will happen in a separate job, once all the binary wheels are built. Source dist is also built in this job.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -216,6 +216,7 @@ jobs:
       if: steps.cache.outputs.cache-hit != 'true'
       run: exit 1
     - name: Disable scmtools local scheme
+      if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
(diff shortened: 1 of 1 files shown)


lpetre (approved):

.github/workflows/build.yml
```diff
@@ -224,20 +224,48 @@ jobs:
...
+      - uses: actions/checkout@v1
+      - name: Download binary wheels
+        id: download
+        uses: actions/download-artifact@v2
+        with:
+          name: wheels
+          path: wheelhouse
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
```
zsol, line 244: This seems wrong, how did this step pass on the PR? `matrix.python-version` shouldn't be set here.

PR #571, opened 2021-12-23
implement PEP-654: except*

zsol:
## Summary

This PR adds support for the new grouped exception handling mechanism introduced by PEP-654. The only significant difference between AST and this CST implementation is a new `ExceptStarHandler` type, which differs from `ExceptHandler` in:
* requiring the `type` attribute to be always present (it can't be `None`)
* owning the whitespace between the `except` keyword and the `*`

The new syntax can only be parsed by the Rust parser (currently needing `LIBCST_PARSER_TYPE=native` environment variable).

## Test Plan

- [x] unit tests
- [x] round-trip tests


zsol:
@stroxler xmas treat for you :)

libcst/_nodes/statement.py
```diff
@@ -948,6 +1024,75 @@ def _codegen_impl(self, state: CodegenState) -> None:
                 finalbody._codegen(state)
 
 
+@add_slots
+@dataclass(frozen=True)
+class TryStar(BaseCompoundStatement):
+    """
+    A ``try`` statement with ``except*`` blocks.
```
stroxler, line 1031: github doesn't let me comment on a non-diff line 😠 

But - should we update the docstring of `Try`?

libcst/_nodes/statement.py
```diff
@@ -948,6 +1024,75 @@ def _codegen_impl(self, state: CodegenState) -> None:
...
+    def _codegen_impl(self, state: CodegenState) -> None:
+        for ll in self.leading_lines:
+            ll._codegen(state)
+        state.add_indent_tokens()
+
+        end_node = self.handlers[-1]
+        orelse = self.orelse
+        end_node = end_node if orelse is None else orelse
+        finalbody = self.finalbody
+        end_node = end_node if finalbody is None else finalbody
```
stroxler, line 1082: I can see that this is consistent with existing style, and I approve of it in terms of this one diff, we ought to respect the style elsewhere.

But do we really think overwriting `end_node` 3 times is clearer than having an if/else block that just sets it to the appropriate value? I find it confusing, and I'm pretty sure it's also less efficient to do everything in backward order like this! I'd be happy to put up a PR trying to clean this up a bit in the whole module.

libcst/matchers/__init__.py
```diff
@@ -4341,6 +4341,142 @@ class ExceptHandler(BaseMatcherNode):
     ] = DoNotCare()
 
 
+@dataclass(frozen=True, eq=False, unsafe_hash=False)
```
stroxler, line 4344: Are we worried about the matcher code? I have been putting off trying to understand how matchers work in detail (this code makes zero sense to me without reading through those docs) I can read up on it if you would like.

native/libcst/src/nodes/statement.rs
```diff
@@ -1713,6 +1715,68 @@ impl<'a> Inflate<'a> for ExceptHandler<'a> {
...
+            n.codegen(state);
+        }
+        self.whitespace_before_colon.codegen(state);
+        state.add_token(":");
+        self.body.codegen(state);
+    }
+}
+
+impl<'a> Inflate<'a> for ExceptStarHandler<'a> {
+    fn inflate(mut self, config: &Config<'a>) -> Result<Self> {
```
stroxler, line 1755: I didn't read enough of the original Rust parser PR to understand `Inflate`. What does it do, and what is `Config`? From the code it seems related to parsing, but that's what the parser does.

Is this a post-processing step after the parser, where we basically move whitespace up from child CST structs to the parent CST structs? At a skim that kind of seems to be what's going on here, e.g. reassigning whitespace from the except token to self.

native/libcst/src/parser/grammar.rs
```diff
@@ -498,6 +499,13 @@ parser! {
                     make_try(kw, b, ex, el, f)
             }
 
+        // Note: this is separate because TryStar is a different type in LibCST
+        rule try_star_stmt() -> TryStar<'a>
```
stroxler, line 503: The official grammar has this as a variant under `try_stmt`. We're sure it's worth diverging? I can see some value in it since the actual AST node variant is different, but there's a price we'll pay keeping things in sync for every rule that diverges.

stroxler (approved):
LGTM, I left a couple comments.

The big thing I skipped was the matcher code, I can take a look at that but it will require some self-training so I figured I'd let you decide whether it's worth it first.

libcst/matchers/__init__.py
stroxler: Update: I went ahead and read the matcher docs and I think this makes sense to me, I guess it's generated code and therefore probably correct.

I don't quite follow the OneOf / AllOf business, seems a bit strange to me but that shouldn't block landing

libcst/_nodes/statement.py
zsol: I also found it confusing, feel free to put up that PR!

native/libcst/src/nodes/statement.rs
zsol: > Is this a post-processing step after the parser, where we basically move whitespace up from child CST structs to the parent CST structs

Yes, exactly. `Config` is all the data whitespace parsing needs (e.g. default indentation)

native/libcst/src/parser/grammar.rs
zsol: Not 100% sure it's worth it, I was a bit lazy and wanted to avoid an additional enum layer just so we can have a single rule.

stroxler: Oh are the rules tied more tightly to types in the rust PEG library than in C?

In CPython it's a single PEG variant, but two different AST types still

PR #573, opened 2021-12-29
Make sure dedents are emitted for inputs without trailing newlines

### LINKED ISSUE #572: Rust parser fails on trailing indentation
Repro by @tushar-deepsource:

```pycon
Python 3.10.0
>>> import libcst
>>> code = '''\
... if x == type(y):
...     pass
...     ''' # This indent whitespace is important
>>> libcst.parse_statement(code)
  File ".../libcst/_parser/entrypoints.py", line 138, in parse_statement
    result = _parse(
  File ".../libcst/_parser/entrypoints.py", line 55, in _parse
    return parse(source_str)
libcst._exceptions.ParserSyntaxError: Syntax Error @ 3:5.
parser error: error at 3:4: expected one of (, *, +, -, ..., AWAIT, DEDENT, False, NAME, NUMBER, None, True, [, break, continue, lambda, match, not, pass, ~

    pass
       ^
>>>
```

_Originally posted by @tushar-deepsource in https://github.com/Instagram/LibCST/issues/568#issuecomment-1002574953_

zsol:
## Summary

The parser can get input that is not newline terminated. Normally any previous indent blocks are closed by the tokenizer (by emitting a `Dedent` token) on a final newline, or when a fake newline was emitted (to make sure the Python grammar still works). This PR fixes the case when there's no fake newline (because the last token is a real newline), however there are still more bytes before the last newline and EOF. This must mean comments and whitespace between the last newline and EOF, for example:
```
if 1:
  if 2:
    pass
    # foo but no newline
```

or just with an indent and no comment at the last line:

```
if 1:
  if 2:
    pass
    
```

Fixes #572.

## Test Plan

Added unit and roundtrip tests.


tusharsadhwani:
Hey @zsol

Is it possible for you to release an alpha/beta version of 0.3.24 after this PR is merged? It'd help me out with my deploys. Thanks

zsol:
Yeah I'm planning to do a prerelease today or tomorrow, but I'd need someone with repo write permission to review this PR.

zsol:
@tusharsadhwani here you go: https://pypi.org/project/libcst/0.4.0a1/

tushar-deepsource:
@zsol 0.4.0a1 is working splendidly so far. Thanks!

native/libcst/src/tokenizer/core/mod.rs
```diff
@@ -342,7 +342,20 @@ impl<'t> TokState<'t> {
...
+                        let hanging_indents = self.indent_stack.len() as i32;
+                        if self.pending_indents == 0 && hanging_indents != 0 {
+                            // We've reached EOF but there are still pending indents not
+                            // accounted for. Flush them out.
+                            self.pending_indents = -hanging_indents;
+                            self.indent_stack.clear();
+                            self.alt_indent_stack.clear();
+                            self.missing_nl_before_eof = false;
+                        }
+                        if let Some(t) = self.process_pending_indents() {
```
lpetre, line 354: This seems like it would only generate a single dedent, how does it handle the case where there are multiple levels of indentation to flush?

zsol: This function is getting called repeatedly and should only produce one token for each invocation. So in effect we'll be calling `process_pending_indents()` repeatedly until it returns `None`, at which point we're done with the input, so we fall down to the else branch on line 356.

lpetre: Ok, that makes more sense. I see why `hanging_indents` is a good name, and how this works conceptually. As long as `pending_indents` is a positive integer we will emit the dedent, and when it is zero we emit the end marker.

lpetre (approved):

PR #579, opened 2022-01-05
Fix all type errors

zsol:
This PR
* upgrades pyre to 0.9.9 which brings a new type checking backend
* fixes CI to run pyre on the working copy instead of the latest libcst release 🤡
* fixes some amount of type errors that 0.9.9 detects now
* fixes some amount of type errors that have crept in since type checking was dysfunctional in CI 🤡
* suppresses the rest of the type errors
* adds stubs for the rust parser


stroxler (approved):
Looks good to me.

Some of the complaints about unbound type variables actually seem wrong to me, I can make a task to look at that from the pyre side

PR #581, opened 2022-01-05
only run CI on pushes to main and pull requests

zsol:
## Summary

Don't waste capacity by running checks twice on pull requests originating from a branch on the same repo.


lpetre (approved):
Thanks

PR #584, opened 2022-01-05
[WIP] Support Parenthesized With Statements

stroxler:
Creating an early draft PR for two reasons:
- I want to see CI results, I'm too lazy to run all the checks locally
- in case someone wants to look over the parts that are finished - at this point I've added parentheses to the python node, but not yet started any of the rust-side work.

### DIFF
--- a/libcst/_nodes/statement.py
+++ b/libcst/_nodes/statement.py
@@ -57,6 +57,7 @@ from libcst._nodes.op import (
 from libcst._nodes.whitespace import (
     BaseParenthesizableWhitespace,
     EmptyLine,
+    ParenthesizedWhitespace,
     SimpleWhitespace,
     TrailingWhitespace,
 )
@@ -2050,7 +2051,7 @@ class With(BaseCompoundStatement):
             and self.items[-1].comma != MaybeSentinel.DEFAULT
         ):
             raise CSTValidationError(
-                "The last WithItem in a With cannot have a trailing comma."
+                "The last WithItem in an unparenthesized With cannot have a trailing comma."
             )
         if self.whitespace_after_with.empty and not (
             isinstance(self.lpar, LeftParen)
@@ -2085,6 +2086,17 @@ class With(BaseCompoundStatement):
             ll._codegen(state)
         state.add_indent_tokens()
 
+        needs_paren = False
+        for item in self.items:
+            comma = item.comma
+            if isinstance(comma, Comma):
+                if isinstance(
+                    comma.whitespace_after,
+                    (EmptyLine, TrailingWhitespace, ParenthesizedWhitespace),
+                ):
+                    needs_paren = True
+                    break
+
         with state.record_syntactic_position(self, end_node=self.body):
             asynchronous = self.asynchronous
             if asynchronous is not None:
@@ -2094,12 +2106,16 @@ class With(BaseCompoundStatement):
             lpar = self.lpar
             if isinstance(lpar, LeftParen):
                 lpar._codegen(state)
+            elif needs_paren:
+                state.add_token("(")
             last_item = len(self.items) - 1
             for i, item in enumerate(self.items):
                 item._codegen(state, default_comma=(i != last_item))
             rpar = self.rpar
             if isinstance(rpar, RightParen):
                 rpar._codegen(state)
+            elif needs_paren:
+                state.add_token(")")
             self.whitespace_before_colon._codegen(state)
             state.add_token(":")
             self.body._codegen(state)
--- a/libcst/_nodes/statement.py
+++ b/libcst/_nodes/statement.py
@@ -2049,9 +2049,10 @@ class With(BaseCompoundStatement):
             raise CSTValidationError(
                 "The last WithItem in a With cannot have a trailing comma."
             )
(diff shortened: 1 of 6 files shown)


libcst/_nodes/statement.py
```diff
@@ -2017,13 +2017,30 @@ class With(BaseCompoundStatement):
     #: Sequence of empty lines appearing before this with statement.
     leading_lines: Sequence[EmptyLine] = ()
 
+    #: Optional open parenthesis for multi-line with bindings
+    lpar: Union[LeftParen, MaybeSentinel] = MaybeSentinel.DEFAULT
+
+    #: Optional close parenthesis for multi-line with bindings
+    rpar: Union[RightParen, MaybeSentinel] = MaybeSentinel.DEFAULT
```
zsol, line 2024: If we use `MaybeSentinel` here instead of `Optional[LeftParen]`, we need to be very clever during code generation.
The main difference between these options is that `Optional` signals that the caller is responsible for determining if a `LeftParen` is needed or not (and `With` may or may not validate this). If we go with `MaybeSentinel`, then `With` needs to make sure the parenthesis are emitted when necessary (i.e. if the `with` statement's header spans multiple lines).

The second option is a better UX overall if we can come up with the cleverness to detect when the parens are necessary (which would be nice to do in `_validate_parens` even if we go with `Optional`).

libcst/_nodes/tests/test_with.py
```diff
@@ -178,6 +180,18 @@ class WithTest(CSTNodeTest):
                 "parser": parse_statement,
                 "expected_position": CodeRange((1, 0), (1, 36)),
             },
+            # Parenthesized
```
zsol, line 183: Don't forget to add a test case with multiple `WithItem`s spanning multiple lines

zsol:
Oh also let's add a test case that covers `_validate_parens()`

libcst/_nodes/statement.py
stroxler: Makes sense.

It seems like the rust side doesn't care which we do, so I think I'll try to get the parser working first. I think it will be easier to figure out the right approach once I can generate the CST from an example.

zsol: >  the rust side doesn't care

Absolutely. Ergonomy of the rust API is a tertiary priority (if that) 😄

native/libcst/src/nodes/statement.rs
```diff
@@ -2027,7 +2044,9 @@ impl<'a> Inflate<'a> for With<'a> {
 
         self.whitespace_after_with =
             parse_simple_whitespace(config, &mut (*self.with_tok).whitespace_after.borrow_mut())?;
+        self.lpar = self.lpar.map(|lpar| lpar.inflate(config)).transpose()?;
         self.items = self.items.inflate(config)?;
+        self.rpar = self.rpar.map(|rpar| rpar.inflate(config)).transpose()?;
```
zsol, line 2049: For other (parenthesized) comma separated lists, `rpar` owns the whitespace after a trailing comma. This implementation currently diverges because the item's `inflate` will consume the whitespace before `rpar` gets inflated. Check out `Tuple`'s `Inflate` implementation and the `inflate_element` associated function on `Element`

stroxler: I think I have this working correctly now, although it still needs a test case.

Although I do wish there were some docs on how the lower-level components of the inflation work - it seems like the 'config' has a mutable store of whitespace and inflate kind of vacuums up anything adjacent to a token that hasn't already been claimed? I'm hoping to read that code in more detail tomorrow, today I just focused on making it run.

libcst/_nodes/statement.py
stroxler: Hoping to do this tomorrow, for both the rust and python codegen steps.

I *think* that I probably only need to look for whitespace on the commas of the WithItem entries in `items` to make this work properly.

native/libcst/src/nodes/statement.rs
zsol: Yes, all of this code is in dire need of some documentation.

zsol:
LGTM!

stroxler:
Updates:
- patched an error in the grammar - I didn't have trialing comma support properly as of yesterday
- included trailing comma for items in test case (this verifies the inflate behavior)
- adjusted validation code for trailing comma - it's only invalid when there aren't parentheses
- added tests for parenthesis balancing validation
- added MaybeSentinel handling and a test

I believe this is now ready. The rust codegen api doesn't currently attempt to infer parentheses, I'll try to add that in the next week but I'd rather get the important change out first - since parentheses are just `Option`s in rust and there's no actual client code yet it doesn't seem nearly as important.

stroxler:
this will also close #584

zsol (approved):
code looks good! Would you mind adding a couple of test cases in `native/libcst/tests/fixtures/with_wickedness.py` before merging please?

tushar-deepsource:
@zsol would you mind making another alpha release with this patch?

zsol:
I'm planning to release 0.4.0 sometime this week if Life doesn't intervene

PR #585, opened 2022-01-06
Support relative imports in AddImportsVisitor.

### LINKED ISSUE #578: support relative imports in AddImportsVisitor
The definition of an `import` is currently `(module, obj, asname)`, leaving no way to handle relative imports in codemods that use the visitor.

martindemello:
* Adds an Import dataclass to represent a single imported object
* Refactors AddImportsVisitor to pass around Import objects
* Separates out the main logic in get_absolute_module_for_import so that
  it can be used to resolve relative module names outside of a cst.Import
  node
* Resolves relative module names in AddImportsVisitor if we have a
  current module name set.

Fixes #578

### DIFF
--- a/libcst/codemod/visitors/_imports.py
+++ b/libcst/codemod/visitors/_imports.py
@@ -19,7 +19,9 @@ class ImportItem:
     relative: int = 0
 
     def __post_init__(self) -> None:
-        if self.module_name.startswith("."):
+        if self.module_name is None:
+            object.__setattr__(self, "module_name", "")
+        elif self.module_name.startswith("."):
             mod = self.module_name.lstrip(".")
             rel = self.relative + len(self.module_name) - len(mod)
             object.__setattr__(self, "module_name", mod)
@@ -31,7 +33,11 @@ class ImportItem:
 
     def resolve_relative(self, base_module: Optional[str]) -> "ImportItem":
         """Return an ImportItem with an absolute module name if possible."""
+        mod = self
+        # `import ..a` -> `from .. import a`
+        if mod.relative and mod.obj is None:
+            mod = replace(mod, module_name="", obj=mod.module_name)
         if base_module is None:
-            return self
-        m = get_absolute_module(base_module, self.module_name, self.relative)
-        return self if m is None else replace(self, module_name=m, relative=0)
+            return mod
+        m = get_absolute_module(base_module, mod.module_name or None, self.relative)
+        return mod if m is None else replace(mod, module_name=m, relative=0)
--- a/libcst/codemod/visitors/_imports.py
+++ b/libcst/codemod/visitors/_imports.py
@@ -1,0 +1,37 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from dataclasses import dataclass, replace
+from typing import Optional
+
+from libcst.helpers import get_absolute_module
+
+
+@dataclass(frozen=True)
+class Import:
+    """Representation of individual import items for codemods."""
+
+    module_name: str
+    obj: Optional[str] = None
+    alias: Optional[str] = None
+    relative: int = 0
+
+    def __post_init__(self):
+        if self.module_name.startswith("."):
+            mod = self.module_name.lstrip(".")
+            rel = self.relative + len(self.module_name) - len(mod)
+            object.__setattr__(self, "module_name", mod)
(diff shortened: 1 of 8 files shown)


zsol (changes requested):
Overall nice!

How about a test case with a `from .. import E as foo` style import?

libcst/codemod/visitors/_imports.py
```diff
@@ -0,0 +1,37 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from dataclasses import dataclass, replace
+from typing import Optional
+
+from libcst.helpers import get_absolute_module
+
+
+@dataclass(frozen=True)
+class Import:
```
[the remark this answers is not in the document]
martindemello: good point. done.

libcst/codemod/visitors/_imports.py
```diff
@@ -0,0 +1,37 @@
...
+    """Representation of individual import items for codemods."""
+
+    module_name: str
+    obj: Optional[str] = None
+    alias: Optional[str] = None
+    relative: int = 0
+
+    def __post_init__(self):
```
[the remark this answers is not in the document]
martindemello: done

libcst/codemod/visitors/_imports.py
```diff
@@ -0,0 +1,37 @@
...
+        if self.module_name.startswith("."):
+            mod = self.module_name.lstrip(".")
+            rel = self.relative + len(self.module_name) - len(mod)
+            object.__setattr__(self, "module_name", mod)
+            object.__setattr__(self, "relative", rel)
+
+    @property
+    def module(self) -> str:
+        return "." * self.relative + self.module_name
```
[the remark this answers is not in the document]
martindemello: i usually prefer the f-string version too but it looks more cluttered in this case.

libcst/codemod/visitors/tests/test_add_imports.py
```diff
@@ -481,11 +481,11 @@ def bar() -> int:
                 return 5
         """
 
-        self.assertCodemod(before, after, [("a.b.c", "D", None)])
+        self.assertCodemod(before, after, [Import("a.b.c", "D", None)])
 
     def test_add_import_preserve_doctring_multiples(self) -> None:
         """
-        Should preserve any doctring if adding to the beginning.
+        Should preserve any doctring if addImport(ng to the beginning.
```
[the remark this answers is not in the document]
martindemello: hit the wrong key in vim and didn't notice (:

libcst/codemod/visitors/_imports.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def resolve_relative(self, base_module: Optional[str]) -> "ImportItem":
+        """Return an ImportItem with an absolute module name if possible."""
+        mod = self
+        # `import ..a` -> `from .. import a`
+        if mod.relative and mod.obj is None:
+            mod = replace(mod, module_name="", obj=mod.module_name)
```
zsol, line 39: Unfortunately `replace`'s first argument is also called `obj`, so this won't work (as shown by the unit tests)

martindemello: huh. any idea why the local tests passed? (perhaps because i only ran them on 3.9?)

zsol: Yeah this only seems to be a problem with 3.6 because we use the dataclasses package from pypi, where `obj` is not a positional-only argument.

zsol: See also https://bugs.python.org/issue37163

libcst/codemod/visitors/_imports.py
```diff
@@ -0,0 +1,49 @@
...
+        mod = self
+        # `import ..a` -> `from .. import a`
+        if mod.relative and mod.obj is None:
+            # py3.6 dataclass backport does not like `obj` as an arg to replace()
+            mod = ImportItem(
+                module_name="",
+                obj=mod.module_name,
+                alias=mod.alias,
+                relative=mod.relative,
+            )
```
zsol, line 45: I don't really like this because we'll have to remember to update this call when we add/remove fields to/from `ImportItem`. It'd be easier to just rename the `obj` field to `obj_name`

martindemello: yeah, good point

martindemello: done

zsol (approved):
Nice, thanks! I'll merge if CI is green.

PR #586, opened 2022-01-06
Remove setup matrix

lpetre:
## Summary

I was porting some of this to Fixit and I noticed that the setup matrix takes a ton of time and we're idle waiting for it to finish, when we could be running downstream jobs.  So just populate the cache inside each job

## Test Plan
CI


zsol (approved):
LGTM

PR #588, opened 2022-01-06
Remove tox references

lpetre:
## Summary
We're no longer using tox in the CI and so the tox.ini is getting stale. Remove it

## Test Plan
CI


stroxler:
This is fine by me, but I guess an open question is whether having tox available for local testing of the matrix is worthwhile.

I've been making draft PRs to make github do it for me, and it's not terribly hard to spin up a new pyenv virtualenv with a pinned python version, so I have no strong opinions.

lpetre:
> This is fine by me, but I guess an open question is whether having tox available for local testing of the matrix is worthwhile.
> 
> I've been making draft PRs to make github do it for me, and it's not terribly hard to spin up a new pyenv virtualenv with a pinned python version, so I have no strong opinions.

I know that @zsol and I both use pyenv to manage different python versions, and then separate virtualenvs for each python version. 

My concern is that keeping both the github actions and tox in sync is extra overhead, and using tox in the github actions feels like extra overhead and pain re: caching etc

stroxler (approved):
LGTM

PR #589, opened 2022-01-06
Use pyre site-package feature

lpetre:
## Summary

If you set a search path (for stubs etc) then pyre won't automatically include your site-packages, and you need to jump through big hoops to generate a .pyre_configuration (see https://github.com/Instagram/LibCST/blob/main/.github/workflows/build.yml#L84-L92)

Move to the new site-package feature in pyre to be explicit about the search path.

## Test Plan
CI


zsol:
Mind updating the readme too? And also remove `.pyre_configuration.example`

zsol (approved):

PR #590, opened 2022-01-07
Fix variable name in lambda

stroxler:
The compiler doesn't care, but it confused me for a few seconds when I saw this.


zsol (approved):
oops :)

PR #591, opened 2022-01-07
Use precise signature matching when inserting function type annotations

martindemello:
Matches signatures on shape, and on compatibility of existing annotations if `overwrite_existing_annotations` is False.

Matches posargs with different names if `strict_posargs_matching` is set to False

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -43,6 +43,29 @@ def _get_import_names(imports: Sequence[Union[cst.Import, cst.ImportFrom]]) -> S
     return import_names
 
 
+def _is_set(x) -> bool:
+    return x is not None and x != cst.MaybeSentinel.DEFAULT
+
+
+@dataclass(frozen=True)
+class FunctionKey:
+    name: str
+    pos: int
+    kwonly: str
+    posonly: int
+    star_arg: bool
+    star_kwarg: bool
+
+    @classmethod
+    def make(cls, name: str, params: cst.Parameters):
+        pos = len(params.params)
+        kwonly = ",".join(sorted(x.name.value for x in params.kwonly_params))
+        posonly = len(params.posonly_params)
+        star_arg = _is_set(params.star_arg)
+        star_kwarg = _is_set(params.star_kwarg)
+        return cls(name, pos, kwonly, posonly, star_arg, star_kwarg)
+
+
 @dataclass(frozen=True)
 class FunctionAnnotation:
     parameters: cst.Parameters
@@ -99,7 +122,9 @@ class TypeCollector(cst.CSTVisitor):
             self._handle_Annotation(annotation=returns) if returns is not None else None
         )
         parameter_annotations = self._handle_Parameters(node.params)
-        self.function_annotations[".".join(self.qualifier)] = FunctionAnnotation(
+        name = ".".join(self.qualifier)
+        key = FunctionKey.make(name, node.params)
+        self.function_annotations[key] = FunctionAnnotation(
             parameters=parameter_annotations, returns=return_annotation
         )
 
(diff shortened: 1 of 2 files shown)


libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -593,6 +637,60 @@ def _insert_empty_line(
             *statements[1:],
         ]
 
+    def _match_signatures(  # noqa: C901: Too complex
```
stroxler, line 640: I wonder if we might want to separate out structural compatiblity from type annotation compatibility.

I'll explain more in the tests

libcst/codemod/visitors/tests/test_apply_type_annotations.py
```diff
@@ -298,7 +298,7 @@ def foo(x = 1) -> Iterable[Any]:
                 """
                 from typing import Iterable, Any
 
-                def foo(x: int = 1) -> Iterable[Any]:
+                def foo(x = 1) -> Iterable[Any]:
```
stroxler, line 301: The three changes above are not really desirable from the standpoint of `MonkeyType`, which is the main client I know well: in that setting by default we don't want to overwrite existing type annotations, but we do want to guess at any types that lack existing annotation.

`pyre infer` is similar, although I think in most cases it wouldn't produce incompatible types the way MonkeyType would because the type inference algorithm works differently from sampling runtime types.

This is why I think it might be worth separating out the structural aspects of signature matching - which by default we probably do want - from the type-annotation-compatibility aspects which by default I'm guessing many clients won't want.

stroxler: To add more color to my comment

**Why the type-annotation compatibility might not be a great default**

for a function like
```
def f(x) -> Iterable[int]:
    return [x]
```
if `x` is always an int then the return type annotation is completely correct, but MonkeyType is going to infer something like
```
def f(x: int) -> list[int]: ...
```
which will fail the compatibility check and cause us to not apply the (correct and desirable) annotation of `int` for `x.`

The underlying issue is that `Iterable[int]` is a perfectly valid annotation for `list[int]` values, but libcst isn't type-system aware and therefore doesn't know this, and there's not an easy way to change that.

**Why structural compatibility probably is a good default**

On the other hand a structural mismatch is different and much less controversial, I don't think we'd ever want to partially annotate a function
```
def f(x, y):
    return 5
```
based on a stub file containing
```
def f(x: str) -> None: ...
```

stroxler (changes requested):
Nice! The functionality looks great to me, thanks @martindemello

But I think the check that all existing type annotations are the same is probably less useful for most clients than the structural checks, and should be off be default.

Details on my qualms (thinking about MonkeyType is instructive) are in inline comments


I'd be fine with merging this as-is if we change the default to be off, and then making a separate PR to separate out the structural versus type-compatiblity restrictions

libcst/codemod/visitors/tests/test_apply_type_annotations.py
martindemello: yes, that's a very good point - annotations may be compatible but not identical, and we don't have enough information about the type system here to tell that. i think i can do this with a two-pass approach - collect pairs of overlapping type annotations, let the calling code pass them to another library to determine which types are compatible, and then run the apply type annotation pass with a mapping of compatible types. it might end up being beyond the scope of this visitor if it clutters the code too much for a low-use feature, but some sort of subclass solution should be possible. for now i'll put the identical-type matching behind a flag and turn it off by default.

martindemello: btw, i did think of the false negative issue, but it seemed more important to not merge incompatible signatures, especially given that pyi files allow overloads. i guess that's a philosophical difference, i can definitely see your point about clients wanting maximal insertion of annotations by default. i haven't yet come up with a good way to handle overloads either, though i would like to at some point.

stroxler:
This looks good to me

stroxler (approved):
This looks good to me, thanks for the changes!

stroxler:
I'll hold off for now on merging in case you're planning to make more changes - @martindemello let me know when this is ready.

martindemello:
thanks, it should be good to merge!

PR #593, opened 2022-01-08
Add a `Building` subsection to the `Developing` docs

stroxler:
## Summary

Our README said to `pip install -e .` for type checking, but at this point that won't work
unless `cargo` is on the `PATH`, so I thought we should add some instructions.

## Test Plan

I verified that having cargo and python should be enough by building LibCST in a
docker container, see https://github.com/stroxler/dockerfiles/tree/main/LibCST-minimal

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -172,21 +172,15 @@ this:
 
 See the `unittest documentation <https://docs.python.org/3/library/unittest.html>`_
 for more examples of how to run tests.
-
+    
 Building
 ~~~~~~~~
 
-In order to build LibCST, which has a native parser module, you will
-need to have the Rust build tool ``cargo`` on your path. You can
+In order to build LibCST, which includes a native parser module, you
+will need to have the Rust build tool ``cargo`` on your path. You can
 usually install ``cargo`` using your system package manager, but the
 most popular way to install cargo is using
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -173,26 +173,48 @@ this:
...
+~~~~~~~~
 
-To verify types for the library, do the following in the root:
+In order to build LibCST, which has a native parser module, you will
+need to have the Rust build tool ``cargo`` on your path. You can
+usually install ``cargo`` using your system package manager, but the
+most popular way to install cargo is using
+`rustup <https://rustup.rs/>`_.
+
+To generate documents, do the following in the root:
```
zsol, line 185: Looks like the docs section moved up here :)

README.rst
```diff
@@ -173,26 +173,48 @@ this:
...
+
+To build just the native parser, do the following from the ``native``
+directory:
+
+.. code-block:: shell
 
-*Note*: You may need to remove installed LibCST and install from the source code prior to type checking:
+    cargo build
+
+To build the ``libcst.native`` module, run this from root:
```
zsol, line 198:
```suggestion
To build and install the ``libcst.native`` module, run this from root:
```

zsol (approved):

PR #594, opened 2022-01-11
Codemod for PEP 484 Assign w / type comments -> PEP 526 AnnAssign

stroxler:
Summary:

This codemod is intended to eventually handle all type comments from
PEP 484. This is a partial implementation specifically handling
assignment type comments, which as of PEP 526 are better dealt
with using AnnAssign nodes.

There is more work to do because there are two other kinds of
comments to support: function heading comments and function parameter
inline comments. But the PEP 526 functionality is complete so I feel
like it's worth having a PR / CI signals / code review at this stage.

Implementation Note:

It's a little weird that inside the transform we serialize small bits of the CST back
to code in order to use the `ast` module, but that's the most robust way to get
type comments and guaranteed to be PEP 484 compliant. Perf isn't an issue here
and I discussed with @zsol, we agreed this was the best way to go because
implementing our own type comment extraction would be a pain for no real benefit.

The resulting implementation is simple, but it won't work on Python < 3.8. This seems
fine for now; there's a `type_ast` library and if anyone is motivated to make this
work on 3.6 or 3.7 they could use it - I left a note in the code about where to look.

Test Plan:

```
python -m unittest libcst.codemod.commands.tests.test_convert_type_comments
```

### DIFF
--- a/libcst/codemod/commands/convert_type_comments.py
+++ b/libcst/codemod/commands/convert_type_comments.py
@@ -7,10 +7,11 @@ import ast
 import builtins
 import functools
 import re
+import sys
 from typing import Optional, Pattern, Set, Union
 
 import libcst as cst
-from libcst.codemod import VisitorBasedCodemodCommand
+from libcst.codemod import CodemodContext, VisitorBasedCodemodCommand
 
 
 @functools.lru_cache()
@@ -24,19 +25,7 @@ def _code_for_node(node: cst.CSTNode) -> str:
 
 def _ast_for_node(node: cst.CSTNode) -> ast.Module:
     code = _code_for_node(node)
-    try:
-        return ast.parse(code, type_comments=True)
-    except TypeError:
-        # The ast module did not get `type_comments` until Python 3.7.
-        # In 3.6, we should error than silently running a nonsense codemod.
-        #
-        # NOTE: it is possible to use the typed_ast library for 3.6, but this is
-        # not a high priority right now. See, e.g., the mypy.fastparse module.
-        raise NotImplementedError(
-            "You are trying to run ConvertTypeComments on a "
-            "python version without type comment support. Please "
-            "try using python 3.8 to run your codemod"
-        )
+    return ast.parse(code, type_comments=True)
 
 
 def _assign_type_comment(node: cst.SimpleStatementLine) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


stroxler:
The cancelled check is showing as red, but it's not actually a problem

libcst/codemod/commands/convert_type_comments.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def _is_type_comment(self, comment: Optional[cst.Comment]) -> bool:
+        return (
+            comment is not None
+            and self.TYPE_COMMENT_REGEX.match(comment.value) is not None
+            and not comment.value.startswith(self.TYPE_IGNORE_PREFIX)
+        )
+
+    def _should_strip_comment(self, comment: Optional[cst.Comment]) -> bool:
+        return self._should_strip_type_comments and self._is_type_comment(comment)
```
grievejia, line 93: This seems to be an unreliable way of detecting whether a comment is a type comment or not. Wouldn't it be possible to detect whether to strip comments based on whether the containing ast node has `type_comments` attribute set instead?

libcst/codemod/commands/convert_type_comments.py
```diff
@@ -0,0 +1,163 @@
...
+        Convert any SimpleStatementLine containing an Assign with a
+        type comment into a one that uses a PEP 526 AnnAssign.
+        """
+        # manage the state used for determining comments to delete
+        self._should_strip_type_comments = False
+        # determine whether to apply an annotation
+        body = updated_node.body
+        if len(body) != 1:
+            return updated_node
+        statement = body[0]
```
grievejia, line 131: Not sure if this is not going to work if you have multiple statements on the same line. E.g.
```
foo(); x = 42  # type: int
```
The type comment should apply to the last stmt, not the first one.

libcst/codemod/commands/convert_type_comments.py
```diff
@@ -0,0 +1,163 @@
...
+    def _should_strip_comment(self, comment: Optional[cst.Comment]) -> bool:
+        return self._should_strip_type_comments and self._is_type_comment(comment)
+
+    def _convert_Assign(
+        self,
+        assign: cst.Assign,
+        type_comment: str,
+    ) -> Union[cst.AnnAssign, cst.Assign]:
+        if len(assign.targets) != 1:
+            return assign
```
grievejia, line 101: I presume you have future work planned for this case? According to PEP 484 and `mypy`, all of the followings are possible:
```
x, y, z = [], [], []  # type: List[int], List[int], List[str]
x, y, z = [], [], []  # type: (List[int], List[int], List[str])
a, b, *c = range(5)   # type: float, float, List[float]
a, b = 1, 2  # type: Tuple[int, int]
```

libcst/codemod/commands/convert_type_comments.py
stroxler: Hmm... I don't think it's going to be possible to do this in general using a visitor pattern. I could try extracting type information in a `visit_` method on the parent, but that won't work later on when I handle more cases because the same problem is going to come up with function defs, with statements, and for statements.

I guess the answer is going to be to transform whitespace manually, which I was hoping to avoid because in the FunctionDef case it's not always in the same place. But that will probably be much more reliable even though it's ugly.

libcst/codemod/commands/convert_type_comments.py
stroxler: Unclear. Supporting 100% of PEP 484 might not be worth the effort, as opposed to codemodding away enough cases to manually fix the rest.

But it does seem like I ought to be able to parse those type annotations and do something reasonable if I put my mind to it. The main problem is that I'm not yet sure how to inject multiple statements in place of a single statement. 

I'd be shocked if there's no way to do this but the libcst docs aren't always easy to navigate if I don't already know the name of what I'm looking for

stroxler: The new type comment stripping is much more specific and will skip these cases (there's a unit test checking each of them).

stroxler:
Updates:

- We now handle multi-statement cases with semicolons.
- The type comment handling is now uglier but much tighter, we should never strip un-consumed comments

stroxler:
Unclear why the rust CI jobs are failing, but it's definitely not related

grievejia (approved):
The new revision looks much cleaner. I'm not super familiar with libcst idioms but I trust that you did the right thing there. 

pikachu_stamp

PR #595, opened 2022-01-12
Remove unused argument to handle_Index

martindemello:
the method is already being called without the `node` argument here: https://github.com/Instagram/LibCST/blob/main/libcst/codemod/visitors/_apply_type_annotations.py#L274

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -232,7 +232,7 @@ class TypeCollector(cst.CSTVisitor):
         else:
             return dequalified_node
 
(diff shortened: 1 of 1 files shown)


stroxler:
Looks good to me, thanks!

stroxler:
Grr - I keep expecting that button to open a dropdown where I can accept

stroxler (approved):
Good catch!

stroxler:
I'll merge once all CI jobs finish

martindemello:
found it via pytype, not sure why pyre missed it but i'll file a bug there

stroxler:
I think it's a known problem that pyre neglects to check the arity of function calls (I remember seeing an internal bug report about this recently)

But we should definitely file a report, I can probably prioritize fixing it soonish

PR #596, opened 2022-01-12
Merge in TypeVars and Generic base classes in ApplyTypeAnnotationVisitor

martindemello:
* Tracks TypeVars that are used in type annotations in the pyi file, and
  adds their Assign statements to the merged file.
* Adds Generic[T] as a base class if needed.

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -8,6 +8,8 @@ from dataclasses import dataclass, field
 from typing import Dict, List, Optional, Sequence, Set, Tuple, Union
 
 import libcst as cst
+import libcst.matchers as m
+
 from libcst.codemod._context import CodemodContext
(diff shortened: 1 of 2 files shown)


stroxler (commented):
This looks great, thanks!

One question for you: will it work for your use case to match on just `TypeVar`, under the assumption that a stub always uses `from typing import TypeVar`?

Reason being, I think the current match won't work if the call is to `typing.TypeVar`.

If that's good enough for you, I'd say let's merge and I'll look at generalizing later. If not, my suggestion would be to get rid of the filter on Name, and instead use the qualified name provider to get the full name and compare to `typing.TypeVar`.

martindemello:
yes, i think just `TypeVar` should be fine for now, we typically don't qualify typing imports. the logic to not add unnecessary typevars is not 100% but it should be good enough to merge and augment later.

stroxler (approved):
Sounds good.

I'll aim to use QalifiedNameProvider at some later point but happy to unblock you now

PR #598, opened 2022-01-12
change pyo3 as optional dependency in native Python Parser

sehz:
@zsol great work on Python Parser.  This makes it more accessible outside this project.

## Summary

Native Rust parser has `pyo3` as a mandatory dependency.  Since I am only interested in Python Parser, I want to change to optional dependency to reduce compile and build complexity.    With the change,`pyo3` is still included as a default feature thus everything is the same as before.   

Pyo3 related code changes (conversion) are segregated together. 

## Test Plan

The existing test plan should be sufficient since this is feature separation.


zsol (changes requested):
Thanks for the PR! Looks largely good, but I'd like to change two things before merging:
1. let's run something in CI to make sure the crate isn't broken without the new feature (at least it should build)
2. can we call the new feature `py` or `python`? Just in case later we decide to change pyo3 to something else

sehz:
> Thanks for the PR! Looks largely good, but I'd like to change two things before merging:
> 
> 1. let's run something in CI to make sure the crate isn't broken without the new feature (at least it should build)
> 2. can we call the new feature `py` or `python`? Just in case later we decide to change pyo3 to something else

Done.  Change to `py` feature flag

zsol (approved):
looks good, thanks for the contribution

PR #599, opened 2022-01-12
Full handling for applying type comments to Assign

stroxler:
## Summary

In the previous PR, I added basic support for converting an
Assign with a type comment to an AnnAssign, as long as there was
only one target.

This PR handles all fully PEP 484 compliant cases:
- multiple assignments
- multiple elements in the LHS l-value

We cannot handle arity errors because there's no way to do it. And
we don't try to handle the ambiguous case of multiple assignments with
mismatched arities (PEP 484 isn't super clear on which LHS is supposed
to pick up the type, we are conservative here). The ambiguous case is
probably very uncommon in real code anyway, multiple assignment is not
a widely used feature.

NOTE: my local pyre is complaining and I don't understand it, putting
a PR up anyway to see what CI does.

## Test Plan

There are new test cases covering:
- multiple elements in the LHS
- multiple assignment
- both of the above together
- semicolon expansion, which is handled differently in the cases
  where we have to add type declarations
- new error cases:
  - mismatched arity in both directions on one assignment
  - mismatched arity in multiple assignment
```
> python -m unittest libcst.codemod.commands.tests.test_convert_type_comments
.....
Ran 5 tests in 0.150s

OK
```

### DIFF
--- a/libcst/codemod/commands/convert_type_comments.py
+++ b/libcst/codemod/commands/convert_type_comments.py
@@ -105,7 +105,7 @@ class AnnotationSpreader:
     @staticmethod
     def unpack_target(
         target: cst.BaseExpression,
-    ) -> UnpackedTargets:
+    ) -> UnpackedTargets:  # pyre-ignore: I'm unsure what the problem is.
         """
         Take a (non-function-type) type comment and split it into
         components. A type comment body should always be either a single
@@ -208,17 +208,18 @@ class ConvertTypeComments(VisitorBasedCodemodCommand):
     """
 
     def __init__(self, context: CodemodContext) -> None:
-        if (sys.version_info.major, sys.version_info.minor) < (3, 8):
-            # The ast module did not get `type_comments` until Python 3.7.
-            # In 3.6, we should error than silently running a nonsense codemod.
+        if (sys.version_info.major, sys.version_info.minor) < (3, 9):
+            # The ast module did not get `unparse` until Python 3.9,
+            # or `type_comments` until Python 3.8
             #
-            # NOTE: it is possible to use the typed_ast library for 3.6, but
-            # this is not a high priority right now. See, e.g., the
-            # mypy.fastparse module.
+            # For earlier versions of python, raise early instead of failing
+            # later. It might be possible to use libcst parsing and the typed_ast
+            # library to support earlier python versions, but this is not a
+            # high priority.
             raise NotImplementedError(
                 "You are trying to run ConvertTypeComments on a "
                 + "python version without type comment support. Please "
-                + "try using python 3.8+ to run your codemod."
+                + "try using Python 3.9+ to run your codemod."
             )
         super().__init__(context)
 
--- a/libcst/codemod/commands/convert_type_comments.py
+++ b/libcst/codemod/commands/convert_type_comments.py
@@ -221,13 +221,15 @@ class ConvertTypeComments(VisitorBasedCodemodCommand):
             # or `type_comments` until Python 3.8
             #
             # For earlier versions of python, raise early instead of failing
-            # later. It might be possible to use libcst parsing and the typed_ast
-            # library to support earlier python versions, but this is not a
-            # high priority.
+            # later. It might be possible to use libcst parsing and the
+            # typed_ast library to support earlier python versions, but this is
+            # not a high priority.
             raise NotImplementedError(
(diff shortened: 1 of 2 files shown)


stroxler:
Looks like I'll need to bump the required python version or find a 3.8 substitute for ast.unparse

libcst/codemod/commands/convert_type_comments.py
```diff
@@ -56,29 +58,168 @@ def _convert_annotation(raw: str) -> cst.Annotation:
...
+        list of strings matching the tuple structure of the type comment.
+        """
+        # pyre-ignore[16]: the ast module stubs do not have full details
+        annotation_ast = ast.parse(type_comment, "<type_comment>", "eval").body
+        return AnnotationSpreader._unparse_annotation(annotation_ast)
+
+    @staticmethod
+    def unpack_target(
+        target: cst.BaseExpression,
+    ) -> UnpackedTargets:  # pyre-ignore: I'm unsure what the problem is.
```
stroxler, line 108: @grievejia any idea why pyre would complain about `UnpackedTargets` not being a valid type here?

The type error I was getting was:
```
libcst/codemod/commands/convert_type_comments.py:108:9 Undefined or invalid type [11]: Annotation `UnpackedTargets` is not defined as a type
```

But
(a) as far as I can tell there's not much difference between `UnpackedAnnotations` and `UnpackedTargets`, so I can't figure out why only one of them complains
(b) even weirder, it doesn't complain on line 127 but that's exactly the same type name

grievejia: > as far as I can tell there's not much difference between UnpackedAnnotations and UnpackedTargets

This definition doesn't make sense to me:
```
UnpackedTargets: TypeAlias = Union[cst.BaseExpression, "UnpackedTargets"]
```
It's like saying `type t = t` in OCaml -- you can't define `t` in terms of just `t` itself without some sort of additional labeling. Otherwise it would be impossible to inductively prove that any object *cannot* have that type. We should probably emit an error there, but iirc that might be tricky at implementation level. 

On the other hand, the definition of `UnpackedTargets` is OK. The self-recursion is prevented by the existence of `List`, and we can inductively say that the inhabitants of this type are integer trees and nothing else.  

Either way, what you are doing relies on Pyre's support for recursive type, which is a non-standard extension and I personally consider that experimental. If possible I would suggest avoiding that and simple do an old-school class hierarchy (e.g. one abstract base `UnpackedAnnotations` and two concrete subtypes `SingleUnpackedAnnotation` and `MultipleUnpackedAnnotations`, where the latter contains attributes of the parent type), which is guaranteed to always work. 

> even weirder, it doesn't complain on line 127 but that's exactly the same type name

Pyre does not guarantee it always emit all errors. We do some filtering to prevent diagnostics from getting too spammy. iirc one of the filtering rule is that if the invalid type error for a given name is reported once, we don't report it elsewhere in the same file.

stroxler: I hesitate to use a class hierarchy because that adds quite a lot of boilerplate - it becomes harder to follow the intent of the code from the types themselves plus I'd need unwrapping and wrapping logic.

I think using recursive types in a codemod ought to be alright, I probably wouldn't use it in the core of LibCST

stroxler: I made a PR to use a class hierarchy: https://github.com/stroxler/LibCST/pull/2

The change seems pretty bad to me, I'd much prefer to stick with recursive types

libcst/codemod/commands/convert_type_comments.py
```diff
@@ -56,29 +58,176 @@ def _convert_annotation(raw: str) -> cst.Annotation:
...
-            # NOTE: it is possible to use the typed_ast library for 3.6, but
-            # this is not a high priority right now. See, e.g., the
-            # mypy.fastparse module.
+            # For earlier versions of python, raise early instead of failing
+            # later. It might be possible to use libcst parsing and the typed_ast
+            # library to support earlier python versions, but this is not a
+            # high priority.
             raise NotImplementedError(
                 "You are trying to run ConvertTypeComments on a "
                 + "python version without type comment support. Please "
```
zsol, line 229: maybe update the error message here to avoid misleading users

zsol (approved):
lgtm

PR #600, opened 2022-01-13
Format _apply_type_annotations.py in margin-oriented style

stroxler:
In python code - and especially typed python code - I find it much easier to skim function headers if we *always* follow the style of
```
def f(
    x: SomeType
) -> ReturnType:
    ...
```

With this format I find it far easier to skim code because all the arg names are along the margin and the return types are in a predictable column. I don't usually change other folks' code to this style, but in the case of this module I'm the primary maintainer so I think it makes sense to go with what I find easier to maintain :)

(Someday I'm going to write a LibCST codemod that enforces this style!)


stroxler:
Closing this because it's going to clash with #596

I'm going to write a codemod that can do this automatically, and then run it on top of those changes after they merge :)

Issue #574, opened 2022-01-03
Slots added to many classes, but not their bases


ariebovenberg:
👋 Hi there!

I found you use `add_slots` to add slots to a lot of classes (e.g. `libcst._nodes.expression.Annotation`). However, it seems `__slots__` is not present in many of their base classes (e.g. `libcst._nodes.base.CSTNode`), meaning the benefits are reduced.

Unless I'm missing something, we should be able to fix this for a free performance boost!

If you agree, I'd be happy to submit a PR

edit: a full list of slotless bases:

```
       - libcst._nodes.base.BaseLeaf
       - libcst._nodes.base.BaseValueToken
       - libcst._nodes.base.CSTNode
       - libcst._nodes.expression.BaseAssignTargetExpression
       - libcst._nodes.expression.BaseComp
       - libcst._nodes.expression.BaseDelTargetExpression
       - libcst._nodes.expression.BaseDict
       - libcst._nodes.expression.BaseDictElement
       - libcst._nodes.expression.BaseElement
       - libcst._nodes.expression.BaseExpression
       - libcst._nodes.expression.BaseFormattedStringContent
       - libcst._nodes.expression.BaseList
       - libcst._nodes.expression.BaseNumber
       - libcst._nodes.expression.BaseSet
       - libcst._nodes.expression.BaseSimpleComp
       - libcst._nodes.expression.BaseSlice
       - libcst._nodes.expression.BaseString
       - libcst._nodes.expression._BaseElementImpl
       - libcst._nodes.expression._BaseExpressionWithArgs
       - libcst._nodes.expression._BaseParenthesizedNode
       - libcst._nodes.expression._BasePrefixedString
       - libcst._nodes.expression._BaseSetOrDict
       - libcst._nodes.op.BaseAugOp
       - libcst._nodes.op.BaseBinaryOp
       - libcst._nodes.op.BaseBooleanOp
       - libcst._nodes.op.BaseCompOp
       - libcst._nodes.op.BaseUnaryOp
       - libcst._nodes.op._BaseOneTokenOp
       - libcst._nodes.op._BaseTwoTokenOp
       - libcst._nodes.statement.BaseCompoundStatement
       - libcst._nodes.statement.BaseSmallStatement
       - libcst._nodes.statement.BaseStatement
       - libcst._nodes.statement.BaseSuite
       - libcst._nodes.statement._BaseSimpleStatement
       - libcst._nodes.whitespace.BaseParenthesizableWhitespace
[...]

PR #605, opened 2022-01-14
add slots to base classes, @add_slots takes bases into account

ariebovenberg:
## Summary

See #574: many slots were missing in base classes, meaning we didn't get the full memory savings or the ability to block arbitrary attribute setting.

Additionally, I found some slots overlapping where `add_slots` dataclasses inherited from each other. I think this [isn't even done right](https://bugs.python.org/issue46382) in the stdlib...

Some changes were needed to the tests now that arbitrary attributes could not be set. Here I was less sure of the desired outcome.

## Test Plan

Future violations can be prevented with [slotscheck](https://github.com/ariebovenberg/slotscheck) (which I developed).
See here a recent PR where I added it to the [returns library](https://github.com/dry-python/returns/pull/1233/).

I'm of course happy to include in dev/CI plan [as you indidated](https://github.com/Instagram/LibCST/issues/574#issuecomment-1006397621)


libcst/tests/test_batched_visitor.py
```diff
@@ -16,57 +16,57 @@ def test_simple(self) -> None:
         mock = Mock()
 
         class ABatchable(BatchableCSTVisitor):
-            def visit_Pass(self, node: cst.Pass) -> None:
-                mock.visited_a()
-                object.__setattr__(node, "a_attr", True)
+            def visit_Del(self, node: cst.Del) -> None:
+                object.__setattr__(node, "target", mock.visited_a())
```
ariebovenberg, line 20: I had to change to using `Del`, because `Pass` didn't have two fields available to reproduce the original test.

due to the slots, arbitrary fields can no longer be set. Existing fields need to be (ab)used.

libcst/tests/test_batched_visitor.py
```diff
@@ -16,57 +16,57 @@ def test_simple(self) -> None:
...
+        self.assertEqual(object.__getattribute__(del_, "semicolon"), mock.visited_b())
+
     def test_all_visits(self) -> None:
         mock = Mock()
 
         class Batchable(BatchableCSTVisitor):
-            def visit_Pass(self, node: cst.Pass) -> None:
-                mock.visit_Pass()
-                object.__setattr__(node, "visit_Pass", True)
+            def visit_If(self, node: cst.If) -> None:
```
ariebovenberg, line 41: I had to change to using `If`, because `Pass` didn't have four fields available to reproduce the original test.

due to the slots, arbitrary fields can no longer be set. Existing fields need to be (ab)used.

ariebovenberg:
Wow, I did a rough benchmark:

```python
from pympler.asizeof import asizeof
from libcst import parse_module
from pathlib import Path

parsed = parse_module(Path('libcst/_typed_visitor.py').read_text())
print(asizeof(parsed))
```

without patched slots: 15_240_816
with patched slots: 5_660_976

This cuts memory usage by a factor 3 :)

libcst/_nodes/base.py
```diff
@@ -109,6 +109,12 @@ def _clone(val: object) -> object:
 
 @dataclass(frozen=True)
 class CSTNode(ABC):
+
+    # pyre-ignore[4]: Attribute `__slots__` of class `CSTNode`
+    # has type `typing.Tuple[]` but no type is specified.
+    # But we know it's str
+    __slots__ = ()
```
zsol, line 116: I think you can remove this pyre-ignore by doing
```suggestion
    __slots__: Sequence[str] = ()
```

zsol (changes requested):
Thanks for the high quality PR! Much appreciated. All of this looks good to me. There's just one change I'd like to request before merging: can you please add a unit test that exercises the new logic in `add_slots`?

libcst/_nodes/base.py
ariebovenberg: I needed an additional `ClassVar` to prevent it from being picked up as a dataclass field

zsol (approved):
Let's go🚀! Thanks for the contribution

PR #611, opened 2022-01-17
Allow trailing whitespace without newline at EOF

### LINKED ISSUE #608: Native parser does not allow indentation mismatch at EOF
Short reproducer: (Only happens when the `LIBCST_PARSER_TYPE=native`)
```py
import ast
import libcst

src = "if bleh:\n  print()\n "
ast.parse(src) # works perfectly
libcst.parse_module(src) # INDENT of 2 space does not match with indent of 1 space
```

```
0,0-0,0:            ENCODING       'utf-8'        
1,0-1,2:            NAME           'if'           
1,3-1,7:            NAME           'bleh'         
1,7-1,8:            OP             ':'            
1,8-1,9:            NEWLINE        '\n'           
2,0-2,1:            INDENT         '\t'           
2,1-2,6:            NAME           'print'        
2,6-2,7:            OP             '('            
2,7-2,8:            OP             ')'            
2,8-2,9:            NEWLINE        '\n'           
3,0-3,0:            DEDENT         ''             
3,0-3,0:            ENDMARKER      ''             
```

zsol:
## Summary

The native tokenizer was a bit too strict about indentation checking at EOF. This is valid Python: `if 1:\n    pass\n ` (note the extra space at the end without a newline), but the tokenizer was trying to match the single space at the end as if it was a dedent from level 4 -> 1, which failed (there was never a level 1 indentation).

This PR relaxes the tokenizer a bit to allow for such cases. Indentation checking is skipped if after consuming all whitespace at the beginning of the line, we encounter EOF.

Fixes #608 

## Test Plan

Added unit test.

### DIFF
--- a/native/libcst/src/tokenizer/core/mod.rs
+++ b/native/libcst/src/tokenizer/core/mod.rs
@@ -533,9 +533,13 @@ impl<'t> TokState<'t> {
             }
         }
 
-        // Lines with only whitespace and/or comments and/or a line continuation character shouldn't
-        // affect the indentation and are not passed to the parser as NEWLINE tokens.
-        self.blank_line = matches!(self.text_pos.peek(), Some('#') | Some('\n') | Some('\\'));
+        // Lines with only whitespace and/or comments and/or a line continuation
+        // character shouldn't affect the indentation and are not passed to the parser
+        // as NEWLINE tokens.
+        self.blank_line = matches!(
+            self.text_pos.peek(),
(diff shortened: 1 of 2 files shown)


zsol:
cc @bgw AFAICT this was a bug in your original implementation as well

zsol:
CPython pure python tokenizer does the same: https://github.com/python/cpython/blob/main/Lib/tokenize.py#L495-L496

stroxler (approved):
Looks good to me

PR #622, opened 2022-01-22
Add support for methods with func type comment excluding self/cls

stroxler:
## Summary

PEP 484 doesn't really specify carefully how function type
comments should work on methods, but since usually the type of
`self` / `cls` is automatic, most use cases choose to only annotate
the other arguments.

As a result, this commit modifies our codemod so that non-static
methods can specify either all the arguments, or all but one of
them. We'll correctly zip together the inline and func-type-comment
types either way, typically getting no type for `cls` or `self`.

We accomplish this by using matchers to trigger the visit
method for FunctionDef rather than using visit_FunctionDef, which gives
us enough context to determine when a function def is a regular
function versus a method (plus also matching the decorators against
`@staticmethod`, so that we trigger the normal function logic in
that case).

## Test Plan

I added a new test verifying various method cases: static and non-static,
with and without annotations of `self` / `cls` in the func type comment,
with and without inline type comments.

### DIFF
--- a/libcst/codemod/commands/convert_type_comments.py
+++ b/libcst/codemod/commands/convert_type_comments.py
@@ -811,7 +811,7 @@ class ConvertTypeComments(VisitorBasedCodemodCommand):
         self,
         original_node: cst.ClassDef,
         updated_node: cst.ClassDef,
-    ) -> None:
+    ) -> cst.ClassDef:
         """
         Keep track of when we are and are not inside of classes, to help with
         heuristics to handle methods correctly.
--- a/libcst/codemod/commands/convert_type_comments.py
+++ b/libcst/codemod/commands/convert_type_comments.py
@@ -645,17 +645,12 @@ class ConvertTypeComments(VisitorBasedCodemodCommand):
     #
     # PEP 484 underspecifies how to apply type comments to (non-static)
     # methods - it would be possible to provide a type for `self`, or to omit
-    # it. We use heuristics to determine when a function is a method, and allow
-    # for the `self` or `cls` argument to have no type provided.
-    #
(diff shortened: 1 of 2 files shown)


libcst/codemod/commands/convert_type_comments.py
```diff
@@ -623,10 +642,31 @@ def leave_With(
...
+    #
+    # The heuristics are:
+    # - Is it inside of a class body? If not, it is never a method.
+    # - Is the name of the first parameter `cls` or `self`? This isn't required
+    #   but it's a strong enough convention that it should work as a heuristic.
 
-    def visit_FunctionDef(
+    def visit_Class(
         self,
         node: cst.FunctionDef,
```
zsol, line 658: How is this not a type error?
```suggestion
        node: cst.Class,
```

libcst/codemod/commands/convert_type_comments.py
```diff
@@ -748,6 +812,17 @@ def leave_FunctionDef(
         else:
             return updated_node
 
+    def leave_Class(
+        self,
+        original_node: cst.FunctionDef,
+        updated_node: cst.FunctionDef,
```
zsol, line 818:
```suggestion
        original_node: cst.Class,
        updated_node: cst.Class,
```

zsol (approved):

libcst/codemod/commands/convert_type_comments.py
stroxler: Woah.

I think it might be because the visitor pattern, as far as I can tell, is entirely dynamic so pyre may have no idea what's actually going on.

PR #624, opened 2022-01-22
[native] Don't redundantly nest StarredElement inside another Element

### LINKED ISSUE #623: Redundant nesting of StarredElement in Starred with native parser on some cases
Short reproducer:
```py
a, *b
```

Native parser represents `*b` like this:
```py
Element(
  value=StarredElement(
    value=Name(
      value='b',
    ),
  ),
),
```
But the python version represents it in a flat fashion:
```py
StarredElement(
  value=Name(
    value='b',
  ),
),
```

isidentical:
## Summary
Resolves #623

## Test Plan
Added 2 tests, which fail with the current `main` but pass with this patch.

### DIFF
--- a/libcst/_nodes/tests/test_tuple.py
+++ b/libcst/_nodes/tests/test_tuple.py
@@ -91,6 +91,47 @@ class TupleTest(CSTNodeTest):
                 "parser": parse_expression,
                 "expected_position": CodeRange((1, 1), (1, 11)),
             },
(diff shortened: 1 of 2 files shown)


stroxler (approved):
Looks good to me, thanks @isidentical!

cc @zsol for awareness, but since this is just making the native parser match existing behavior I'm going to go ahead and merge it

Issue #603, opened 2022-01-13
Getting `error: can't find Rust compiler` with 0.4.0 release


brtknr:
```
  Building wheel for libcst (pyproject.toml): started
  Building wheel for libcst (pyproject.toml): finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /usr/local/bin/python /usr/local/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py build_wheel /tmp/tmpo26gslix
       cwd: /tmp/pip-install-_dck7y8c/libcst_891d423a87d141e9a22cc7888d71a098
  Complete output (312 lines):
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.linux-aarch64-3.9
  creating build/lib.linux-aarch64-3.9/libcst
  copying libcst/_version.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_tabs.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_flatten_sentinel.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_batched_visitor.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_metadata_dependent.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_types.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_typed_visitor.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_exceptions.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_removal_sentinel.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_add_slots.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/__init__.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_maybe_sentinel.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_visitors.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_type_enforce.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_position.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/tool.py -> build/lib.linux-aarch64-3.9/libcst
  copying libcst/_typed_visitor_base.py -> build/lib.linux-aarch64-3.9/libcst
[...]

PR #630, opened 2022-01-26
build wheels on main branch for linux/arm64

zsol:
## Test Plan
In [this commit](https://github.com/Instagram/LibCST/pull/630/commits/e3b89e499fd9984ef10d0e1a9a8b46d24df2de24) I enabled running the arm builds on this PR only. That commit is removed before merging.
See https://github.com/Instagram/LibCST/actions/runs/1752265369 for an example run.

Should help address #603.

### DIFF
--- a/.github/build-matrix.json
+++ b/.github/build-matrix.json
@@ -1,0 +1,31 @@
+[
+    {
+        "vers": "x86_64",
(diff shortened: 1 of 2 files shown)


lpetre (approved):
I like it!

PR #641, opened 2022-02-08
Fix space validation for AsName and Await

### LINKED ISSUE #633: Incorrect CST node validation: space is not necessary
At least for `WithAs` and `Await`, space is not required after the keywords in certain cases.

```
python -c 'import ast; ast.parse("with foo()as bar: await(bar)")'
```

_Originally posted by @zsol in https://github.com/Instagram/LibCST/issues/625#issuecomment-1023326563_

zsol:
## Summary

Fixes #633. Unfortunately we have to move the validation of `AsName` to its parent because of different rules for different parents as well as the need for the preceding expression.

## Test Plan

Added unit tests.

### DIFF
--- a/libcst/_nodes/tests/test_import.py
+++ b/libcst/_nodes/tests/test_import.py
@@ -592,7 +592,7 @@ class ImportFromCreateTest(CSTNodeTest):
                                 ),
                                 whitespace_before_as=cst.SimpleWhitespace(""),
                             ),
-                        )
+                        ),
                     ),
                 ),
                 "expected_re": "one space before as keyword",
--- a/libcst/_nodes/tests/test_import.py
+++ b/libcst/_nodes/tests/test_import.py
@@ -197,6 +197,20 @@ class ImportCreateTest(CSTNodeTest):
             },
             {
                 "get_node": lambda: cst.Import(
+                    names=(
+                        cst.ImportAlias(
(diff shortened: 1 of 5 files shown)


stroxler (approved):
Looks good to me!

stroxler:
Whoops, looks like there's a type check error

libcst/_nodes/tests/test_import.py
```diff
@@ -564,6 +578,25 @@ def test_valid(self, **kwargs: Any) -> None:
...
+                            cst.Name("bar"),
+                            asname=cst.AsName(
+                                cst.Name(
+                                    "baz",
+                                    lpar=(cst.LeftParen(),),
+                                    rpar=(cst.RightParen(),),
+                                ),
+                                whitespace_before_as=cst.SimpleWhitespace(""),
+                            ),
+                        )
```
stroxler, line 595: Found the type check problem - need a trailing comma to make `name` a tuple:
```suggestion
                        ),
```

I'd be tempted to just use lists instead, tuples are confusing because of the trailing-comma-for-singleton issue. But if we switch to lists we'd want to do it everywhere in a dedicated PR.

stroxler (approved):
I think I found the typecheck problem, once that's patched we should be good to merge.

zsol:
hehe thanks!

PR #643, opened 2022-02-09
Create an AddTrailingCommas codemod

stroxler:
## Summary

This codemod adds trailing commas to parameter and arguments
lists when there are sufficient arguments or parameters.

The idea is this:
- both black and yapf will generally split lines when there
  are trailing commas at the end of a parameter / arguments list
- It's easier on my eye to have names and types in more predictable
  locations within a function header, i.e. left-aligned. And in
  function calls, I also find it easier to compare arguments to
  function parameters whenever the arguments are one-per line, at
  least when there are more than two arguments.

By default, we ensure trailing commas for functions with one or more
parameters (but do not include `self` or `cls` method arguments) which
is suitable for `black`, and calls with 3 or more arguments.

Both the parameter count and the argument count can be overridden.
Moreover, by passing `--formatter yapf` someone can use the
yapf-suitable default of 2 parameters which is handy since then the
user doesn't have to memorize black vs yapf settings; this is necesary
because yapf does not split lines after a trailing comma in one-argument
defines.

## Test Plan

```
> python -m unittest libcst.codemod.commands.tests.test_add_trailing_commas
......
Ran 6 tests in 0.134s

OK
```

For a real-world test (and an example of the affect of applying this change) see #642 where I
apply this codemod to _apply_type_annotations.py

### DIFF
--- a/libcst/codemod/commands/add_trailing_commas.py
+++ b/libcst/codemod/commands/add_trailing_commas.py
@@ -53,7 +53,7 @@ class AddTrailingCommas(VisitorBasedCodemodCommand):
         if presets is None:
             raise ValueError(
                 f"Unknown formatter {formatter!r}. Presets exist for "
-                + ', '.join(presets_per_formatter.keys())
+                + ", ".join(presets_per_formatter.keys())
             )
         self.parameter_count: int = parameter_count or presets["parameter_count"]
         self.argument_count: int = argument_count or presets["argument_count"]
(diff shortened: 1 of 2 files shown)


libcst/codemod/commands/add_trailing_commas.py
```diff
@@ -0,0 +1,127 @@
...
+from typing import Dict, Optional
+
+import libcst as cst
+from libcst.codemod import CodemodContext, VisitorBasedCodemodCommand
+
+
+presets_per_formatter: Dict[str, Dict[str, int]] = {
+    "black": {
+        "parameter_count": 1,
+        "argument_count": 3,
```
zsol, line 17: this `3` feels arbitrary, but whatever :)

zsol (approved):

libcst/codemod/commands/add_trailing_commas.py
stroxler: Yeah. I'm pretty sure 1 is too aggressive on function calls, I was torn between 2 and 3.

PR #648, opened 2022-02-16
Add a check for copyright headers to lint stage

stroxler:
## Summary

I had a copy-paste error in add_trailing_commas.py.

I'll fix it in a follow-up PR, but first I want to make sure we can make CI fail so that this doesn't go unnoticed again.

## Test Plan

Wait for github actions - if lint fails, this change is good.


stroxler:
Lint failed, looks like the change worked.

PR #649, opened 2022-02-16
Fix the copyright heading on add_trailing_commas.py

stroxler:
## Summary

I had a copy-paste error.

## Test Plan

We now check headers in the lint stage, so as soon as that comes back clean we are good (I also did run the check locally, but I want to see CI working end-to-end before merging).


stroxler:
Lint came back clean

Since there's no code change I'll go ahead and merge

PR #657, opened 2022-03-08
Define gather global names visitor

shannonzhu:
## Summary

Having a global name gatherer available helps us determine which names are in or out of scope when applying annotations.  This is similar to how we are currently using the import gatherer visitor in a pass prior to applying annotations, but that gatherer seems to have use cases perhaps beyond typing codemods - having a stand alone one for names seemed appropriate. 

Ideally we'd like to have better tracking for which names are in/out of scope that combines imports and globals/classes, which we could use in the apply type annotations codemod visitor to compare qualified names against. For now, having global names pre-calculated lets us determine whether or not a name in an annotation is a forward reference by checking if it's a name defined in the current module that we haven't yet seen (all others are considered builtins). 

## Test Plan

```
python3 -m unittest libcst.codemod.visitors.tests.test_gather_global_names
...
Ran 3 tests in 0.070s

OK
```


zsol (approved):
Seems reasonable, thanks!

PR #660, opened 2022-03-22
Update relative import logic to match cpython

lpetre:
## Summary

 * LibCST does not correctly handle relative imports from an __init__.py file
 * Python relative imports work based on package name, not module name.
 * `a/b.y` and `a/b/__init__.py` have the same module name, but different package names.
 * So the same relative import statement in both of those files mean different things.
 * Currently libcst only properly handles the `a/b.y` case correctly

## Test Plan
I've added a bunch of test cases to show that this was not working correctly.

### DIFF
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -241,11 +241,6 @@ def _execute_transform(  # noqa: C901
                 ),
             )
 
-        # determine the module and package name for this file
-        module_name_and_package = calculate_module_and_package(
-            config.repo_root or Path.cwd(), filename
-        )
-
         # Somewhat gross hack to provide the filename in the transform's context.
         # We do this after the fork so that a context that was initialized with
         # some defaults before calling parallel_exec_transform_with_prettyprint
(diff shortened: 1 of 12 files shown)


libcst/codemod/_cli.py
```diff
@@ -241,27 +241,23 @@ def _execute_transform(  # noqa: C901
                 ),
             )
 
+        # determine the module and package name for this file
+        module_name_and_package = calculate_module_and_package(
+            config.repo_root or Path.cwd(), filename
```
lpetre, line 246: I think this is breaking tests on github so I will revisit this

libcst/codemod/_cli.py
```diff
@@ -251,16 +251,18 @@ def _execute_transform(  # noqa: C901
...
+            module_name_and_package = calculate_module_and_package(
+                config.repo_root or ".", filename
+            )
             transformer.context = replace(
                 transformer.context,
                 full_module_name=module_name_and_package.name,
                 full_package_name=module_name_and_package.package,
             )
+        except ValueError:
+            pass
```
zsol, line 265: Is this something we should at least log about?

zsol (approved):
thank you!

libcst/codemod/_cli.py
lpetre: Good idea

PR #662, opened 2022-03-23
Support module and package names in the codemod context

lpetre:
## Summary

Relative imports are relative to the module's package, not the module's name. 

Therefore we need to compute the package from the file.

## Test Plan
```
~/EllipticalFrightenedOpposites$ tree -I __pycache__ x/
x/
├── __init__.py
└── y
    ├── __init__.py
    └── z.py

1 directory, 3 files

~/EllipticalFrightenedOpposites$ python3 -c 'import x.y.z as mod; print(mod.__package__, mod.__name__)'
x.y x.y.z

~/EllipticalFrightenedOpposites$ python3 -c 'import x.y as mod; print(mod.__package__, mod.__name__)'
x.y x.y
```


libcst/codemod/_cli.py
```diff
@@ -184,28 +184,31 @@ def exec_transform_with_prettyprint(
     return maybe_code
 
 
-def _calculate_module(repo_root: Optional[str], filename: str) -> Optional[str]:
+def _calculate_module_and_package(repo_root: Optional[str], filename: str) -> Tuple[Optional[str], Optional[str]]:
```
zsol, line 187: They can only be both None or both not-None
```suggestion
def _calculate_module_and_package(repo_root: Optional[str], filename: str) -> Optional[Tuple[str, str]]:
```

zsol (approved):
Nice.

PR #663, opened 2022-03-23
Drop support for running libcst using a python 3.6 interpreter

lpetre:
## Summary
Python 3.6 security support ended in December 2021. Some major libraries (including some transitive dependencies) have stopped supporting 3.6.


setup.py
```diff
@@ -30,7 +30,7 @@ def no_local_scheme(version: str) -> str:
         ),
     },
     name="libcst",
-    description="A concrete syntax tree with AST-like properties for Python 3.5, 3.6, 3.7 and 3.8 programs.",
+    description="A concrete syntax tree with AST-like properties for Python 3.7, 3.8, 3.9, and 3.10 programs.",
```
zsol, line 33: we still support parsing 3.5, 3.6 syntax

zsol (changes requested):
Readme needs some changing too

zsol (approved):

PR #664, opened 2022-03-23
Preserve as-imports when merging type annotations.

### LINKED ISSUE #661: ApplyTypeAnnotationsVisitor handles aliased imports badly
merging py:
```
def f(x): pass
```
and pyi:
```
from foo import A as B
def f(x: B): ...
```
yields
```
from foo import A
def f(x: B): pass
```

martindemello:
# Summary

Preserves the aliased import name when merging in annotations like
```
from foo import A as B

def f(x: B): ...
```

Fixes #661

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -369,8 +369,7 @@ class TypeCollector(m.MatcherDecoratableVisitor):
         return (relative_prefix + qualifier, target)
 
     def _handle_qualification_and_should_qualify(
-        self,
-        qualified_name: str,
+        self, qualified_name: str, node: Optional[cst.CSTNode] = None
     ) -> bool:
         """
         Based on a qualified name and the existing module imports, record that
@@ -384,10 +383,15 @@ class TypeCollector(m.MatcherDecoratableVisitor):
             if module in self.existing_imports:
                 return True
             else:
+                if node and isinstance(node, cst.Name):
+                    asname = node.value
+                else:
+                    asname = None
                 AddImportsVisitor.add_needed_import(
                     self.context,
                     module,
                     target,
+                    asname=asname,
                 )
                 return False
         return False
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -383,7 +383,7 @@ class TypeCollector(m.MatcherDecoratableVisitor):
             if module in self.existing_imports:
                 return True
             else:
-                if node and isinstance(node, cst.Name):
+                if node and isinstance(node, cst.Name) and node.value != target:
                     asname = node.value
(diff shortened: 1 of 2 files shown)


martindemello:
i do not think any of the CI failures are related to this pull request - could someone please take a look and verify that?

stroxler:
The CI failures appear to be related to some kind of bad environment setup - most of them seem to be `black` existing witn nonzero status.

That seems unrelated, except that there are more recent pull requests that are succeeding and I can't explain why that would be. I can try to dig in but also @zsol it might be worth glancing at them in case you can see the problem right away

stroxler:
The CI failures appear to be related to some kind of bad environment setup - most of them seem to be `black` existing witn nonzero status.

That seems unrelated, except that there are more recent pull requests that are succeeding and I can't explain why that would be. I can try to dig in but also @zsol it might be worth glancing at them in case you can see the problem right away

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -369,8 +369,7 @@ def _module_and_target(
         return (relative_prefix + qualifier, target)
 
     def _handle_qualification_and_should_qualify(
-        self,
-        qualified_name: str,
+        self, qualified_name: str, node: Optional[cst.CSTNode] = None
```
stroxler, line 372: I'm not seeing how this code changes the end-to-end behavior, given that we don't currently include the extra argument in any of the callsites.

I think this might be related to what seems like a gap in the tests

libcst/codemod/visitors/tests/test_apply_type_annotations.py
```diff
@@ -292,6 +292,23 @@ def foo() -> bar.Baz:
                     return returns_baz()
                 """,
             ),
+            "with_as_import": (
+                """
+                def foo(x): ...
+                """,
```
stroxler, line 298: In the test cases, the stub comes first, then the input code, then the output code.

So I don't think this test case is verifying what you are hoping for - here we're just not applying a stub with no annotations.

stroxler (changes requested):

stroxler:
@martindemello  I made a separate, even tinier PR and got the same CI failures, so I'm confident it isn't related - my best guess is that it's a version mismatch due to new packages on pypi.

libcst/codemod/visitors/tests/test_apply_type_annotations.py
martindemello: oops, no wonder it was still passing when i messed up a rebase :) fixed both now.

libcst/codemod/visitors/_apply_type_annotations.py
martindemello: i messed up a rebase, it was meant to add the node argument when setting `should_qualify` below

stroxler (approved):
Looks good to me

PR #665, opened 2022-03-24
Clean warnings in context when transforming multiple files

jschavesr:
## Summary

Using libCST cli tool allows to run a codemod over files inside a folder recursively. Context warning messages were not being cleaned between files, so warnings from one files could be passed to other files in the context. Because of that warnings were being shown wrongly in the summary and the total number of warnings was wrong as well.

## Test Plan

`python3 -m unittest libcst.codemod.tests.test_codemod_cli `

### DIFF
--- a/libcst/codemod/tests/test_codemod_cli.py
+++ b/libcst/codemod/tests/test_codemod_cli.py
@@ -7,11 +7,12 @@
 
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 
 from libcst._parser.entrypoints import is_native
 from libcst.testing.utils import UnitTest
-
+from libcst.codemod import CodemodTest
 
 class TestCodemodCLI(UnitTest):
     def test_codemod_formatter_error_input(self) -> None:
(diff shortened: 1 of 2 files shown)


stroxler (approved):
Thanks for the fix @jschavesr!

libcst/codemod/tests/test_codemod_cli.py
```diff
@@ -38,3 +39,34 @@ def test_codemod_formatter_error_input(self) -> None:
...
+                    "codemod",
+                    "convert_format_to_fstring.ConvertFormatStringCommand",
+                    p,
+                ],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+            )
+            # Each module will generate 2 warnings, so we should get 6 warnings in total
+            self.assertIn(
+                "- 6 warnings were generated.",
```
stroxler, line 70: I take it 6 because there are 2 per file - would it have been 2 + 4 + 6 = 12 prior to this fix?

stroxler:
Running CI, I'll merge if it comes back clean

libcst/codemod/tests/test_codemod_cli.py
```diff
@@ -38,3 +39,34 @@ def test_codemod_formatter_error_input(self) -> None:
...
+            (p / "mod2.py").write_text(CodemodTest.make_fixture_data(mod))
+            (p / "mod3.py").write_text(CodemodTest.make_fixture_data(mod))
+            rlt = subprocess.run(
+                [
+                    "python",
+                    "-m",
+                    "libcst.tool",
+                    "codemod",
+                    "convert_format_to_fstring.ConvertFormatStringCommand",
+                    p,
```
stroxler, line 63: Looks like we're getting test failures on Python 3.7 + Windows.

I think the problem may be here - the other test in this module calls `str` on the Path before passing it to `subprocess.run`.

stroxler (changes requested):
Unfortunately we need to fix a Windows bug before we merge - I commented where I think the problem is coming from.

From now on, I think every time you push changes it should run CI automatically - it has to be manually approved the first time you make a PR

kiri11:
@jschavesr Do you plan on completing this? Looks very useful, without this change it's impossible to use warnings in multiprocessing context. They all get duplicated and assigned to wrong files. If you are no longer interested, I could take over and submit an updated copy of this PR. Thanks!

Issue #656, opened 2022-03-08
QualifiedNameProvider return multi nodes strangely


wiyr:
### env

python 3.7.5
libcst 0.3.21 or 0.4.1
Linux ubuntu 4.15.0-122-generic #124-Ubuntu SMP Thu Oct 15 13:03:05 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

### replay code
```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

from typing import Union, Optional
import libcst as cst
from libcst.metadata import PositionProvider, QualifiedNameProvider, ScopeProvider, ParentNodeProvider
from libcst._metadata_dependent import _T, _UNDEFINED_DEFAULT


class BaseConverter(cst.CSTTransformer):
    METADATA_DEPENDENCIES = (PositionProvider, QualifiedNameProvider, ScopeProvider, ParentNodeProvider)

    def __init__(self):
        super().__init__()
        self.wrap_converter = None

    def get_full_name_for_node(self, node: Union[str, cst.CSTNode]) -> Optional[str]:
        name_list = list(self.get_metadata(QualifiedNameProvider, node))
        print(f"full name list: {name_list}")
        return name_list

    def visit_Call(self, node: cst.Call):
        self.get_full_name_for_node(node)


if __name__ == '__main__':
    code = '''
from a import A, Ab
Ab()
'''

    converter = BaseConverter()
    wrapped_module = cst.MetadataWrapper(cst.parse_module(code))
    new_module = wrapped_module.visit(converter)
    # print(new_module.code)
```

### output
```
full name list: [QualifiedName(name='a.A.b', source=<QualifiedNameSource.IMPORT: 1>), QualifiedName(name='a.Ab', source=<QualifiedNameSource.IMPORT: 1>)]
```

### expect

`self.get_metadata(QualifiedNameProvider, node)` return only one node, named `a.Ab`

zsol:
Looks like this is a bug here: https://github.com/Instagram/LibCST/blob/17137014ac38b066f3b6e86f5eb36e20e2efbb58/libcst/metadata/scope_provider.py#L306

`a.A` is a prefix of `a.Ab` so that `if` condition is getting triggered for the first `ImportAlias` but it shouldn't.
This was introduced by https://github.com/Instagram/LibCST/pull/557 so I'm wondering if @giomeg has thoughts ;)

wiyr:
waiting for final conclusion or pull request

[...]

PR #669, opened 2022-04-07
fix qualified name get bug

wiyr:
## Summary
resolves https://github.com/Instagram/LibCST/issues/656
## Test Plan
 python -m unittest libcst.metadata.tests.test_scope_provider.ScopeProviderTest.test_get_qualified_names_for_the_same_prefix

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -304,7 +304,10 @@ class ImportAssignment(Assignment):
                         if eval_alias is not None:
                             as_name = eval_alias
                     if full_name.startswith(as_name):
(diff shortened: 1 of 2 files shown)


stroxler:
Thanks for the fix @wiyr!

This looks good to me, but it would be clearer if it shipped with a new unit test that would have failed on the old code. Do you have time to add that?

stroxler:
FYI the CI failures are not related - they are due to a click/black version mismatch and I'm working on a fix

wiyr:
added one simple unittest

stroxler (approved):
Awesome, thanks for the fix!

PR #674, opened 2022-04-09
Qualify imported symbols when the dequalified form would cause a conflict

### LINKED ISSUE #673: ApplyTypeAnnotationsVisitor: imported symbols should remain qualified when they conflict
t.py:
```
def f(x): pass
```
t.pyi:
```
import foo
from bar import A

def f(x: A) -> foo.A: ...
```
merge result:
```
from bar import A
from foo import A

def f(x: A) -> A: pass
```

martindemello:
## Summary

Adds a preliminary pass that scans the stub file for all imported
symbols, and collects the ones that cannot be safely dequalified.

Fixes #673

## Test Plan

Added a test case to test_apply_type_annotations.

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -873,23 +873,23 @@ class ApplyTypeAnnotationsVisitor(ContextAwareTransformer):
         import_gatherer = GatherImportsVisitor(CodemodContext())
         stub.visit(import_gatherer)
         symbol_map = import_gatherer.symbol_mapping
-        isc = ImportedSymbolCollector(existing_import_names, self.context)
-        cst.MetadataWrapper(stub).visit(isc)
+        symbol_collector = ImportedSymbolCollector(existing_import_names, self.context)
+        cst.MetadataWrapper(stub).visit(symbol_collector)
         module_imports = {}
-        for sym, isyms in isc.imported_symbols.items():
-            if len(isyms) == 1:
+        for imported_symbols in symbol_collector.imported_symbols.values():
+            if len(imported_symbols) == 1:
                 # If we have a single use of a symbol we can from-import it
                 continue
             used = False
-            for isym in isyms:
-                if not isym.symbol:
+            for imp_sym in imported_symbols:
+                if not imp_sym.symbol:
                     continue
-                imp = symbol_map.get(isym.symbol)
-                if not used and imp and imp.module_name == isym.module_name:
+                imp = symbol_map.get(imp_sym.symbol)
+                if not used and imp and imp.module_name == imp_sym.module_name:
                     # We can only import a symbol directly once.
                     used = True
                 else:
-                    imp = symbol_map.get(isym.module_symbol)
+                    imp = symbol_map.get(imp_sym.module_symbol)
                     if imp:
                         # imp will be None in corner cases like
                         #   import foo.bar as Baz
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -561,7 +561,7 @@ class TypeCollector(m.MatcherDecoratableVisitor):
             qualified_node = (
                 cst.parse_module(qualified_name) if isinstance(node, cst.Name) else node
             )
-            return qualified_node
+            return qualified_node  # pyre-ignore[7]
         else:
             dequalified_node = node.attr if isinstance(node, cst.Attribute) else node
             return dequalified_node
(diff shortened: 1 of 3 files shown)


libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -697,6 +851,54 @@ def transform_module_impl(
...
+        # annotation visitor hits `quux.X` it will retrieve the canonical name
+        # `foo.X` and then note that `foo` is in the module imports map, so it will
+        # leave the symbol qualified.
+        import_gatherer = GatherImportsVisitor(CodemodContext())
+        stub.visit(import_gatherer)
+        symbol_map = import_gatherer.symbol_mapping
+        isc = ImportedSymbolCollector(existing_import_names, self.context)
+        cst.MetadataWrapper(stub).visit(isc)
+        module_imports = {}
+        for sym, isyms in isc.imported_symbols.items():
```
stroxler, line 879: What does `isyms` refer to?

I'm not sure about LibCST, but on pyre we're generally anti abbreviations; I personally don't mind some acronyms like `sym` too much but the combination of `isc`, `sym`, `isym`, and `imp` starts to be a bit alphabet-soup-y

libcst/codemod/visitors/_gather_imports.py
```diff
@@ -114,3 +120,9 @@ def visit_ImportFrom(self, node: libcst.ImportFrom) -> None:
                     return
 
                 self.object_mapping[module].update(new_objects)
+            for ia in nodenames:
```
stroxler, line 123: It's probably better to use a longer name for `ia` here as well (imported_attribute I'm guessing?)

stroxler (changes requested):
This looks good to me, up to being a bit too abbreviation-heavy for the existing style.

But it looks to me like it probably only handles collisions within the stub - that's all that's tested, and as far as I can tell we're only running `import_gatherer` on the stub.

If that's the case, we should consider also handling collisions with the existing code - I'm fine with deferring that (I could try adding it later) but curious if it would be easy to add right now since you already have context.

libcst/codemod/visitors/_apply_type_annotations.py
martindemello: done

martindemello:
handling collisions in the existing code looks simple enough to add to this PR after all, will push an update.

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -697,6 +844,69 @@ def transform_module_impl(
...
+                used = False
+            for imp_sym in imported_symbols:
+                if not imp_sym.symbol:
+                    continue
+                imp = symbol_map.get(imp_sym.symbol)
+                if not used and imp and imp.module_name == imp_sym.module_name:
+                    # We can only import a symbol directly once.
+                    used = True
+                elif sym in existing_import_names:
+                    module_imports[imp.module_name] = imp
```
stroxler, line 898: Pyre thinks you need an `and imp` condition here - it's not type checking because `imp` could be `None`

libcst/codemod/visitors/_apply_type_annotations.py
```diff
@@ -407,17 +552,18 @@ def _handle_NameOrAttribute(
...
         should_qualify = self._handle_qualification_and_should_qualify(
             qualified_name, node
         )
         self.annotations.names.add(qualified_name)
         if should_qualify:
-            return node
+            qualified_node = (
+                cst.parse_module(qualified_name) if isinstance(node, cst.Name) else node
+            )
+            return qualified_node
```
stroxler, line 564: It looks like `pyre` thinks this can potentially have type `Module`, which means we may need to update the annotation on line 555.

stroxler (commented):
Nice, this looks good.

Pyre is complaining about a couple of lines. We don't yet have proper github integration (hopefully in a couple of weeks!) so it's a pain to find the problem, I commented on the relevant lines

libcst/codemod/visitors/_apply_type_annotations.py
martindemello: i suppressed the error; `qualified_node` will always have type `cst.Attribute` here but we cannot determine that statically.

libcst/codemod/visitors/_apply_type_annotations.py
martindemello: fixed

libcst/codemod/visitors/_apply_type_annotations.py
martindemello: just running `pyre check` in my checked out copy did identify the errors, btw.

stroxler (approved):
Looks good!

stroxler:
Thanks for the fix @martindemello - I've been wanting to handle this edge case for a long time, we actually hit it pretty often in `pyre infer`

martindemello:
i've been wanting to handle it for a while too :) imports are the last thing blocking us from using this in pytype.

PR #682, opened 2022-05-03
Consider access information when computing qualified names for nodes

lpetre:
## Summary
`Scope.get_qualified_names_for` has to do a lot of work to determine the qualified name for a node. The current version does not consider the data we have for accesses. If the given node is known to be an access, we can take advantage of that to work out the proper name.

This also means we now get the qualified names for string annotations for imports.

## Test Plan
I've added a new test case in the QualifiedNameProviderTest to show the before/after effects of this change


zsol (approved):
lgtm!

PR #688, opened 2022-05-16
Add package links to PyPI

adamchainz:
## Summary

Add two links to the sidebar on PyPI, to make it easier for folks to jump to the relevant pages. Example: https://pypi.org/project/django-htmx/ (PyPI adds the icons based on the link names).

## Test Plan

You'll see them on next release.


zsol:
TIL. Thanks!

PR #696, opened 2022-06-12
Add support for PEP-646

zsol:
## Summary
Implementation and tests based on https://github.com/python/cpython/pull/31018
The CST is modeled slightly differently from the AST:
- `Index` can now have an optional `star` (and corresponding `whitespace_after_star`)
- `StarredElement` is now a `BaseExpression`

## Test Plan

- [x] roundtrip tests
- [x] negative parse tests
- [x] parse tests
- [x] node construction tests

## TODO

- [ ] Maybe make `whitespace_after_star` into a `MaybeSentinel` instead of `Optional` and add corresponding codegen logic

### DIFF
--- a/libcst/_nodes/expression.py
+++ b/libcst/_nodes/expression.py
@@ -1438,10 +1438,29 @@ class Index(BaseSlice):
     #: The index value itself.
     value: BaseExpression
 
+    #: An optional string with an asterisk appearing before the name. This is
+    #: expanded into variable number of positional arguments. See PEP-646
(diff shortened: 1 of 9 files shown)


tusharsadhwani:
Can `whitespace_after_star` not be a `MaybeSentinel` then? Are there complications?

zsol:
@tusharsadhwani I decided against it because there doesn't seem to be any difference in behavior (the value of None always translates to the same as MaybeSentinel would) and Optional seems more semantically correct. I'm open to be convinced otherwise, though

Issue #668, opened 2022-03-30
Escaping quote characters in raw string literals causes a tokenizer error


lpetre:
This string breaks the native parser: `rf"\"{feature_name.lower()}\""`

```
$ echo 'rf"\"{feature_name.lower()}\""' | LIBCST_PARSER_TYPE=native python -m libcst.tool print - > /dev/null
Traceback (most recent call last):
  File "<string>", line 49, in <module>
  File "<string>", line 47, in __run
  File "lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "libcst/__main__.py", line 25, in <module>
    sys.exit(fbcode_main())
  File "libcst/__main__.py", line 19, in fbcode_main
    return libcst.tool.main(
  File "libcst/tool.py", line 834, in main
    return lookup.get(args.action or None, _invalid_command)(proc_name, command_args)
  File "libcst/tool.py", line 278, in _print_tree_impl
    tree = parse_module(
  File "libcst/_parser/entrypoints.py", line 109, in parse_module
    result = _parse(
  File "libcst/_parser/entrypoints.py", line 55, in _parse
    return parse(source_str)
libcst._exceptions.ParserSyntaxError: Syntax Error @ 0:1.
tokenizer error: unexpected characters after a line continuation

rf"\"{feature_name.lower()}\""
^
$
```

```
$ echo 'rf"\"{feature_name.lower()}\""' | python -m libcst.tool print - > /dev/null
$
```

lpetre:
Not sure if is the same issue, but this also breaks the native parser: `f"regexp_like(path, '.*\{file_type}$')"`

lpetre:
This is another example: `f"\{{\}}"` which fails both libcst parsers but `python3 -m py_compile` parses it.

PR #700, opened 2022-06-14
Only skip supported escaped characters in f-strings

zsol:
## Summary
When tokenizing f-strings, we should only skip characters that are allowed to be escaped ([spec](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals))

## Test Plan

Added test cases.

### DIFF
--- a/native/libcst/src/tokenizer/core/mod.rs
+++ b/native/libcst/src/tokenizer/core/mod.rs
@@ -940,7 +940,25 @@ impl<'t> TokState<'t> {
                             // skip escaped char (e.g. \', \", or newline/line continuation)
                             self.text_pos.next();
                         }
-                    } else {
+                    } else if let Some(
+                        '\n'
+                        | '\\'
+                        | '\''
+                        | '"'
+                        | 'a'
+                        | 'b'
+                        | 'f'
+                        | 'n'
+                        | 'r'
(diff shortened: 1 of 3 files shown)


zsol:
@bgw if you have some time I'd appreciate your eyes on this :) There's one more fix like this coming (to address #668

native/libcst/src/tokenizer/core/mod.rs
```diff
@@ -940,7 +940,25 @@ impl<'t> TokState<'t> {
                             // skip escaped char (e.g. \', \", or newline/line continuation)
                             self.text_pos.next();
                         }
-                    } else {
+                    } else if let Some(
```
bgw, line 943: Nice, this syntax was finally added in Rust 1.53.0: https://blog.rust-lang.org/2021/06/17/Rust-1.53.0.html#or-patterns

We could probably clean up the rest of the tokenizer to use this more compact match format.

bgw (approved):
LGTM. Thanks for fixing!

I think we could greatly reduce the number of bugs here if we stopped splitting f-strings, and instead recursively ran the parser, like CPython does. But that would involve a bigger architectural change.

zsol:
Agreed on both counts. Thanks for taking a look

PR #702, opened 2022-06-15
convert_type_comments now preserves comments following type comments

superbobry:
## Summary

Prior to this PR convert_type_comments codemod stripped any comments following a type comment. So, for example,

    y = 5  # type: int  # foo

was converted to

    y: int = 5

instead of

    y: int = 5  # foo

The PR changes the codemod to preserve any such comments.

## Test Plan

Added a new test case.

### DIFF
--- a/libcst/codemod/commands/convert_type_comments.py
+++ b/libcst/codemod/commands/convert_type_comments.py
@@ -126,6 +126,19 @@ def _is_type_comment(comment: Optional[cst.Comment]) -> bool:
     return True
 
 
+def _strip_type_comment(comment: Optional[cst.Comment]) -> Optional[cst.Comment]:
+    """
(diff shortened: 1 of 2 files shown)


zsol:
I'll let @stroxler chime in, but let me ask: what happens for `y = 5  # foo  # type: int`?

superbobry:
Thanks @zsol! `y = 5 # foo # type: int` is not a valid type comment according to [PEP-484](https://peps.python.org/pep-0484/#type-comments):

> In some cases, linting tools or other comments may be needed on the same line as a type comment. In these cases, the type comment should be before other comments and linting markers:
>
> ```
> # type: ignore # <comment or other marker>
> ```

zsol:
Thanks for the contribution :)

Issue #636, opened 2022-01-29
SimpleString.raw_value cannot handle """"foo"""


richtier:
A multi-line string can start with a speech mark, but [SimpleString.quote](https://github.com/Instagram/LibCST/blob/c5b073599d9a732bd138b6933a6bf0209ac8e038/libcst/_nodes/expression.py#L648) does not like that and therefore result in `SimpleString.raw_value` raising an exception for perfectly cromulent strings. Take for example these valid strings:

```
>>> """"foo"""  # four "
'"foo'  # evaluates to string starting with one "
```

libcst raises an exception `Invalid string`

These strings are used in the wild. Probably due to typo e.g., [here](https://github.com/ansible/ansible/blob/d19b506ce8c5ee43865b1cead2246fc07cc8902b/lib/ansible/parsing/vault/__init__.py#L761)

zzl0:
@zsol I was looking into this issue and found the bug has already been fixed.

```
$ cat multi_line.py
def foo():
    """"
    foo function
    """
    return 1

$ cat a.py
import sys
import libcst as cst
from libcst.tool import dump

filepath = sys.argv[1]
code = open(filepath).read()

print(dump(cst.parse_module(code)))

$ python a.py multi_line.py
Module(
  body=[
    FunctionDef(
      name=Name(
        value='foo',
      ),
      params=Parameters(),
      body=IndentedBlock(
        body=[
          SimpleStatementLine(
            body=[
              Expr(
                value=SimpleString(
                  value='""""\n    foo function\n    """',
                ),
              ),
            ],
          ),
          SimpleStatementLine(
            body=[
              Return(
                value=Integer(
                  value='1',
                ),
              ),
            ],
          ),
        ],
      ),
    ),
  ],
)
```

zzl0:
@zsol I see, thanks for the quick response and explanation. If no one is working this issue, could I work on it?

zsol:
Absolutely!

PR #704, opened 2022-06-17
expression: fix SimpleString's quote method

zzl0:
## Summary

Fix https://github.com/Instagram/LibCST/issues/636

```
>>> from libcst import *
>>> m = parse_module('''""""foo"""''')
>>> m.body[0].body[0].value.raw_value
  File "C:\Users\zsolz\projects\libcst\libcst\_nodes\expression.py", line 682, in raw_value
    quote_len = len(self.quote)
  File "C:\Users\zsolz\projects\libcst\libcst\_nodes\expression.py", line 666, in quote
    raise Exception("Invalid string {self.value}")
Exception: Invalid string {self.value}
```

## Test Plan

1. Unit test
2. Manual test

```
>>> from libcst import *
>>> m = parse_module('''""""foo"""''')
>>> m.body[0].body[0].value.raw_value
'"foo'
```

### DIFF
--- a/libcst/_nodes/tests/test_simple_string.py
+++ b/libcst/_nodes/tests/test_simple_string.py
@@ -1,3 +1,8 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
 import unittest
 
 import libcst as cst
--- a/libcst/_nodes/tests/test_simple_string.py
+++ b/libcst/_nodes/tests/test_simple_string.py
@@ -1,0 +1,26 @@
+import unittest
+
+import libcst as cst
+
+
+class TestSimpleString(unittest.TestCase):
+    def test_quote(self) -> None:
+        test_cases = [
+            ('"a"', '"'),
+            ("'b'", "'"),
+            ('""', '"'),
+            ("''", "'"),
+            ('"""c"""', '"""'),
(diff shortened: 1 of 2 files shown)


libcst/_nodes/tests/test_simple_string.py
```diff
@@ -0,0 +1,26 @@
+import unittest
```
zsol, line 1: Linter says you need to add a license header here ;)

zsol (approved):
LGTM, thanks! Let's add that license header before merging

zzl0:
Thanks @zsol, @lpetre for reviewing the code.

I noticed that we have code like below ([link](https://github.com/Instagram/LibCST/blob/main/libcst/_nodes/expression.py#L647)). Just curious why does it need type hint? since `quote = ""` is more succinct, and both human and machine can easily infer its type is a string.

```
quote: str = ""
```

bgw:
> Just curious why does it need type hint?

You're probably right, there's probably no reason for it to be there.

zzl0:
@bgw thanks for your reply. Yeah, I also think there's probably no reason for it to be there:
1. It can hurt readability (I have seen some code like below).
```
        a_long_name_name: AVeryLongLongLongLongLongClassName = 
            AVeryLongLongLongLongLongClassName(param1, param2, ...)
```
2. Statically-typed languages (Rust, C++, Java etc) recommends using type inference for this simple initialization case to improve DevX by reducing the ceremony.

PR #709, opened 2022-06-20
Prevent unnecessary work in Scope.get_qualified_names_for_

lpetre:
## Summary

This is a no-op reorganization to prevent unnecessary work. It saves 10s of seconds on very large files

## Test Plan

python -m unittest libcst.metadata.tests.test_name_provider


libcst/metadata/scope_provider.py
```diff
@@ -559,26 +559,26 @@ def f(self) -> "c":
...
+                    node, assignment.node
+                ):
+                    return assignment.get_qualified_names_for(full_name)
+
+        results = set()
         for assignment in assignments:
-            names = assignment.get_qualified_names_for(full_name)
-            if (
-                isinstance(assignment, Assignment)
-                and not isinstance(node, str)
```
lpetre, line 577: this isinstance can be pulled out of the loop

zsol (approved):

PR #710, opened 2022-06-21
Faster qualified name formatting

lpetre:
## Summary
As @zsol pointed out, using f-strings and a conditional is 4x faster 

## Test Plan
python -m unittest libcst.metadata.tests.test_name_provider


zsol (approved):
Gotta resolve that type error but lgtm

PR #720, opened 2022-07-05
Implement lazy loading mechanism for expensive metadata providers

Chenguang-Zhu:
## Summary
- Implement a lazy metadata loading mechanism
- Add a `LazyValue` class, which postpones the loading of some expensive metadata, e.g., qualified names, to the time of calling `get_metadata`
- Modify the implementation of `QualifiedNameProvider` to use this lazy loading mechanism
- Add support for lazy values in metadata matchers
- Modify existing unit tests to use the lazy loading

## Test Plan

With this lazy loading, the running time of the function `QualifiedNameVisitor.on_visit` is supposed to be reduced by 20% on average.

### DIFF
--- a/libcst/_metadata_dependent.py
+++ b/libcst/_metadata_dependent.py
@@ -44,11 +44,11 @@ class LazyValue:
     :func:`~libcst.MetadataDependent.get_metadata`.
     """
 
-    def __init__(self, callable: Callable) -> None:
+    def __init__(self, callable: Callable[[], _T]) -> None:
         self.callable = callable
-        self.return_value = _SENTINEL
+        self.return_value: object = _SENTINEL
 
-    def __call__(self) -> None:
+    def __call__(self) -> object:
         if self.return_value is _SENTINEL:
             self.return_value = self.callable()
         return self.return_value
--- a/libcst/_metadata_dependent.py
+++ b/libcst/_metadata_dependent.py
@@ -34,6 +34,7 @@ _UNDEFINED_DEFAULT = object()
 
 _SENTINEL = object()
 
+
 class LazyValue:
     """
     The class for implementing a lazy metadata loading mechanism that improves the
--- a/libcst/_metadata_dependent.py
+++ b/libcst/_metadata_dependent.py
@@ -11,11 +11,13 @@ from typing import (
     cast,
     ClassVar,
     Collection,
+    Generic,
     Iterator,
     Mapping,
     Type,
     TYPE_CHECKING,
     TypeVar,
+    Union,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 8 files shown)


zsol:
Nice! Looks like there are some test failures, most of them in `libcst.metadata.tests.test_name_provider.QualifiedNameProviderTest`. Can you please fix these and add some new cases that cover the new functionality directly?

libcst/_metadata_dependent.py
```diff
@@ -31,6 +32,26 @@
...
+class LazyValue:
+    """
+    The class for implementing a lazy metadata loading mechanism that improves the
+    performance when retriving expensive metadata (e.g., qualified names). Providers
+    including :class:`~libcst.metadata.QualifiedNameProvider` use this class to load
+    the metadata of a certain node lazily when calling
+    :func:`~libcst.MetadataDependent.get_metadata`.
+    """
+
+    def __init__(self, callable: Callable) -> None:
```
zsol, line 46: This will need to be a bit more strictly typed. `LazyValue` will need to be generic over some `T` value, and this callable will be typed as `Callable[[], T]`

zsol (changes requested):
Let's fix test failures and add some new unit tests that exercise the new functionality directly. There are some type checker failures too, there are [instructions in the readme](https://github.com/Instagram/LibCST/blob/main/README.rst#type-checking) about how you can set this up on your local machine without relying on CI.

libcst/metadata/name_provider.py
```diff
@@ -17,8 +17,10 @@
     ScopeProvider,
 )
 
+_UNDEFINED_DEFAULT = object
 
-class QualifiedNameProvider(BatchableMetadataProvider[Collection[QualifiedName]]):
+
+class QualifiedNameProvider(BatchableMetadataProvider[_UNDEFINED_DEFAULT]):
```
zsol, line 23: Looks great overall, we just gotta figure out this type here. I'll look at it tomorrow a bit more closely

Chenguang-Zhu: Sure, thanks a lot!

Chenguang-Zhu:
> I've fixed the typing issues and added LazyValue support to the base metadata provider in my latest commit. We still need:
> 
> * [x]  unit tests that cover all the new logic in MetadataDependent
> * [x]  unit tests that cover the new logic in BaseMetadataProvider
> * [x]  unit tests that cover the new logic in BaseMatcher

The tests `MatchersMetadataTest.test_lambda_metadata_matcher` and `MatchersMetadataTest.test_lambda_metadata_matcher_with_no_metadata` already cover the new logic via `QualifiedNameProvider`.

zsol (approved):
Nice!

PR #724, opened 2022-07-16
Fix parse error message for number parsing

zzl0:
## Summary

TLDR:
* Fixed a `missing 'f' for f-string` issue.
* Tried to fix the issue by creating a codemod, but my solution has many false positives, so it is not included in this PR code changes.

When I was working on another PR (#704), I encountered a `missing 'f' for f-string` issue ([code link](https://github.com/Instagram/LibCST/pull/704/files#diff-556d83ff3f2718b09556ff7002e044ff03e255423ffe269220a0c316256d2d7bL666)). I was wondering if I can write a codemod command to solve this automatically. Below code is my first attempt to solve this issue, it can find the `missing 'f' for f-string` typos, but it has many false positives and I realized it's not simple problem as I thought. 

This PR is to fix the found `missing 'f' for f-string` issue and also to see if you have tried to solve this issue before.

```python
import re
import difflib

import libcst as cst
import libcst.matchers as m
from libcst.codemod import VisitorBasedCodemodCommand
from libcst.codemod._context import CodemodContext
from libcst.metadata.scope_provider import ScopeProvider

PATTERN = re.compile('\{([^}]+)\}')


class FixFormattedStringCommand(VisitorBasedCodemodCommand):
    DESCRIPTION: str = "Fix f-string."
    METADATA_DEPENDENCIES = (ScopeProvider,)

    def leave_SimpleString(
        self,
        original_node: cst.SimpleString,
        updated_node: cst.SimpleString
    ) -> cst.FormattedString:
        str_val = updated_node.value
        if vars := PATTERN.findall(str_val):
            scope = self.get_metadata(ScopeProvider, original_node)
            if not all(scope.get_qualified_names_for(v) for v in vars):
                return updated_node
            formatted_string = cst.parse_expression(f"f{str_val}")
            return formatted_string
        return updated_node


if __name__ == '__main__':
    py_source = """
def foo() -> str:
    a = "123"
    return "x: {type} = {val}",

def bar() -> str:
    a = "123"
    return "{a}"
"""

    source_tree = cst.parse_module(py_source)
    wrapper = cst.MetadataWrapper(source_tree)

    transformer = FixFormattedStringCommand(CodemodContext())
    modified_tree = wrapper.visit(transformer)

    print(
        "".join(
            difflib.unified_diff(py_source.splitlines(1), modified_tree.code.splitlines(1))
        )
    )
```

Below is a test run:
```
$ python libcst/codemod/commands/fix_fstring.py
---
+++
@@ -5,4 +5,4 @@

 def bar() -> str:
     a = "123"
-    return "{a}"
+    return f"{a}"
```

## Test Plan

👀

### DIFF
--- a/libcst/_parser/conversions/expression.py
+++ b/libcst/_parser/conversions/expression.py
@@ -870,7 +870,7 @@ def convert_atom_basic(
                 Imaginary(child.string), child.whitespace_before
             )
         else:
-            raise Exception("Unparseable number {child.string}")
+            raise Exception(f"Unparseable number {child.string}")
     else:
         raise Exception(f"Logic error, unexpected token {child.type.name}")
 


zsol:
Thanks for fixing this! As for the codemod, I think it doesn't need to call `get_qualified_names_for`, it's enough to just check if `v in scope`. But this sounds more like a linter than a codemod

zzl0:
Thanks @zsol for reviewing this PR. I agree it more like linter, and a PR for pylint tried to add this feature https://github.com/PyCQA/pylint/pull/4787

PR #748, opened 2022-08-16
Update Sphinx to 5.1.1

### LINKED ISSUE #747: Pinned on a heavily outdated Sphinx fork is causing at least 138 errors in the document
While reading libcst document, I notice that there are some confusing discrepancies in the document:

1. The `visit()` method of `libcst.CSTNode` is [shown to be a zero-argument function](https://libcst.readthedocs.io/en/latest/nodes.html#libcst.CSTNode.visit). However the following doc immediately states that it needs a visitor instance to operate on. An acute reader has to go to read the source code to figure out its precise usage.
2. The `libcst.helpers.ensure_type()` is [shown to be a zero-argument function](https://libcst.readthedocs.io/en/latest/helpers.html#libcst.helpers.ensure_type), which is definitely confusing.
3. The `findall/extract/extractall/replace` functions in the `libcst.matchers` module are [all shown to be zero-argument functions](https://libcst.readthedocs.io/en/latest/matchers.html#functions). It's not being helpful for readers, if not misleading.

There are numerous other discrepancies in the documents. For a full list, consult the warning output of Sphinx build in the latest GitHub Action run [here](https://github.com/Instagram/LibCST/runs/7726595193?check_suite_focus=true#step:7:36). A quick counting shows that there are at least 138 errors reported in the document.

I investigated the problem intensively. Turn out it has to do with how static type feature in CPython had some substantial and deeply-refactored updates on the advent of 3.9, which broke Sphinx due to its reliance on some CPython internal details. It had impacted and been reported on some other projects using Sphinx.

It's a little complicated to explain all the details, but the good news is, we don't have to understand the details to fix our document. **Sphinx already fixed itself on the upstream, so what we have to do, is to just update the version of Sphinx** we use to build our document.

It turns out that the version of Sphinx which LibCST [uses](https://github.com/Instagram/LibCST/blob/977504f10407b3d94bbe45157f9286709dd74754/requirements-dev.txt#L5) to build document, is an old [personal fork](https://github.com/jimmylai/sphinx), which is heavily outdated, left intact for three years, shown to be 6057 commits behind the upstream Sphinx! That reasonably explains why we are seeing all these confusing errors which should have been fixed by Sphinx around the advent of Python 3.9.

MapleCCC:
This pull request should fix #747.

I have tried to build document of LibCST with Sphinx 5.1.1 locally. It works fine and smoothly. The only trivial change we need to make to accommodate is to specify the `language` setting to non-`None` value in `conf.py`.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -71,7 +71,7 @@ master_doc = "index"
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,7 +2,7 @@ black==22.6.0
 coverage>=4.5.4
 fixit==0.1.1
 flake8>=3.7.8,<5
-git+https://github.com/jimmylai/sphinx.git@slots_type_annotation
+Sphinx>=5.1.1
 hypothesis>=4.36.0
 hypothesmith>=0.0.4
 jupyter>=1.0.0


MapleCCC:
The CI encountered errors when installing the `lxml` library in the environment of CPython 3.11 + Windows. This is due to that `lxml` library has not published built wheels for the environment of CPython 3.11 + Windows. Building `lxml` locally from scratch has always been considered a hard task.

I believe we are safe to ignore these CI errors. The PR is now ready to merge, whenever you feel convenient.

PR #756, opened 2022-08-26
Fix bug when `TypeOf` is one of options in `OneOf` / `AllOf`

### LINKED ISSUE #528: `OneOf` causes match to fail in some cases
The version of libcst I am using is 0.3.21. Following is the code snippet to reproduce the bug:

```python
import libcst as cst
import libcst.matchers as m

s = """
def f():
    pass
"""

module = cst.parse_module(s)

pattern_1 = (m.FunctionDef | m.Lambda)(m.Name("f"))
pattern_2 = m.FunctionDef(m.Name("g"))
pattern_3 = m.OneOf(pattern_1, pattern_2)

print(len(m.findall(module, pattern_1)))   # 1
print(len(m.findall(module, pattern_2)))   # 0
print(len(m.findall(module, pattern_3)))   # 0
```

As the result shows, pattern_1 has match, and pattern_2 has no match. pattern_3 is supposed to have match when it combines pattern_1 and pattern_2 with `OneOf`.

MapleCCC:
This PR should fix #528.

I also added unit test, to prevent regression.

### DIFF
--- a/libcst/matchers/_matcher_base.py
+++ b/libcst/matchers/_matcher_base.py
@@ -226,7 +226,7 @@ class OneOf(Generic[_MatcherT], BaseMatcherNode):
         for option in options:
             if isinstance(option, AllOf):
                 raise Exception("Cannot use AllOf and OneOf in combination!")
-            elif isinstance(option, OneOf):
+            elif isinstance(option, (OneOf, TypeOf)):
                 actual_options.extend(option.options)
             else:
                 actual_options.append(option)
--- a/libcst/matchers/_matcher_base.py
+++ b/libcst/matchers/_matcher_base.py
@@ -302,6 +302,8 @@ class AllOf(Generic[_MatcherT], BaseMatcherNode):
         for option in options:
             if isinstance(option, OneOf):
                 raise Exception("Cannot use AllOf and OneOf in combination!")
+            elif isinstance(option, TypeOf):
+                raise Exception("Cannot use AllOf and TypeOf in combination!")
             elif isinstance(option, AllOf):
                 actual_options.extend(option.options)
             else:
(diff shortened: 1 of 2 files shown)


MapleCCC:
The reason I choose to disallow combination of `AllOf` and `TypeOf`, is because I think this is analogous to how the combination of `AllOf` and `OneOf` is disallowed. It seems natural to think of `TypeOf` and `OneOf` having similar semantic, although on different levels. Please feel free to let me know if you prefer other design.

MapleCCC:
Static type checkers didn't catch this bug for us. I wonder if we can do something in our code to help type checkers catch these bugs for us.

The logic in the `matchers` sub-system are large and complicated, making it hard for human developers to manually detect all kinds of logic bugs by eyes. It would be a huge gain if we manage to have type checkers help us here.

zsol:
Thanks! I agree with your sentiment, but don't see a way to express these kinds of constraints using Python's type system atm. Tests are our only tool here unfortunately

MapleCCC:
Thank you for the comment.

I am more optimistic on this though. I have a hunch that we can greatly improve type-check-ability here if we designate some of the types in `libcst.matchers` package to be so-called `sealed type` [[1]](https://docs.oracle.com/en/java/javase/15/language/sealed-classes-and-interfaces.html). 

This feature is actively discussed in Python typing community recently [[2]](https://mail.python.org/archives/list/typing-sig@python.org/thread/AKXUBJUUHBBKTLNIAFCA6HII5QQA2WFX/). Current workaround seems to be to use union type (a.k.a. sum type) to simulate sealness.

I will play with this idea a bit more, and try to see whether it brings outweighing benefits here.

PR #757, opened 2022-08-27
Raise informative exception when metadata is unresolved in a metadata-based match

### LINKED ISSUE #526: The example code in document of `MatchMetadataIfTrue` doesn't work as expected
The document of `MatchMetadataIfTrue` [suggests an example code](https://libcst.readthedocs.io/en/latest/matchers.html#libcst.matchers.MatchMetadataIfTrue) to "match against any `Name` node for the identifier `foo` as long as that identifier is found at the beginning of an unindented line". I tried to reproduce the example code:

```python
import libcst as cst
import libcst.matchers as m
from libcst.metadata import PositionProvider

source = """
foo
"""

module = cst.parse_module(source)

names = m.findall(
    module,
    m.Name(
        value="foo",
        metadata=m.MatchMetadataIfTrue(
            PositionProvider, lambda position: position.start.column == 0
        ),
    ),
)

print(names)  # this statement prints an empty list, which it should not
```

I follow the example code verbatim, yet the result is not as expected. The `names` list should contain one `cst.Name` node, yet it's empty. The source code of LibCST is so voluminous that I can't figure out where the bug originates. I am presenting the bug here, so that someone familiar with the LibCST codebase can kindly take a look.

MapleCCC:
I also tuned the related unit test to accommodate this change.

This PR closes #526.

### DIFF
--- a/libcst/matchers/_matcher_base.py
+++ b/libcst/matchers/_matcher_base.py
@@ -581,8 +581,11 @@ class MatchMetadata(_BaseMetadataMatcher):
     """
     Matcher that looks up the metadata on the current node using the provided
     metadata provider and compares the value on the node against the value provided
-    to :class:`MatchMetadata`. If the metadata value does not exist for a particular
-    node, :class:`MatchMetadata` will always be considered not a match.
+    to :class:`MatchMetadata`.
+    If the metadata provider is unresolved, a :class:`LookupError` exeption will be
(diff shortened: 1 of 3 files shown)


zsol:
The change looks great, thanks! Mind adding a quick note about this exception to the docs?

MapleCCC:
Okay, in the new commit I documented this behavior to `MatchMetadata` and `MatchMetadataIfTrue`. 📚

zsol:
Nice :)

PR #760, opened 2022-08-29
Tighten the metadata type of `ExpressionContextProvider`

### LINKED ISSUE #758: The metadata type of `ExpressionContextProvider` should be `ExpressionContext` instead of `Optional[ExpressionContext]`
Looking at the [implementation code](https://github.com/Instagram/LibCST/blob/main/libcst/metadata/expression_context_provider.py) of `ExpressionContextProvider`, we can conclude that every invocation of `set_metadata()` inserts an `ExpressionContext` object into the metadata table. It never once inserts a `None` object.

Therefore we are safe to narrow down the metadata type for it. I can imagine a more precise type being beneficial for downstream code, such as reducing unnecessary `is not None` checks everywhere.

MapleCCC:
This PR closes #758.

### DIFF
--- a/libcst/metadata/expression_context_provider.py
+++ b/libcst/metadata/expression_context_provider.py
@@ -201,7 +201,7 @@ class ExpressionContextVisitor(cst.CSTVisitor):
         return False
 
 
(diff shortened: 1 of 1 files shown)


zsol:
Thanks!

PR #763, opened 2022-09-04
Use the correct doc build status badge

MapleCCC:
We have been showing the wrong badge, which shows the build status of the `pip` library, instead of the `LibCST` library.


zsol:
Haha nice, well spotted :)

PR #769, opened 2022-09-06
Fix black configuration

MapleCCC:


MapleCCC:
The CI error is due to a unit test `test_codemod_formatter_error_input` trying to test that, when `libcst.codemod` is run on a file containing `async/await` reserved keywords, it's expected to always fail trying to apply the black formatter on the file.

The formatter error turns out to be a total unfortunate incident, relying on black incidentally picking up the target version configuration (i.e., "py36") in our own `pyproject.toml` simply because it happens to be run in our root folder.

This is so horrible in many ways! I think we can all agree on that this unit test is problematic, and needs some proper fix ?

MapleCCC:
For the record, another PR #771 has been opened to address the CI errors we see here.

MapleCCC:
Now I am going to merge #771 into us here, so that we can have an all-green CI pass.

MapleCCC:
I believe the CI error is GitHub's fault, not ours :-)

PR #770, opened 2022-09-06
Several trivial refactors

MapleCCC:
Summary of changes:

1. Remove the unnecessary overhead of copying singleton objects.
2. Avoid generating unnecessary `pass` statements.
3. Remove the unnecessary cost of building intermediate lists.
4. Remove an unused import, an overlook from commit 8e6bf9e9.
5. Simplify an invocation of `Mapping.get()`.
6. Just use unittest's `assertRaises()` to specify expected exception types, instead of catching every possible `Exception`s, which could suppress legitimate errors and hide bugs
7. Several other small refactors.


zsol (approved):
Thanks!

PR #771, opened 2022-09-06
Fix the bug that the use of formatter in codemods has undetermined target Python version, resulting in hard-to-reason-with behavior

MapleCCC:
This PR tries to fix the bug manifesting in #769. In summary, we specify that the black formatter should use the same target Python version which libcst itself uses.


MapleCCC:
The CI errors are because the CI images fail to download some Rust crates due to spurious network issue. Not our fault, I guess.

zsol (approved):
Thank you!

PR #807, opened 2022-11-08
Add AccessorProvider

matthewshaer:
Adding a provider which can tell what accessor to use to go from the parent to that child node

## Summary
I want to be able to see what accessor needs to be used to go from the parent to the given child

## Test Plan
Added unit tests

### DIFF
--- a/libcst/metadata/tests/test_accessor_provider.py
+++ b/libcst/metadata/tests/test_accessor_provider.py
@@ -1,0 +1,68 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import dataclasses
+
+from textwrap import dedent
+
+import libcst as cst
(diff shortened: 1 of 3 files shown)


libcst/metadata/tests/test_accessor_provider.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, *, test: UnitTest) -> None:
+        self.test = test
+
+    def on_visit(self, node: cst.CSTNode) -> bool:
+        for f in dataclasses.fields(node):
+            child = getattr(node, f.name)
+            if type(child) is cst.CSTNode:
+                accessor = self.get_metadata(AccessorProvider, child)
+                self.test.assertEqual(accessor, f.name)
```
zsol, line 26: This is not a very valuable test as it just duplicates the core logic. I suppose that's true for the parent metadata provider tests too

zsol: > Should I serialise something like a tree of accessors and confirm it gets the right serialization ?

If you have the time, that would be great. No big deal if not.

PR #816, opened 2022-11-10
Add py3.11 classifier

and-semakin:
## Summary

Keep PyPI classifiers in sync with actually supported Python versions :)


and-semakin:
Just signed CLA

PR #841, opened 2022-12-30
[AddImportsVisitor] Docstring Check Only for the Top Element of the Body

sagarbadiyani:
## Summary
Refer https://github.com/Instagram/LibCST/pull/343#issuecomment-1367622138
Instead of checking for docstring for all the elements of the body, this PR only keeps the check for the 0th element since our objective is to only honor the module docstrings.

The reason I am asking for this, I found a case in one of the repositories where the docstrings were written outside the function like how we have in Java, obviously, this is unconventional and unpythonic, but my only concern is why are we checking for docstrings which are not module docstrings?

Consider a case
```
from __future__ import annotations
from __future__ import division

import abc
from xyz import def

"""Some docstring"""
def some_func():
    pass
```    
Let's say we are trying to add an import `from typing import List`, the final code after applying the add imports tranformation will look like
```
from __future__ import annotations
from typing import List
from __future__ import division

import abc
from xyz import def

"""Some docstring"""
def some_func():
    pass
```
This is because the docstring is a part of the body of the module and when the for loop reached there, it set the import_add_location to 1, which should not have been the case
And this will break the code because all the `from __future__ imports` must occur at the beginning of the file

## Test Plan
Added a Testcase.

### DIFF
--- a/libcst/codemod/visitors/tests/test_add_imports.py
+++ b/libcst/codemod/visitors/tests/test_add_imports.py
@@ -892,8 +892,8 @@ class TestAddImportsCodemod(CodemodTest):
             ),
         )
 
-    def test_import_in_docstring_module_with_docstring_not_as_the_top_line(
-        self
+    def test_import_in_module_with_standalone_string_not_a_docstring(
+        self,
     ) -> None:
         """
         The import should be added after the __future__ imports.
--- a/libcst/codemod/visitors/tests/test_add_imports.py
+++ b/libcst/codemod/visitors/tests/test_add_imports.py
@@ -891,3 +891,36 @@ class TestAddImportsCodemod(CodemodTest):
                 full_module_name="a.b.foobar", full_package_name="a.b"
             ),
         )
+
+    def test_import_in_docstring_module_with_docstring_not_as_the_top_line(
+        self
+    ) -> None:
+        """
+        The import should be added after the __future__ imports.
+        """
+        before = """
+            from __future__ import annotations
+            from __future__ import division
+
+            '''docstring.'''
+            def func():
+                pass
+        """
+        after = """
+            from __future__ import annotations
+            from __future__ import division
+            import typing
+
+            '''docstring.'''
+            def func():
+                pass
+        """
+
+        self.assertCodemod(
+            before,
+            after,
+            [ImportItem("typing", None, None)],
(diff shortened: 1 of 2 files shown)


sagarbadiyani:
Hey @zsol can you please take a look?

sagarbadiyani:
@jimmylai @carljm would be great if you could take a look at this one :)

libcst/codemod/visitors/tests/test_add_imports.py
```diff
@@ -891,3 +891,36 @@ def test_import_in_docstring_module(self) -> None:
                 full_module_name="a.b.foobar", full_package_name="a.b"
             ),
         )
+
+    def test_import_in_docstring_module_with_docstring_not_as_the_top_line(
```
carljm, line 895: nit: A "docstring" in the wrong location is not really a docstring at all, technically speaking. It's just a random string expression in a statement by itself, which is thrown away at runtime. So I would prefer to rename this test to `test_import_in_module_with_standalone_string_not_a_docstring`.

carljm (changes requested):
This looks like a correct fix to me. Thanks for adding a test!

One test-naming nit, and it looks like CI is not fully happy.

libcst/codemod/visitors/tests/test_add_imports.py
sagarbadiyani: Agreed, thank you

sagarbadiyani: Fixed

sagarbadiyani:
Sorry for bothering again @carljm but any pointers on how can I fix the failing CI checks?
Running `python setup.py test` locally did not help

carljm:
> any pointers on how can I fix the failing CI checks?

It looks to me like an infra failure unrelated to this PR.

carljm (approved):
This looks good to me. I think the failing checks are an infra problem that isn't related to this diff.

I haven't been an active maintainer of LibCST for quite a while so I'd prefer to wait a bit and see if someone else can verify that and merge this. If nobody weighs in in the next few days I can probably merge, I think the fix is clearly correct.

sagarbadiyani:
Great, thank you very much!

zsol:
Thanks for this @sagarbadiyani!

PR #861, opened 2023-02-15
Fix Github issue 855 - fail to parse with statement

### LINKED ISSUE #855: Fails to parse parenthesized `with` statement without trailing comma
#584 added parenthesized `with` statements to the native parser, but they’re only accepted when a trailing comma is present. Python accepts them with or without a trailing comma.

```console
$ pip freeze
libcst==0.4.9
mypy-extensions==0.4.3
PyYAML==6.0
typing-inspect==0.8.0
typing_extensions==4.4.0

$ export LIBCST_PARSER_TYPE=native

$ python3 -c 'import libcst; libcst.parse_statement("with (a as a, b as b,): pass")'

$ python3 -c 'import libcst; libcst.parse_statement("with (a as a, b as b): pass")'
  File "<string>", line 1, in <module>
  File "/tmp/v/lib/python3.10/site-packages/libcst/_parser/entrypoints.py", line 138, in parse_statement
    result = _parse(
  File "/tmp/v/lib/python3.10/site-packages/libcst/_parser/entrypoints.py", line 55, in _parse
    return parse(source_str)
libcst._exceptions.ParserSyntaxError: Syntax Error @ 1:1.
parser error: error at 1:20: expected *

with (a as a, b as b): pass
^
```

Cc @stroxler

stroxler:
Fixes #855 

## Summary

When we added support for parenthesized with statements, the grammar on the with itself was correct (it's a right and left parenthesis around a comma-separated list of with-items, with a possible trailing comma).

But inside of the "as" variation of the with_item rule we have a peek at the next character, which was allowing for a comma or a colon. That peek needs to also accept right parentheses - otherwise, if the last item contains an `as` and has no trailing comma we fail to parse.

The bug is exercisecd by, for example, this code snippet:
```
with (foo, bar as bar,):
    pass
```

The with_wickedness test fixture has been revised to include both the plain and async variations of this example snippet with and without trailing comma, and tests pass after the peek rule fix.

## Test Plan

```
cd native
cargo test
```

### DIFF
--- a/native/libcst/src/parser/grammar.rs
+++ b/native/libcst/src/parser/grammar.rs
@@ -516,7 +516,7 @@ parser! {
             }
 
         rule with_item() -> WithItem<'input, 'a>
-            = e:expression() a:lit("as") t:star_target() &(lit(",") / lit(":")) {
+            = e:expression() a:lit("as") t:star_target() &(lit(",") / lit(":") / rpar()) {
                 make_with_item(e, Some(a), Some(t))
             }
             / e:expression() {
--- a/native/libcst/tests/fixtures/with_wickedness.py
+++ b/native/libcst/tests/fixtures/with_wickedness.py
@@ -1,13 +1,34 @@
 # with_wickedness
 
-with foo : ...
+with foo :
+    pass
+
+with foo, bar as bar:
+    pass
+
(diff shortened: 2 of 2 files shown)


zsol (approved):
amazing :)

Issue #722, opened 2022-07-11
`ValueError` when using `FullyQualifiedNameProvider`


zwade:
Hi!

I'm getting the following `ValueError` when using the `FullyQualifiedNameProvider` in my Transformer. 

```
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nuvo/bin/cst", line 8, in <module>
    sys.exit(main())
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nuvo/lib/python3.10/site-packages/sync_to_async/cli.py", line 5, in main
    tool_main("cdm", sys.argv[1:])
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nuvo/lib/python3.10/site-packages/libcst/tool.py", line 835, in main
    return lookup.get(args.action or None, _invalid_command)(proc_name, command_args)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nuvo/lib/python3.10/site-packages/libcst/tool.py", line 572, in _codemod_impl
    result = parallel_exec_transform_with_prettyprint(
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nuvo/lib/python3.10/site-packages/libcst/codemod/_cli.py", line 588, in parallel_exec_transform_with_prettyprint
    metadata_manager.resolve_cache()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nuvo/lib/python3.10/site-packages/libcst/metadata/full_repo_manager.py", line 66, in resolve_cache
    cache[provider] = handler(
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nuvo/lib/python3.10/site-packages/libcst/metadata/name_provider.py", line 115, in gen_cache
    cache = {path: calculate_module_and_package(".", path) for path in paths}
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nuvo/lib/python3.10/site-packages/libcst/metadata/name_provider.py", line 115, in <dictcomp>
    cache = {path: calculate_module_and_package(".", path) for path in paths}
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nuvo/lib/python3.10/site-packages/libcst/helpers/module.py", line 136, in calculate_module_and_package
    relative_filename = PurePath(filename).relative_to(repo_root)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nuvo/lib/python3.10/pathlib.py", line 816, in relative_to
[...]

PR #867, opened 2023-02-18
Pass root path to FullyQualifiedNameProvider

amyreese:
## Summary

This allows FullyQualifiedNameProvider to work with absolute paths,
rather than assuming all paths given will be relative to the current
directory. This enables tools like Fixit to provide a root path, and
have the FullyQualifiedNameProvider correctly scope the final results
relative to that root path.

This does require that both the root path and the given file paths
match the other as relative or absolute, due to the
`calculate_module_and_package` helper comparing file paths relative
to the root path, but this seems like a reasonable tradeoff, and
unlikely to cause a problem in normal use cases.

## Test Plan

Updated test cases. Works with in-progress Fixit2 changes.

### DIFF
--- a/libcst/_types.py
+++ b/libcst/_types.py
@@ -4,7 +4,8 @@
 # LICENSE file in the root directory of this source tree.
 
 
-from typing import TYPE_CHECKING, TypeVar
+from pathlib import PurePath
+from typing import TYPE_CHECKING, TypeVar, Union
 
 if TYPE_CHECKING:
     from libcst._nodes.base import CSTNode  # noqa: F401
@@ -12,3 +13,4 @@ if TYPE_CHECKING:
 
 CSTNodeT = TypeVar("CSTNodeT", bound="CSTNode")
 CSTNodeT_co = TypeVar("CSTNodeT_co", bound="CSTNode", covariant=True)
+StrPath = Union[str, PurePath]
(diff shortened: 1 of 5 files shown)


davidshepherd7:
FWIW I also ran into this bug and tried the same fix (passing `root_path` instead of `"."` in `gen_cache`, and this fix allowed me to use `FullyQualifiedNameProvider` as a `METADATA_DEPENDENCIES` in a codemod where before it was broken.

I was just about to start writing a PR for the same thing when I saw this one! :tada: :heart_eyes: 

Also: fixes #722  (someone else there had the same issue and the same solution).

libcst/_types.py
```diff
@@ -4,11 +4,13 @@
...
+from pathlib import Path
+from typing import TYPE_CHECKING, TypeVar, Union
 
 if TYPE_CHECKING:
     from libcst._nodes.base import CSTNode  # noqa: F401
 
 
 CSTNodeT = TypeVar("CSTNodeT", bound="CSTNode")
 CSTNodeT_co = TypeVar("CSTNodeT_co", bound="CSTNode", covariant=True)
+StrPath = Union[str, Path]
```
zsol, line 16: Would `os.PathLike` suffice?

zsol (approved):

libcst/_types.py
amyreese: Turns out that `os.PathLike` behaves differently (and mutually incompatible from a typing perspective) between 3.8 and 3.9. So I went ahead and reverted back to pathlib, though using `PurePath` in hopes of covering more potential usage. We can revisit in the future if we drop support for older runtimes, or if there's a better alternative available.

zsol: TIL

PR #868, opened 2023-02-18
Support PEP 604 in ApplyTypeAnnotationsVisitor

hauntsaninja:


hauntsaninja:
The CI failures on Python 3.11 are unrelated: `ModuleNotFoundError: No module named 'setuptools_rust'`

zsol (approved):
Sorry this took a while to get to. LGTM but let me ping @stroxler in case he wants to take a look too

libcst/codemod/visitors/tests/test_apply_type_annotations.py
```diff
@@ -1119,6 +1119,33 @@ def test_annotate_functions_with_existing_annotations(
...
+    def test_annotate_functions_pep_604(
+        self, stub: str, before: str, after: str
+    ) -> None:
+        self.run_test_case_with_flags(
+            stub=stub,
+            before=before,
+            after=after,
+            overwrite_existing_annotations=True,
+        )
+
```
stroxler, line 1148: It might be worth adding a test showing the import collector works (that's where a lot of the new logic lives, and I don't think we have anything exercising that code yet)

stroxler (approved):
LGTM other than I think it may be worth adding one more test.

Thanks for this!

zsol:
@hauntsaninja do you think you'll have time to add an extra test?

hauntsaninja:
Thanks for the suggestion! I added a test, then noticed that the collectors don't reach all the way into e.g. `Callable[[XXX], int]` because of the List.

Instead of effectively writing my own visitor to get all the nodes, I decided to just use the LibCST visiting. So in `ImportedSymbolCollector` we now just store a boolean to keep track of whether we're in an annotation or not. `TypeCollector` was a little more complicated since it was transforming the tree, so I just made a CSTTransformer.

I think the net result is overall a little simpler, but ended up being a larger change than I was expecting to make.

hauntsaninja:
cc @carljm

zsol (approved):
Amazing, thanks!

PR #869, opened 2023-02-18
Fix generating invalid trailing commas in import statements

### LINKED ISSUE #532: ImportFrom can produce invalid code without parenthesis
In fact this is captured as a test case in https://github.com/instagram/libcst/blob/f36eacb1327830fad51e08084b46601cad681ae0/libcst/_nodes/tests/test_import.py#L394-L401

```
cst.ImportFrom(
    module=cst.Name("foo"),
    names=(
        cst.ImportAlias(cst.Name("bar"), comma=cst.Comma()),
        cst.ImportAlias(cst.Name("baz"), comma=cst.Comma()),
    ),
)
```

This node generates `from foo import bar,baz,` which isn't syntactically correct.

davidshepherd7:
## Summary

Fixes #532

## Test Plan

Unit tests added/tweaked for this case, also the remove imports codemod tests serve as an integration test.

### DIFF
--- a/libcst/_nodes/statement.py
+++ b/libcst/_nodes/statement.py
@@ -1172,18 +1172,24 @@ class ImportAlias(CSTNode):
             comma=visit_sentinel(self, "comma", self.comma, visitor),
         )
 
-    def _codegen_impl(self, state: CodegenState, default_comma: bool = False) -> None:
+    def _codegen_impl(
+        self,
+        state: CodegenState,
+        default_comma: bool = False,
+        comma_is_valid: bool = True,
+    ) -> None:
         with state.record_syntactic_position(self):
             self.name._codegen(state)
             asname = self.asname
             if asname is not None:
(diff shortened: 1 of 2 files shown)


zsol:
Thanks for your contribution! This all looks good to me. The logic will also need to be ported to https://github.com/Instagram/LibCST/blob/46509dd5e119ab3df7b53ef4eb281bde72a30409/native/libcst/src/nodes/statement.rs#L782 before merging.

davidshepherd7:
> Hello. Something I can do to speed up this?

Yeah, I don't really have time for open source work these days so if someone can figure out and implement the missing rust part that would speed things up a lot.

sureshjoshi:
Are the Rust and Python implementations intended to be kept in lock-step?

I just started using libcst and ran into this bug, but it looks like it was "partially" solved already?

PR #874, opened 2023-02-22
Use new setup-python caching actions

amyreese:
## Summary

With the latest setup-python actions, there is a better caching
mechanism available that also requires less setup, and provides better
fallback behavior that should help avoid the random CI failures that
have been happening on 3.11 for setuptools-rust. This ensures that we
install the necessary dependencies before attempting to build the
package or run tests, while still enabling speedups in best case
scenario when requirements files haven't changed.

See the upstream readme for details:
https://github.com/actions/setup-python#caching-packages-dependencies

## Test Plan

CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,13 +43,9 @@ jobs:
       - uses: actions/setup-python@v4
         if: ${{ !contains(matrix.os, 'self-hosted') }}
         with:
+          cache: pip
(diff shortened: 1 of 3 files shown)


amyreese:
@zsol any thoughts? The previous caching solution keeps breaking on 3.11, and it struck again on #887

zsol (approved):
Fantastic :) Thx

PR #879, opened 2023-02-28
Allow running codemods without configuring in YAML

akx:
## Summary

This enables codemodding things by just plonking a CodemodCommand class into any old importable module and running `python -m libcst.tool codemod -x some_module.SomeClass ...`

### DIFF
--- a/libcst/tool.py
+++ b/libcst/tool.py
@@ -395,14 +395,10 @@ def _codemod_impl(proc_name: str, command_args: List[str]) -> int:  # noqa: C901
 
     # Now, try to load the class and get its arguments for help purposes.
     if args.command is not None:
-        command_path = args.command.split(".")
-        if len(command_path) < 2:
+        command_module_name, _, command_class_name = args.command.rpartition(".")
+        if not (command_module_name and command_class_name):
             print(f"{args.command} is not a valid codemod command", file=sys.stderr)
             return 1
-        command_module_name, command_class_name = (
-            ".".join(command_path[:-1]),
-            command_path[-1],
-        )
         command_class = None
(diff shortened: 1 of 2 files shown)


povilasb:
👍

akx:
@zsol How about a flag to _allow_ the fallback behavior? 

```
python -m libcst.tool codemod -x some_module.SomeClass
```
(`-x` for eXternal)

akx:
@zsol Yeah, good point – I was about to edit my comment to a "never mind" because I realized that was silly after all, but you got there first 😁 

Anyway, um... _are_ there tests for the CLI at all? `test_tool.py` seems to have unrelated things in it...

libcst/tool.py
```diff
@@ -391,38 +391,49 @@ def _codemod_impl(proc_name: str, command_args: List[str]) -> int:  # noqa: C901
...
-                file=sys.stderr,
+        if args.external:
+            # There's no error handling here on purpose; if the user opted in for `-x`,
+            # they'll probably want to see the exact import error too.
+            command_class = getattr(
+                importlib.import_module(command_module_name),
+                command_class_name,
             )
-            return 1
+        else:
```
zsol, line 416: Nit: this else block is not needed, we can just dedent the entire block since we always return in the if branch

zsol (approved):
Looks great, thanks!

libcst/tool.py
akx: Um... no, we don't `return` from the `if args.external` branch?

zsol: Oh you're absolutely right 😂

zsol:
Thanks!

PR #887, opened 2023-03-06
Attach a supplied Path when parsing a module

amyreese:
## Summary

This adds an optional `path` parameter to `parse_module`, and then
attaches that path object to the final `Module` object after parsing.

This allows frameworks like µsort or Fixit to supply the path being
parsed, so that the path can be later accessed as a property of the
module, without needing to track or pass that through a stack via
alternative means. Visitors can then fetch this path information
from the root of the tree when needed, rather than needing separate
metadata providers or state information.

## Test Plan

Regenerated files, updated test cases to show the path being tracked.


zsol:
- The new field will need to be added here as well: https://github.com/Instagram/LibCST/blob/f9536b522f58d2b70ae8beb6b607b45ef08620e6/native/libcst/src/nodes/module.rs#L35
- I think a fully qualified name would make more sense than a path (which has some undesirable semantics coupled to filesystems - absolute/relative/symlinks/`..`/extensions)

Having said that, this property of a `Module` isn't really part of the CST: two CSTs are currently considered equal even if they're in separate files. That won't be true anymore after this PR.

What's the reason metadata providers are not good enough for µsort or Fixit?

amyreese:
> What's the reason metadata providers are not good enough for µsort or Fixit?

Internally, tools like Pyre need absolute path on disk in order to fully resolve the buck dependency graph, etc. AFAIK, none of the LibCST metadata providers offer that information, and I'm pretty sure they can't, based on my understanding of how LibCST is used, though I would love to be wrong about that.

Is there another way that Fixit can inject this information somewhere that a CSTLintRule can get it when needed, without having to use Fixit-specific global state or API?

> Having said that, this property of a Module isn't really part of the CST: two CSTs are currently considered equal even if they're in separate files. That won't be true anymore after this PR.

Noted. That can be resolved by using `dataclasses.field(compare=False)`.

amyreese:
Turns out, yes, that does work, and I don't know why my brain wasn't able to wrap around how that works until I stepped through the end-to-end logic. I have strong opinions about LibCST's endless depth of structure and inheritence, and they're not kind opinions. 😅

zsol:
Happy to make something simpler that achieves the same goals :)

amyreese:
#892 contains the metadata provider approach to this problem.

PR #893, opened 2023-03-14
PEP 621 + hatch to run tests/lint/etc

amyreese:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #893

Moves PEP 621 metadata from `setup.py` and `requirements*.txt` into the
`[project]` table of `pyproject.toml`. This enables using hatch as a
task runner for the project, where previously one would need to remember
a bunch of different commands, or repeatedly consult the readme's
developer guide to find all of the relevant commands.

This creates the following hatch commands:

- docs
- fixtures
- format
- lint
- test
- typecheck

It also updates all of the github actions workflows to use the
appropriate hatch commands, and the readme's developer guide, so that
there is only one source of truth for what constitutes running tests.

The "test" workflows now drop the matrix distinction between "pure" or
"native", and run tests in both modes from a single build.


zsol:
Thanks for this, it's been on my TODO for a long time :)

Looks like a test dep is missing on Windows

.github/workflows/ci.yml
```diff
@@ -15,91 +15,89 @@ jobs:
...
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
-        parser: [pure, native]
         exclude:
           # skip these builds until https://bugs.launchpad.net/lxml/+bug/1977998 is resolved
           - os: windows-latest
             python-version: "3.11"
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
zsol, line 25: Do we need this for these jobs?

pyproject.toml
```diff
@@ -1,12 +1,85 @@
...
+fixtures = [
+    "python scripts/regenerate-fixtures.py",
+    "git diff --exit-code",
+]
+format = "ufmt format libcst scripts"
+lint = [
+    "flake8 libcst",
+    "ufmt check libcst scripts",
+    "python -m slotscheck libcst",
+    "./check_copyright.sh",
```
zsol, line 73: this won't work on windows :(

zsol (approved):
does cibuildwheel work with hatch? You could test by stacking a PR like #890 on top

.github/workflows/ci.yml
amyreese: Without them, hatch fails to create the environment because setuptools_scm shows the version as `0.1+gitsomething`, but the dev tools like usort depend on `libcst>=0.3.7`, creating a version mismatch. Not sure if there's a better fix, but this at least makes it work, and the repo isn't so big that I feel guilty doing it 😅

pyproject.toml
amyreese: Yeah, I'll copy over the script that I made for fixit and get that working.

PR #894, opened 2023-03-14
Fix: relative imports from '' package are not allowed

podtserkovskiy:
## Summary
When we resolve a relative import from root package "", we get the import prefixed by "."
However, this kind of imports are not allowed and cpython throws an error:
```
/Users/michaelpo/PycharmProjects/pythonProject/venv/bin/python /Users/michaelpo/PycharmProjects/pythonProject/main.py 
  File "/Users/michaelpo/PycharmProjects/pythonProject/main.py", line 2, in <module>
    from .l1 import l1_func
ImportError: attempted relative import with no known parent package

Process finished with exit code 1
```

## Test Plan
Unit tests

### DIFF
--- a/libcst/helpers/module.py
+++ b/libcst/helpers/module.py
@@ -90,7 +90,7 @@ def get_absolute_module_from_package(
     if num_dots == 0:
         # This is an absolute import, so the module is correct.
         return module_name
-    if current_package is None:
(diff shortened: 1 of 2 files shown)


amyreese (approved):

amyreese:
thank you!

PR #898, opened 2023-03-23
Ensure current Python interpreter is used for subprocesses

akx:
One spot used `python`, the other used `python3` – `sys.executable` is known to be correct.


zsol:
Looks like you'll need to rebase this pr, it's based on a commit from 2020

akx:
> Looks like you'll need to rebase this pr, it's based on a commit from 2020

Note to self: check where your `origin/master` points at before branching off it.

zsol:
To be fair, github doesn't exactly make this obvious

zsol:
Thanks!

PR #903, opened 2023-03-31
Fix ApplyTypeAnnotationsVisitor behavior on attribute assignments.

rchen152:
* Fixes an issue where ApplyTypeAnnotationsVisitor would crash on code like `SomeClass.some_attribute = 42` with a "Name is not a valid identifier" error message.
* Changes the above-mentioned error message to include the bad name in the message, for easier debugging.
* Adds tests for all valid assignment targets, as described here: https://libcst.readthedocs.io/en/latest/nodes.html#libcst.BaseAssignTargetExpression.

Context: I'm trying to fix https://github.com/google/pytype/issues/1402.


stroxler (approved):
This looks good to me, thanks @rchen152 for the fix!

PR #921, opened 2023-05-10
skip escaped backslash in rf-string

### LINKED ISSUE #917: Native parser crashes on `rf"\\"`
```python
import contextlib, os
from libcst import parse_expression

@contextlib.contextmanager
def native_parser():
    var = os.environ.get("LIBCST_PARSER_TYPE")
    try:
        os.environ["LIBCST_PARSER_TYPE"] = "native"
        yield
    finally:
        os.environ.pop("LIBCST_PARSER_TYPE")
        if var is not None:
            os.environ["LIBCST_PARSER_TYPE"] = var

string_literal = r'rf"\\"'
parse_expression(string_literal)  # fine :-)
with native_parser():
    parse_expression(string_literal)  # fails :-(
```

jakkdl:
## Summary
Fixes #917 
## Test Plan
Added the failing test case from #917, as well as one for a normal raw string to make sure behavior is consistent.

### DIFF
--- a/native/libcst/src/tokenizer/core/mod.rs
+++ b/native/libcst/src/tokenizer/core/mod.rs
@@ -973,8 +973,8 @@ impl<'t> TokState<'t> {
                 }
                 (Some('\\'), _) if is_raw_string => {
                     self.text_pos.next();
-                    if let Some('"' | '\'') = self.text_pos.peek() {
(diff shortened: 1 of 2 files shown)


Zac-HD:
@zsol ping for review?

zsol (approved):
Thanks! LGTM

PR #926, opened 2023-05-22
relax validation rules on decorators

### LINKED ISSUE #486: Support PEP-614 - relaxed grammar restrictions on decorators

jakkdl:
relax validation on decorators

### DIFF
--- a/libcst/_nodes/statement.py
+++ b/libcst/_nodes/statement.py
@@ -1632,19 +1632,6 @@ class Decorator(CSTNode):
     #: Optional trailing comment and newline following the decorator before the next line.
     trailing_whitespace: TrailingWhitespace = TrailingWhitespace.field()
 
-    def _validate(self) -> None:
-        decorator = self.decorator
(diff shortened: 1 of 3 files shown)


zsol:
Thanks! I also relaxed the type of `Decorator.decorator` and added a test for this as well.

jakkdl:
Ah, good catch! :rocket:

PR #957, opened 2023-06-14
Bump pyo3 to 0.18

apbard:


zsol:
Thanks <3

zertosh:
@zsol, isn't this going to bite you later on when there really is a breaking change? Remember that there are no semver guarantees in `0.x`. I think you probably mean to have here at least `">=0.17, <=0.19"` or at most `">=0.17, <1"`.

PR #981, opened 2023-07-25
codemod/rename: Avoid duplicating import statements when the module name doesn't change

zsol:
## Summary

When renaming within the same module (i.e. `logging.warn -> logging.warning`), make sure the codemod doesn't insert an extra `import logging` statement.

## Test Plan

Added unit tests


itamaro (approved):
LG!

PR #994, opened 2023-08-24
Don't gather dirs ending .py

kit1980:
## Summary

`codemod.gather_files` should not try to gather dirs those names ending with .py, only regular files
(the issue encountered in an actual codebase).

## Test Plan

libcst/tests/test_e2e.py updated


kit1980:
The CI errors look unrelated:

Error: Cache folder path is retrieved for pip but doesn't exist on disk: /home/runner/.cache/pip

zsol:
Thanks!

PR #995, opened 2023-08-25
Remove criterion-cycles-per-byte dependency and related benchmark measurement

orf:
## Summary

criterion-cycles-per-byte is not supported on non-x86_64 platforms, so `cargo test` cannot be run:

```
❯ cargo test
   Compiling libcst v0.1.0 (x)
   Compiling criterion-cycles-per-byte v0.1.2
error: criterion-cycles-per-byte currently relies on x86 or x86_64.
  --> /Users/tom/.config/cargo/registry/src/index.crates.io-6f17d22bba15001f/criterion-cycles-per-byte-0.1.2/src/lib.rs:34:1
   |
34 | compile_error!("criterion-cycles-per-byte currently relies on x86 or x86_64.");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This PR makes it conditional and allows the tests to run.

PR #996, opened 2023-08-25
Switch to using thread_local regular expressions to stop mutext contention

### LINKED ISSUE #986: Performance issues when using multiple threads due to regex scratch space contention
There appears to be a lot of mutext contention when using `libcst` with multiple threads. I discovered this while parsing a lot of code across only 8 threads on my local machine. I was averaging about ~1.5k files per second with a flamegraph that spent 80% of the time inside `TextPattern.match_len`:

flamegraph-mutext

This is because the scratch-space for each regex defined in libcst is shared between all threads. As a proof of concept I switched the regex compilations to [use thread locals](https://github.com/orf/LibCST/commit/03f2d60321964c79402a4a9453ae336cfbd8e71c), and immediately saw a ~10x performance jump to ~12k files per second:

flamegraph

This is because the pattern scratch-space is no longer shared between threads, and thus there is no contention.

I'm happy to make a pull request with these changes, but I wanted to spur a discussion here because the trade-offs are not clear. If you create/destroy a lot of threads then this change would have a negative impact, however if your threads are long-lived and you are doing a lot of individual parsing calls then this is a great benefit.

orf:
Fixes #986. This is also rebased on top of https://github.com/Instagram/LibCST/pull/995 to allow me to run the benchmarks.

## Summary

When calling `libcst_native::parse_module` from multiple threads the parsing performance drops dramatically. This is somewhat documented [in the regex_automata crate](https://docs.rs/regex-automata/latest/regex_automata/) and in other places around the internet:

> You need to work-around contention issues with sharing a regex across multiple threads. The meta::Regex::search_with API permits bypassing any kind of synchronization at all by requiring the caller to provide the mutable scratch spaced needed during a search.

## Benchmarks

I've added a multi-threaded benchmark using Rayon. The use of thread locals improves the throughput by 22% and reduces the time by 18%:

```
parse_into_cst_parallel/all
                        time:   [476.86 ms 482.42 ms 488.68 ms]
                        thrpt:  [5.9343 Kelem/s 6.0113 Kelem/s 6.0814 Kelem/s]
                 change:
                        time:   [-19.567% -18.431% -17.452%] (p = 0.00 < 0.05)
                        thrpt:  [+21.142% +22.595% +24.327%]
                        Performance has improved.
```

However the throughput increase I've seen with high-volume "real-world" parsing is much, much higher.

# Benchmarks

These are the results of the benchmark added in this change, with and without the `thread_local!` change. It charts the time taken to parse the fixture files. The number of files is multiplied by the number of threads, which ranges from 1 to 10. At 10 threads the average time taken is 50% faster.

## Benchmark: Before

Screenshot 2023-08-25 at 21 09 00

## Benchmark: Before

Screenshot 2023-08-25 at 21 09 04

### DIFF
--- a/native/Cargo.lock
+++ b/native/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+checksum = "6748e8def348ed4d14996fa801f4122cd763fff530258cdc03f64b25f89d3a5a"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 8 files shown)


zsol (approved):
Thank you! This looks great to me. Happy to merge once #995 is in.

zsol:
Oh, one thing though: you'll have to update the MSRV to (looks like 1.60), but that should be fine.

orf:
Done @zsol! I also bumped some dependencies. notably the `regex` bump _also_ improves parsing speed quite a lot (the benchmarks above used the latest regex version). Let me know if you want me to split them out from this.

zsol:
Thank you!

PR #1007, opened 2023-08-30
Support files with mixed newlines

MichaReiser:
## Summary

This PR fixes an issue in LibCST native where it failed to parse a file that uses mixed `\r` and `\n` or `\r` and `\r\n` newlines. 

It also replaces the Regex to find all newlines with a simple call to `memchr`. This should be much faster because it doesn't require the overhead of a full regex automata. 

## Test Plan

Added a new unit test

> **Note**: Replacing all Regex usages in LibCST could result in a huge performance boost because thread locals in Rust are [slow](https://matklad.github.io/2020/10/03/fast-thread-locals-in-rust.html) (or not as fast as they could be) AND most regex are simple and can be replaced with some simple lexing on the string slice directly


native/libcst/src/tokenizer/whitespace_parser.rs
```diff
@@ -74,12 +74,44 @@ impl<'a> Config<'a> {
...
+        let default_newline = match lines.first().map(|line| line.as_bytes()).unwrap_or(&[]) {
+            [.., b'\r', b'\n'] => "\r\n",
+            [.., b'\n'] => "\n",
+            [.., b'\r'] => "\r",
+            _ => "\n",
+        };
 
         Self {
             input,
-            lines: input.split_inclusive(default_newline).collect(),
```
MichaReiser, line 82: This incorrectly assumed that a file only uses one kind of line break, resulting in some lines not being split.

native/libcst/src/tokenizer/whitespace_parser.rs
```diff
@@ -74,12 +74,44 @@ impl<'a> Config<'a> {
                 break;
             }
         }
-        let default_newline =
-            NEWLINE_RE_2.with(|r| r.find(input).map(|m| m.as_str()).unwrap_or("\n"));
+
+        let mut lines = Vec::new();
```
MichaReiser, line 78: A possible optimisation is to initialize the Vec with `input.len() / 88` (or 100 or 120, depending on how defensive we want to be). This turned out to improve performance for ruff.

native/libcst/src/tokenizer/whitespace_parser.rs
```diff
@@ -74,12 +74,44 @@ impl<'a> Config<'a> {
...
+        let mut start = 0;
+        let mut newline_positions = memchr2_iter(b'\n', b'\r', input.as_bytes());
+
+        while let Some(newline_position) = newline_positions.next() {
+            let newline_character = input.as_bytes()[newline_position] as char;
+
+            let len = if newline_character == '\r'
+                && input.as_bytes().get(newline_position + 1).copied() == Some(b'\n')
```
zsol, line 86: nit: can we get away without calling `.copied()`?

native/libcst/src/tokenizer/whitespace_parser.rs
zsol, line 78: Can you please add a simple unit test for the new logic in this file as well?

zsol (changes requested):
LGTM, but I'd like to have a unit test for `Config::new` before merging

native/libcst/src/tokenizer/whitespace_parser.rs
```diff
@@ -74,12 +74,44 @@ impl<'a> Config<'a> {
...
+            [.., b'\r', b'\n'] => "\r\n",
+            [.., b'\n'] => "\n",
+            [.., b'\r'] => "\r",
+            _ => "\n",
+        };
 
         Self {
             input,
-            lines: input.split_inclusive(default_newline).collect(),
+            lines,
```
MichaReiser, line 114: `has_trailing_newline` may need fixing too, to correctly handle `\r` endings but I'll leave this for another PR

zsol (approved):

zsol:
Thank you!

PR #1008, opened 2023-09-02
parser: remove Regexes from whitespace parser

zsol:
parser: remove Regexes from whitespace parser

removing Regexes from whitespace parser allows ditching of thread local storage + lazy initialization cost

This shows a modest 2% improvement in overall parse time (inflate is improved by 10%)

### DIFF
--- a/native/libcst/src/tokenizer/whitespace_parser.rs
+++ b/native/libcst/src/tokenizer/whitespace_parser.rs
@@ -7,19 +7,12 @@ use crate::nodes::{
     Comment, EmptyLine, Fakeness, Newline, ParenthesizableWhitespace, ParenthesizedWhitespace,
     SimpleWhitespace, TrailingWhitespace,
 };
-use memchr::memchr2_iter;
-use regex::Regex;
+use memchr::{memchr2, memchr2_iter};
 use thiserror::Error;
 
 use crate::Token;
 
 use super::TokType;
-
-thread_local! {
-    static SIMPLE_WHITESPACE_RE: Regex = Regex::new(r"\A([ \f\t]|\\(\r\n?|\n))*").expect("regex");
-    static NEWLINE_RE: Regex = Regex::new(r"\A(\r\n?|\n)").expect("regex");
-    static COMMENT_RE: Regex = Regex::new(r"\A#[^\r\n]*").expect("regex");
-}
 
 #[allow(clippy::upper_case_acronyms, clippy::enum_variant_names)]
 #[derive(Error, Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


zsol:
cc @orf @MichaReiser as recent relevant contributors

native/libcst/src/tokenizer/whitespace_parser.rs
```diff
@@ -231,29 +224,34 @@ pub fn parse_empty_lines<'a>(
...
     let newline_after = config.get_line_after_column(state.line, state.column_byte)?;
-    if let Some(newline_match) = NEWLINE_RE.with(|r| r.find(newline_after)) {
-        let newline_str = newline_match.as_str();
-        advance_this_line(
-            config,
-            state,
-            newline_str.chars().count(),
-            newline_str.len(),
-        )?;
+    let len = match newline_after.as_bytes() {
```
orf, line 246: I was interested in the performance characteristics of this vs other ways of checking a string prefix. With [this benchmark script](https://gist.github.com/orf/d4473dd0dff2bd480680b869f56929d6), 

```rust
match &thing.get(..2) {
        Some("\n") => 1,
        Some("\r\n") => 2,
        Some("\r") => 1,
        _ => 0,
    }
```

is a bit faster in all cases: `914` picoseconds vs `680` picoseconds for the `\r\n` case. Not that it makes much difference, but it is interesting because I would assume they optimize to the same code.

orf: Ahh, that's not exactly equivalent, it would return `None` on single digit characters. Ignore me.
```

**Comment by orf** [comment]
This looks good, but I was just thinking about something. It might be a dumb idea, but wouldn't it be more efficient to work out the byte indexes of whitespaces in one pass and re-use that?

```rust
let regex = regex::bytes::Regex::new(r#"(([ \t\x0c]|(\r\n?|\n))+)"#,).unwrap();

let whitespace_indexes: Vec<Range<usize>> = regex.find_iter(bytes).map(|r| r.range()).collect();
```

There are better datastructures than a vec, but if we have this and a byte offset we know if we're in a range of whitespace and what the end position for that whitespace range is? I dunno, something like:

```rust
let line_ws = config.ws_by_line[state.line];
for range in line_ws {
    if range.contains(state.byte_offset) {
       let ws = &config.input[state.byte_offset..range.end]
       return Ok(SimpleWhitespace(ws))
   }
}
```

MichaReiser:
It depends on how frequently this is called, but searching a string is probably more efficient than running a regex, allocating and writing all the positions, and then searching for it (even if it is a binary search). I'll review the code changes tomorrow or on Monday

native/libcst/src/tokenizer/whitespace_parser.rs
```diff
@@ -231,29 +224,34 @@ pub fn parse_empty_lines<'a>(
...
-    if let Some(comment_match) = COMMENT_RE.with(|r| r.find(newline_after)) {
-        let comment_str = comment_match.as_str();
-        advance_this_line(
-            config,
-            state,
-            comment_str.chars().count(),
-            comment_str.len(),
-        )?;
-        return Ok(Some(Comment(comment_str)));
+    if newline_after.is_empty() || newline_after.as_bytes()[0] != b'#' {
```
MichaReiser, line 227: Nit
```suggestion
    if newline_after.get(0)  != Some(&b'#') {
```

native/libcst/src/tokenizer/whitespace_parser.rs
```diff
@@ -376,13 +374,18 @@ pub fn parse_simple_whitespace<'a>(
...
-        let x = x?;
-        Ok(SIMPLE_WHITESPACE_RE.with(|r| {
-            r.find(x)
-                .expect("SIMPLE_WHITESPACE_RE supports 0-length matches, so it must always match")
-                .as_str()
-        }))
+        let line = config.get_line_after_column(line, col)?;
+        let bytes = line.as_bytes();
+        let mut idx = 0;
+        while idx < bytes.len() {
```
MichaReiser, line 380: An abstraction that worked really well for Ruff (and Rust) is to use the following `Cursor` abstraction to lex strings. It avoids the need for these manual `match` statements. 

https://github.com/astral-sh/ruff/blob/4d49d5e8451277f8159a30b7da187626d3a75494/crates/ruff_python_parser/src/lexer/cursor.rs#L14-L13

MichaReiser (approved):

PR #1012, opened 2023-09-06
build: add github action for publishing crates

### LINKED ISSUE #967: Publishing the parser to crates.io
:wave: I know there's a line in the [README](https://github.com/Instagram/LibCST/blob/a3f5bf97d631e79c3395a249e15645cfbc225a4d/native/libcst/README.md?plain=1#L10) about publishing to `crates.io`, but I just wanted to upvote it and couldn't find a better avenue to do so. We use the LibCST parser to support some autofixes in [Ruff](https://github.com/astral-sh/ruff), and we currently package LibCST as a Git dependency. It'd be more convenient and reliable to pull it in from crates.io (we occasionally see spurious network failures with Git dependencies), and would also enable us to publish Ruff itself to crates.io in the future.

Alternatively, if there isn't interest from the maintainers in publishing this to crates.io, is there any way you'd be open to having me publish (e.g.) an undocumented `ruff_libcst` package of a fork?

Thanks!

manmartgarc:
## Summary

First attempt at automating the publishing of `libcst` and `libcst_derive` to the `crates.io` registry on `release` and `push` to the main branch. Closes #967.

### Notes
- The `--allow-dirty` flag was needed during testing since for some reason the files in the workflow are not committed to git and `cargo publish` complains about this.
- Not sure if we should keep the `libcst_derive` dependency of `libcst` as a local path dependency or if it should be changed to point to the `crates.io` registry.

## Test Plan
The repository owner should set up a new secret with the `crates.io` token, and name it `CRATES_IO_TOKEN` so that the workflow can use the token when publishing to the registry.


.github/workflows/crates_upload.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          toolchain: stable
+
+      - name: Publish libcst_derive to crates.io
+        uses: actions-rs/cargo@v1
+        with:
+          command: publish
+          args: >
+            --manifest-path ./native/libcst_derive/Cargo.toml
+            --allow-dirty
```
zsol, line 31: why `--allow-dirty`?

zsol (commented):
This will publish to the version hardcoded in `Cargo.toml`, right? One of the nice things in the pypi workflow is that the version is derived from git, so we:
1. Don't have to make sure the version number in the two `Cargo.toml` files match the git tag and `pyproject.toml`
2. Can publish intermediate versions to a test registry; this is great for testing the release pipeline itself

Is there anything we can do to have these properties on the rust side?

.github/workflows/crates_upload.yml
manmartgarc: I was testing locally with [nektos/act](https://github.com/nektos/act) and was getting an issue where `cargo publish` wouldn't allow files that weren't committed, even though the files were actually committed. Not sure if this would be the case on GitHub

manmartgarc:
> This will publish to the version hardcoded in `Cargo.toml`, right? One of the nice things in the pypi workflow is that the version is derived from git, so we:
> 
> 1. Don't have to make sure the version number in the two `Cargo.toml` files match the git tag and `pyproject.toml`
> 2. Can publish intermediate versions to a test registry; this is great for testing the release pipeline itself
> 
> Is there anything we can do to have these properties on the rust side?

I am assuming we should be able to derive the version from git, will check more on this. Regarding the second question there is a `--dry-run` flag that can be passed to `cargo publish` which [will do everything](https://doc.rust-lang.org/cargo/reference/publishing.html#packaging-a-crate) up to publishing the crates on a registry.

manmartgarc:
> This will publish to the version hardcoded in `Cargo.toml`, right? One of the nice things in the pypi workflow is that the version is derived from git, so we:
> 
> 1. Don't have to make sure the version number in the two `Cargo.toml` files match the git tag and `pyproject.toml`
> 2. Can publish intermediate versions to a test registry; this is great for testing the release pipeline itself
> 
> Is there anything we can do to have these properties on the rust side?

It looks like someone is doing this by [replacing the version](https://stackoverflow.com/questions/52176130/how-to-tell-cargo-to-use-git-tags-to-determine-the-crate-version) in the `Cargo.toml` file. Is this something you'd be open to?

manmartgarc:
https://github.com/Instagram/LibCST/pull/1012/commits/f23af55a572161c438422a01f879708c1d57ff1f adds:

- the functionality to do a `--dry-run` release on `push` and do the full release on `release`.
- Hackily replace the versions in the `Cargo.toml` files by the one in the current git tag using `sed`.

Not sure if this is good enough, but I wasn't able to find a homegrown way of using the tag in `cargo publish`. Another thing to note is that `libcst` is published using the local version of `libcst_derive`. I am not sure if we should also publish `libcst_derive` but publishing the latter might be avoided if only `libcst` should be on the registry.

manmartgarc:
Huh. Why is there a typing error if no source changes were made?

zsol:
I don't know, but it appears on all PRs recently. Some dependency must've
changed

zsol:
FTR, I fixed this in #1032

manmartgarc:
> FTR, I fixed this in #1032

I'll be able to merge this into the PR later today 👍

manmartgarc:
I had made a mess of the branch with a `git pull --rebase` but it looks like now the git history looks better. Not sure how this change is affecting the actions yet. Could you approve the workflow to see what happens?

manmartgarc:
Bumping this. Are there any blockers currently?

manmartgarc:
@zsol checking in on this. I see that the right versions are already on `crates.io`: https://crates.io/search?q=libcst. Is there anything else to do here?

manmartgarc:
@zsol should I close this PR? I see `v1.0.1` on crates.io, thinking this might be all set up?

PR #1014, opened 2023-09-09
Scope provider changes for type annotations

zsol:
Scope provider changes for type annotations

### DIFF
--- a/libcst/metadata/tests/test_scope_provider.py
+++ b/libcst/metadata/tests/test_scope_provider.py
@@ -11,9 +11,11 @@ from unittest import mock
 
 import libcst as cst
 from libcst import ensure_type
+from libcst._parser.entrypoints import is_native
 from libcst.metadata import MetadataWrapper
 from libcst.metadata.scope_provider import (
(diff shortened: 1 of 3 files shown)


libcst/metadata/tests/test_scope_provider.py
```diff
@@ -1982,3 +1984,210 @@ def something():
...
+        outer = ensure_type(m.body[0], cst.ClassDef)
+        inner = ensure_type(outer.body.body[0], cst.ClassDef)
+        # note: this is different from global scope
+        outer_scope = scopes[outer.body.body[0]]
+        inner_assignment = list(outer_scope["Inner"])[0]
+        self.skipTest(
+            "f's body is subject to the annotation scoping rules"
+            " created by its type parameters, but it shouldn't be"
+        )
+        self.assertEqual(len(inner_assignment.references), 1)
```
zsol, line 2193: Gotta fix this somehow. The things defined in the `AnnotationScope` should still be accessible from the body, but not the immediately enclosing class assignments

Issue #1023, opened 2023-09-22
AddImportsVisitor may add imports after usage


andrecsilva:
Consider the following, suppose the `c()` call was just added by another transformation/codemod:
```python
c()
import a
from b import d
```

Trying to add a `c` import will result in it being added after the call:
```python
In [1]: import libcst as cst

In [2]: from libcst.codemod.visitors import AddImportsVisitor

In [3]: from libcst.codemod.visitors._imports import ImportItem

In [4]: print(cst.parse_module("c()\nimport a\nfrom b import d").visit(AddImportsVisitor(cst.codemod.CodemodContext(), [ImportItem('c',None,None)])).code)
c()
import a
from b import d
import c

In [5]: print(cst.parse_module("c()\nimport a\nfrom b import d").visit(AddImportsVisitor(cst.codemod.CodemodContext(), [ImportItem('b','c',None)])).code)
c()
import a
from b import c, d
```
A possible fix for that is to change it so that imports are only added at the top of the file.

PR #1024, opened 2023-09-22
AddImportsVisitor: add imports before the first non-import statement

andrecsilva:
## Summary
Proposed fix for #1023. It changes `AddImportsVisitor` so it only considers adding new imports after/at the first import block before any other statements, if any.

## Test Plan
Added a few tests to `test_add_imports.py`

### DIFF
--- a/libcst/codemod/visitors/_add_imports.py
+++ b/libcst/codemod/visitors/_add_imports.py
@@ -10,31 +10,19 @@ import libcst
 from libcst import matchers as m, parse_statement
 from libcst._nodes.statement import Import, ImportFrom, SimpleStatementLine
 from libcst.codemod._context import CodemodContext
-from libcst.codemod._visitor import ContextAwareTransformer, ContextAwareVisitor
+from libcst.codemod._visitor import ContextAwareTransformer
+from libcst.codemod.visitors._gather_imports import GatherImportsVisitor
 from libcst.codemod.visitors._imports import ImportItem
 from libcst.helpers import get_absolute_module_from_package_for_import
 from libcst.helpers.common import ensure_type
 
 
-class _GatherTopImportsBeforeStatements(ContextAwareVisitor):
+class _GatherTopImportsBeforeStatements(GatherImportsVisitor):
     """
     Works similarly to GatherImportsVisitor, but only considers imports
     declared before any other statements of the module with the exception
     of docstrings and __strict__ flag.
     """
-
-    def __init__(self, context: CodemodContext) -> None:
-        super().__init__(context)
-        # Track the available imports in this transform
-        self.module_imports: Set[str] = set()
-        self.object_mapping: Dict[str, Set[str]] = {}
-        # Track the aliased imports in this transform
-        self.module_aliases: Dict[str, str] = {}
-        self.alias_mapping: Dict[str, List[Tuple[str, str]]] = {}
-        # Track all of the imports found in this transform
-        self.all_imports: Set[Union[libcst.Import, libcst.ImportFrom]] = set()
-        # Track the import for every symbol introduced into the module
-        self.symbol_mapping: Dict[str, ImportItem] = {}
 
     def leave_Module(self, original_node: libcst.Module) -> None:
         start = 1 if _skip_first(original_node) else 0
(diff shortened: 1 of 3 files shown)


libcst/codemod/visitors/_add_imports.py
```diff
@@ -8,11 +8,47 @@
...
+
+
+class _GatherTopImportsBeforeStatements(GatherImportsVisitor):
+    """
+    Works similarly to GatherImportsVisitor, but only considers imports
+    declared before any other statements of the module with the exception
+    of docstrings and __strict__ flag.
+    """
+
+    def leave_Module(self, original_node: libcst.Module) -> None:
```
zsol, line 27: This isn't a great use of inheritance, `_GatherTopImportsBeforeStatements` doesn't really specialize the behavior of `GatherImportsVisitor`, it completely changes it.

A better way to model this would be to extract the two `_handle_...` methods into a mixin class and let both visitors inherit from that.

Then `_GatherTopImportsBeforeStatements` won't need to override `visit_ImportFrom` and `visit_Import` with dummy implementations (and potentially others in the future). Instead you could just implement `visit_Module` and return `False` at the end. Bonus points for making it significantly faster for large documents.

libcst/codemod/visitors/_add_imports.py
```diff
@@ -169,14 +205,14 @@ def __init__(
             for module in sorted(from_imports_aliases)
         }
 
-        # Track the list of imports found in the file
-        self.all_imports: List[Union[libcst.Import, libcst.ImportFrom]] = []
+        # Track the list of imports found at the top of the file
+        self.all_imports: Set[Union[libcst.Import, libcst.ImportFrom]] = set()
```
zsol, line 209: Unfortunately the List->Set is a breaking change in the API, let's avoid it unless it's absolutely necessary.

libcst/codemod/visitors/_add_imports.py
```diff
@@ -213,6 +249,10 @@ def leave_ImportFrom(
             # There's nothing to do here!
             return updated_node
 
+        # Ensure this is on of the imports at the top
```
zsol, line 252: nit
```suggestion
        # Ensure this is one of the imports at the top
```

libcst/codemod/visitors/_add_imports.py
```diff
@@ -8,11 +8,47 @@
...
+
+    def leave_Module(self, original_node: libcst.Module) -> None:
+        start = 1 if _skip_first(original_node) else 0
+        for stmt in original_node.body[start:]:
+            if m.matches(
+                stmt,
+                m.SimpleStatementLine(body=[m.ImportFrom() | m.Import()]),
+            ):
+                stmt = ensure_type(stmt, SimpleStatementLine)
+                imp = ensure_type(stmt.body[0], Union[ImportFrom, Import])
```
zsol, line 35: I believe this is causing test failures

zsol (changes requested):
I think the current behavior is useful in cases like:
```
from a import side_effect
a()

from b import foo
...
bar()
```
When we request the import for `b.bar`, the current behavior extends the second `ImportFrom` statement, but with this PR, we would end up with:
```
from a import side_effect
from b import bar
a()

from b import foo
...
bar()
```

Which is at least a bit unfortunate. It's also unfortunate that we don't capture this in any test case.

I think my main concern stems from the fact that we'd be always inserting imports before the first non-import statement even if it isn't necessary. If you could change the PR to only apply the new behavior when the old would obviously break code, I'd be happy to accept it

andrecsilva:
> I think the current behavior is useful in cases like:
> 
> ```
> from a import side_effect
> a()
> 
> from b import foo
> ...
> bar()
> ```
> 
> When we request the import for `b.bar`, the current behavior extends the second `ImportFrom` statement, but with this PR, we would end up with:
> 
> ```
> from a import side_effect
> from b import bar
> a()
> 
> from b import foo
> ...
> bar()
> ```
> 
> Which is at least a bit unfortunate. It's also unfortunate that we don't capture this in any test case.
> 
> I think my main concern stems from the fact that we'd be always inserting imports before the first non-import statement even if it isn't necessary. If you could change the PR to only apply the new behavior when the old would obviously break code, I'd be happy to accept it

Thanks for pointing the issues. I've hopefully fixed most of them with the last commit. 

I do have some questions about what is the expected behavior of `AddImportsVisitor`. I understand we want to avoid "duplicated" imports as you've pointed out, but the issue seems a bit more complex.

First, note that the current behavior is a bit more problematic than it seems. Consider the following and suppose we want to add `from x import bar`:
```python
foo()
def f():
    from x import a
    pass
from x import b
bar()
from x import c
```

The current behavior would add bar to `from x import a`. It does so because it finds the first `from x import ...` from the `visit_ImportFrom` perspective and adds the new import there.

Now consider:
```python
foo()
def f():
    from x import a
    bar()
from x import b
foo()
from x import c
bar()
```
What do you feel is the best place(s) to add `bar` import(s) here?

Do we want it to always add it to the "nearest" import block? or do we want a single import that covers all references to `bar()`?

Nevertheless, I feel that a general solution that cover these properties would be far more complex than what we have and the proposed solution in this PR is a good compromise. 

Of course, if you have a good/simple solution for this, I'd be happy to accept suggestions.

zsol:
I agree, `AddImportsVisitor` won't be able to do a perfect job all the time unless we include expensive scope analysis into it (and even then, there's only so much we can do). It's a balance. I do think the current behavior strikes a pretty good balance, and we should only add support for edge cases if they don't compromise on that balance.

And to directly answer your question: no, I have no solution for the cases you point out that wouldn't make `AddImportsVisitor` significantly more expensive.

The changes in this PR look good at a quick glance, but give me a bit more time to properly review them again, before merging.

zsol (approved):

zsol:
Thanks for putting this together, @andrecsilva!

PR #1035, opened 2023-10-05
Fix link in type_inference_provider.py

kit1980:
Same change as https://github.com/Instagram/LibCST/pull/913, but in the docstring.


zsol:
Thanks!

PR #1038, opened 2023-10-06
Update pyproject.toml for Python 3.12 support

itamaro:
Add 3.12 classifier and update description to correctly reflect supported Python versions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["setuptools", "wheel", "setuptools-rust"]
 
 [project]
 name = "libcst"
-description = "A concrete syntax tree with AST-like properties for Python 3.8, 3.9, 3.10, 3.11, and 3.12 programs."
+description = "A concrete syntax tree with AST-like properties for Python 3.0 through 3.12 programs."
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -3,7 +3,7 @@ requires = ["setuptools", "wheel", "setuptools-rust"]
 
 [project]
 name = "libcst"
-description = "A concrete syntax tree with AST-like properties for Python 3.5, 3.6, 3.7, 3.8, 3.9, and 3.10 programs."
+description = "A concrete syntax tree with AST-like properties for Python 3.8, 3.9, 3.10, 3.11, and 3.12 programs."
```
zsol, line 6: Oh, we still support parsing 3.5+ documents, its just the runtime that must be 3.8+

zsol (changes requested):

pyproject.toml
itamaro: ah it's not clear if this refers to runtime version or document version. I'll change it to just consistent with the README :)

zsol (approved):
🎉

PR #1042, opened 2023-10-15
Replace distutils.spawn.find_executable with shutil.which to support setuptools-less environments

hoefling:
## Summary

One of the changes that came with Python 3.12 is that `setuptools` is not installed in a new virtual environment anymore (change introduced in https://github.com/python/cpython/issues/95299). This means that `distutils` which was also removed in 3.12, can not be imported. To reproduce:

```sh
$ python3.12 -m venv tst
$ source tst/bin/activate
$ pip install libcst
$ python -m libcst.tool -h
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/tmp/tst/lib64/python3.12/site-packages/libcst/tool.py", line 13, in <module>
    import distutils.spawn
ModuleNotFoundError: No module named 'distutils'
```
The proposed changes replace the single usage of `distutils`, namely `distutils.spawn.find_executable`, with `shutil.which` that does the same thing.

An alternative could be making `setuptools` a runtime dependency, but IMO it is an overkill just for the single function access.

## Test Plan

Not sure about this - should I add tests for the cli tool itself? Looks like the affected code isn't tested yet, and `setuptools` is pulled by `setuptools-rust` in the development environment automatically so `distutils` is always available in the test run. Although I think it can be circumvented by monkeypatching certain imports in selected tests.

### DIFF
--- a/libcst/tool.py
+++ b/libcst/tool.py
@@ -10,11 +10,11 @@
 
 import argparse
 import dataclasses
-import distutils.spawn
 import importlib
 import inspect
 import os
 import os.path
+import shutil
 import sys
 import textwrap
 from abc import ABC, abstractmethod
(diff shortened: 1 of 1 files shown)


hoefling:
Ok, already fixed in https://github.com/Instagram/LibCST/pull/1099, so doing the cla fuzz was pointless after all :woman_shrugging:

PR #1057, opened 2023-11-29
Fix typo in readme

joethechicken:
Noticed this minor typo while reviewing the README


zsol:
Both forms of "Acknowledgements" are fine, I think.

PR #1058, opened 2023-11-30
Fix bug in MatchMapping class

zrr1999:
This pull request fixes a bug in the MatchMapping class where a trailing comma was allowed without the presence of the **rest parameter. The bug was causing an error in the validation process. This PR ensures that a trailing comma is only allowed when the **rest parameter is present.

Reproduction:
```py
import libcst as cst


cst.parse_module(r"""
match x:
    case {
        "a": a,
        **rest,
    }:
        print(a)
""")

```


zrr1999:
CLA signed

PR #1071, opened 2023-12-19
Allow `Element::codegen` to be used by external users

Wilfred:
The `Codegen` trait is `pub`, but users wanting to explicitly perform codegen for `Element` had to copy-paste this part of the code.

(See also D52275507.)


zsol:
Thanks!

PR #1082, opened 2024-01-01
Update test_fix_pyre_directives.py

anonymousdouble:
## Summary
refactor with chain constant value assignment to make code more Pythonic.


aleivag:
@anonymousdouble 

This changes are not neutral, and have broken tests

aleivag:
@anonymousdouble  thanks for pushing this, the test pass now yeyy... but the file is not formatted correctly... can you run something like `hatch run ufmt format libcst` ?

aleivag:
Thank you for pushing thorough!

Issue #1054, opened 2023-11-19
Binder/IPython (Notebook) interactive online tutorial does not start


AlbertMietus:
On  ReadTheDoc page https://libcst.readthedocs.io/en/latest/tutorial.html (Parsing and Visiting) a (link to) a interactive online tutorial is provides (which is nic) However: **It does not start**

| _Failed to build libcst_
|  _ERROR: Could not build wheels for libcst, which is required to install pyproject.toml-based projects_

Partial related: #450

zsol:
I believe this is fixed now with #1083. Thanks for reporting

PR #1083, opened 2024-01-02
installing rustc/cargo for mybinder demo

aleivag:
## Summary

when we switch to the rust compiler by default, mybinder stop working, as reported in https://github.com/Instagram/LibCST/issues/1054 this is because the binder docker image does not have a rust compiler or tools, this install them by using the apt.txt file

## Test Plan

if you go to https://mybinder.org/v2/gh/aleivag/LibCST/fix-binder?filepath=docs/source/tutorial.ipynb you'll see this works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ libcst/_version.py
 target/
 venv/
 .venv/
+.idea/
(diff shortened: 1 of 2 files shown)


aleivag:
notice that this just fix the binder docker image creation... I think this is still missing installing python dependencies like typing_inspect , i plan to fix that part in a subsequent PR

zsol:
<3

PR #1129, opened 2024-04-04
fix: __all__ references are not ignored

vantaboard:
## Summary
Creating a `RemoveUnusedImports` transformer according to the docs does not properly handle the `__all__` variable.

## Test Plan
I added a test to `libcst/metadata/tests/test_scope_provider.py` that ensures the imports have references that target the strings in `__all__`.

### DIFF
--- a/libcst/metadata/scope_provider.py
+++ b/libcst/metadata/scope_provider.py
@@ -958,7 +958,7 @@ class ScopeVisitor(cst.CSTVisitor):
                 name = name.decode("utf-8")
 
             access = Access(
-                cst.Name(name),
+                node,
                 self.scope,
                 is_annotation=False,
                 is_type_hint=False,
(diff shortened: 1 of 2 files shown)


libcst/metadata/scope_provider.py
```diff
@@ -950,6 +951,30 @@ def _handle_string_annotation(
...
+            name = name.replace('"', "")
+
+            access = Access(
+                cst.Name(name),
+                self.scope,
+                is_annotation=False,
+                is_type_hint=False,
+            )
+
+            list(self.scope.assignments[name])[0].record_access(access)
```
zsol, line 974: I would add an entry into `self.__deferred_accesses` instead, to make sure this `Access` gets the same treatment as any other.

zsol (changes requested):
I wonder what happens for this snippet:
```
import a, b
__all__: "list[str]" = ["a", "b"]
```

zsol:
Gotta fix the type errors (looks like `Name.__init__` needs updating too) and this is good to go!

Issue #924, opened 2023-05-17
RemoveImportsVisitor crashes when parsing Literal["r+"]


Hnasar:
https://peps.python.org/pep-0586/#shortening-unions-of-literals

The parser is trying to see if the string "r+" refers to some stringified-type and fails to parse it as an expression, but it should just take "r+" as a literal string.

This crashes in both the python and rust/native parser as of libcst 0.4.9

Repro:
```python
# foo.py
from typing import Literal
x: Literal["r+"] = "r+"
```

```python
# mods/mymod.py
import libcst as cst
from libcst.codemod import VisitorBasedCodemodCommand
from libcst.codemod.visitors import RemoveImportsVisitor


class CrashMod(VisitorBasedCodemodCommand):
    DESCRIPTION: str = "Crash on literal"

    def leave_AnnAssign(
        self, original_node: cst.AnnAssign, updated_node: cst.AnnAssign,
    ) -> cst.AnnAssign:
        RemoveImportsVisitor.remove_unused_import(self.context, "typing")
        return updated_node
```

```yaml
# .libcst.codemod.yaml
generated_code_marker: '@generated'
formatter: ['black', '-']
blacklist_patterns: []
modules:
- 'mods'
repo_root: '.'

```

Output:
```
python -m libcst.tool codemod mymod.CrashMod foo.py
...
  File "libcst/codemod/visitors/_gather_string_annotation_names.py", line 65, in visit_SimpleString
    self.handle_any_string(node)
  File "libcst/codemod/visitors/_gather_string_annotation_names.py", line 74, in handle_any_string
    mod = cst.parse_module(value)
  File "libcst/_parser/entrypoints.py", line 109, in parse_module
    result = _parse(
  File "libcst/_parser/entrypoints.py", line 56, in _parse
    return _pure_python_parse(
  File "libcst/_parser/entrypoints.py", line 89, in _pure_python_parse
    result = parser.parse()
  File "libcst/_parser/base_parser.py", line 110, in parse
    self._add_token(token)
  File "libcst/_parser/base_parser.py", line 186, in _add_token
    raise ParserSyntaxError(
libcst._exceptions.ParserSyntaxError: Syntax Error @ 1:3.
Incomplete input. Unexpectedly encountered '\n'.

r+
  ^
```

Hnasar:
This also fails when parsing `Literal["1d", "1w"]`

```
  File foo line 635, in transform_module
[...]

PR #1130, opened 2024-04-06
Fix Literal parse error in RemoveImportsVisitor

camillol:
`GatherNamesFromStringAnnotationsVisitor`, which is called by `GatherUnusedImportsVisitor`, which is called by `RemoveImportsVisitor`, attempts to parse any string contained in an annotation as if it were Python code. This behavior was added in https://github.com/Instagram/LibCST/pull/353 with the goal of parsing type names wrapped in strings.

However, not all strings occurring inside annotations can or should be interpreted that way. For example:
```python
from typing import Literal

a: Literal["in"]
b: list[Literal["1x"]]
c: Literal["Any"]
```

`a` and `b` will cause a `ParserSyntaxError`, while `c` will be incorrectly treated as a mention of the symbol `Any`.

This PR makes two changes to `GatherNamesFromStringAnnotationsVisitor`:
- Anything inside a `Literal` is ignored.
  - This fixes all of the above examples.
- Failures to parse a string inside an annotation as Python code are simply ignored.
  - This will take care of any other usages of non-code strings inside an annotation.

Fixes:
- https://github.com/Instagram/LibCST/issues/924
- https://github.com/pydantic/bump-pydantic/issues/124
- https://github.com/pydantic/bump-pydantic/issues/115

### DIFF
--- a/libcst/codemod/visitors/_gather_string_annotation_names.py
+++ b/libcst/codemod/visitors/_gather_string_annotation_names.py
@@ -44,6 +44,11 @@ class GatherNamesFromStringAnnotationsVisitor(ContextAwareVisitor):
     def leave_Annotation(self, original_node: cst.Annotation) -> None:
         self._annotation_stack.pop()
 
+    def visit_Subscript(self, node: cst.Subscript) -> bool:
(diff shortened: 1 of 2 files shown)


zsol:
Thank you!

PR #1135, opened 2024-04-17
✏️ Fix typo in `docs/source/scope_tutorial.ipynb`

tiangolo:
## Summary

✏️ Fix typo in `docs/source/scope_tutorial.ipynb`

## Test Plan

Only docs


zsol:
Thank you for both of these!

PR #1137, opened 2024-04-29
Make nodes fields filtering process public

zaicruvoir1rominet:
## Summary
Make the nodes fields filtering process - from libcst.tool - public, so that other libraries may provide their own custom representation of LibCST graphs.

## Test Plan
Uses the existing test_tool tests, which already cover this code section.

### DIFF
--- a/libcst/tool.py
+++ b/libcst/tool.py
@@ -44,6 +44,98 @@ from libcst.codemod import (
 _DEFAULT_INDENT: str = "  "
 
 
+def filter_node_fields(
+    node: CSTNode,
+    *,
+    show_defaults: bool,
+    show_syntax: bool,
+    show_whitespace: bool,
+) -> Sequence["dataclasses.Field[CSTNode]"]:
+    """
+    Returns a filtered sequence of a node's fields.
+
+    Setting ``show_whitespace`` to ``False`` will filter whitespace fields.
+
+    Setting ``show_defaults`` to ``False`` will filter fields if their value is equal to
+    the default value ;  while respecting  the value of ``show_whitespace``.
+
+    Setting ``show_syntax``  to ``False`` will filter syntax fields ; while respecting
+    the value of ``show_whitespace`` & ``show_defaults``.
+    """
+
+    fields: Sequence["dataclasses.Field[CSTNode]"] = dataclasses.fields(node)
+
+    # Hide all fields prefixed with "_"
+    fields = [f for f in fields if f.name[0] != "_"]
+
+    # Filter whitespace nodes if needed
+    if not show_whitespace:
+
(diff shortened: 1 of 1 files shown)


libcst/tool.py
```diff
@@ -44,6 +44,98 @@
...
+            if "lines_after_decorators" in field.name:
+                return True
+            if isinstance(node, (IndentedBlock, Module)) and field.name in [
+                "header",
+                "footer",
+            ]:
+                return True
+            if isinstance(node, IndentedBlock) and field.name == "indent":
+                return True
+            return False
```
zsol, line 88: This could be its own public helper function

libcst/tool.py
```diff
@@ -44,6 +44,98 @@
...
+                # This is a value that can optionally be specified, so its
+                # definitely syntax.
+                return True
+
+            for name in ["Semicolon", "Colon", "Comma", "Dot", "AssignEqual"]:
+                # These are all nodes that exist for separation syntax
+                if name in type_str:
+                    return True
+
+            return False
```
zsol, line 132: This could be its own public helper function

zsol (changes requested):
High level making these public makes sense, but:
1. it's worth extracting `_is_syntax` and `_is_whitespace` into their own public functions (with docstrings and tests)
2. move all of these under `libcst.helpers`, please

libcst/tool.py
zaicruvoir1rominet: Following functions - with docstring and tests - have been created in ``libcst.helpers``:
- ``get_node_fields``
- ``is_whitespace_node_field``
- ``is_syntax_node_field``
- ``get_field_default_value``
- ``is_default_node_field``
- ``filter_node_fields``  

All above functions are now part the Sphinx Helpers doc.

``libcst.tool.dump`` now uses ``filter_node_fields``.

zaicruvoir1rominet: *(Please check the tests ; I was not really sure on how to properly cover use cases, so do tell if works need to be done there)*

zaicruvoir1rominet:
Following functions - with docstring and tests - have been created in ``libcst.helpers``:
- ``get_node_fields``
- ``is_whitespace_node_field``
- ``is_syntax_node_field``
- ``get_field_default_value``
- ``is_default_node_field``
- ``filter_node_fields``  

All above functions are now part the Sphinx Helpers doc.

``libcst.tool.dump`` now uses ``filter_node_fields``.

zaicruvoir1rominet:
*(Please check the tests ; I was not really sure on how to properly cover use cases, so do tell if works need to be done there)*

zsol:
Gotta fix the type errors and this is good to go.

zsol (approved):
Thank you!

Issue #1127, opened 2024-04-03
Implement PEP 696 (defaults for type params)


zsol:
https://peps.python.org/pep-0696/#grammar-changes will ship with Python 3.13, so I'd like to get this into LibCST by the time the release candidates for 3.13 are out. Assign to yourself or comment if you start working on it, please.

## Node changes
Seems like the most straightforward way is to extend `statement.TypeParam` to own the equals sign (which owns its surrounding whitespace), and the default value itself.

* The [python version](https://github.com/Instagram/LibCST/blob/9f54920d9d32a519aa1702485ed56c55d8189a0c/libcst/_nodes/statement.py#L3644) needs these extra fields. You can take inspiration from [`expression.Param`](https://github.com/Instagram/LibCST/blob/9f54920d9d32a519aa1702485ed56c55d8189a0c/libcst/_nodes/expression.py#L1778-L1782).
* Also add validation to make sure the equals sign and the default value can only appear together (`expression.Param` also has an example for this)
* We should also add validation in `statement.TypeParameters` to ensure that a `TypeParam` without a default value never follows one with.
* You can exercise the Python code with tests [here](https://github.com/Instagram/LibCST/blob/main/libcst/_nodes/tests/test_type_alias.py).
* The [rust version](https://github.com/Instagram/LibCST/blob/9f54920d9d32a519aa1702485ed56c55d8189a0c/native/libcst/src/nodes/statement.rs#L3476-L3479) needs extending in a similar way.
  * [This part](https://github.com/Instagram/LibCST/blob/main/native/libcst/README.md#nodes) explains various naming conventions and helpers available in Rust for defining nodes.
  * There's no validation for the Rust nodes, and also no `MaybeSentinel` - we use `Option<_>` to signal optional values.
  * I _think_ you won't need to store any actual tokens, so the Rust changes should look equivalent to Python.

## Grammar changes
I recommend reading [this readme](https://github.com/Instagram/LibCST/blob/main/native/libcst/README.md) to get acquainted with the rust bits at a high level.
[...]

PR #1141, opened 2024-05-08
Implement PEP-696

thereversiblewheel:
## Summary
This  PR adds the ability for LibCST to  parse the "=" operator inside of the square brackets as outlined in PEP-696

## Major Changes
Additional python tests were added
Additional roundtrip tests were added (taken from the cython implementation)

##Related Links

PEP-696: https://peps.python.org/pep-0696/#grammar-changes
Issue: https://github.com/Instagram/LibCST/issues/1127
## Test Plan

Round trip tests we're added


native/libcst/src/nodes/statement.rs
```diff
@@ -3490,7 +3497,15 @@ impl<'r, 'a> Inflate<'a> for DeflatedTypeParam<'r, 'a> {
     fn inflate(self, config: &Config<'a>) -> Result<Self::Inflated> {
         let param = self.param.inflate(config)?;
         let comma = self.comma.inflate(config)?;
-        Ok(Self::Inflated { param, comma })
+        let equal = self.equal.inflate(config)?;
+        let default = self.default.inflate(config)?;
```
zsol, line 3501: This is not very obvious, but the order of `inflate` calls matter very much because they mutate whitespace that's shared between them. They generally should follow the same order as they are invoked for `codegen`.

zsol (changes requested):
lgtm at a high level, I have some minor comments, the most important thing is to fix the ordering of inflate statements.
Also pls add more test cases for roundtrip

zsol (approved):
Looks great, thanks :)

PR #1148, opened 2024-05-14
FullyQualifiedNameProvider: Optionally consider pyproject.toml files when determining a file's module name and package

camillol:
When determining the module name and package for a source file, LibCST currently takes the file name relative to the repository root. With this change, it will instead look for the closest directory containing a `pyproject.toml` file in the file's ancestors, and consider the file name relative to that directory.

This gives correct module and package names when processing a repository containing multiple packages. For files that are not covered by a `pyproject.toml`, the search ends with the given repository root, thus retaining the old behavior.

### DIFF
--- a/libcst/helpers/module.py
+++ b/libcst/helpers/module.py
@@ -5,7 +5,7 @@
 #
 from dataclasses import dataclass
 from itertools import islice
-from pathlib import PurePath
+from pathlib import Path
 from typing import List, Optional
 
 from libcst import Comment, EmptyLine, ImportFrom, Module
@@ -136,7 +136,18 @@ def calculate_module_and_package(
 ) -> ModuleNameAndPackage:
     # Given an absolute repo_root and an absolute filename, calculate the
     # python module name for the file.
-    relative_filename = PurePath(filename).relative_to(repo_root)
+    # But also look for pyproject.toml files, indicating nested packages in the repo.
+    abs_repo_root = Path(repo_root).resolve()
+    abs_filename = Path(filename).resolve()
+    package_root = abs_filename.parent
+    while package_root != abs_repo_root:
+        if (package_root / 'pyproject.toml').exists():
+            break
+        if package_root == package_root.parent:
+            break
+        package_root = package_root.parent
+
+    relative_filename = abs_filename.relative_to(package_root)
     relative_filename = relative_filename.with_suffix("")
 
     # handle special cases
--- a/libcst/helpers/module.py
+++ b/libcst/helpers/module.py
@@ -142,7 +142,7 @@ def calculate_module_and_package(
         abs_filename = Path(filename).resolve()
         package_root = abs_filename.parent
         while package_root != abs_repo_root:
-            if (package_root / 'pyproject.toml').exists():
+            if (package_root / "pyproject.toml").exists():
                 break
             if package_root == package_root.parent:
                 break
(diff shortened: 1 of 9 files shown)


zsol (changes requested):
I have a few concerns with this approach:
* `calculate_module_and_package` shouldn't do any IO.
* the location of a `pyproject.toml` is a heuristic that shouldn't be part of the core logic, it should be behind a switch that can be enabled/disabled
* I'd argue that the default behavior should be the old one

Maybe it's better if the passed in `repo_root` has already applied this heuristic if it's desirable.

camillol:
@zsol thanks for taking a look! For context, my use case is a single repository containing many interdependent packages. Because of the interdependencies, it needs to be analyzed as a whole (e.g. I need information on the entire class hierarchy, which spans multiple packages).

AFAICT, the main user of `calculate_module_and_package` (and the one I'm interested in) is `FullyQualifiedNameProvider`. It works with `FullRepoManager`, and the `gen_cache` method receives only the repo root path and a list of file paths to process.

So, it's not feasible for me to have a separate repo root per package, and I don't see a way to feed the package information into FQNP.gen_cache given the current signature.

AFAICT, `gen_cache` is meant to do I/O, and does I/O in other providers (e.g. TypeInferenceProvider). I assume the requirement for `calculate_module_and_package` not to do I/O is for testing purposes. I could make a variant of this function and have FQNP call it, although it could cause problems down the road if something else decides to call `calculate_module_and_package` and gets inconsistent results with FQNP.

So I think the best approach would be to have a global switch. Or perhaps an option on the `FullRepoManager`, but in that case I'd still have to change the provider interface a bit to be able to wire it through.

Regarding the default behavior, I'd argue that the existing behavior of inferring the package from the file path relative to the repo root is also a heuristic, and for a repo that contains pyproject.toml files it is less likely to give the intended results than what I am proposing. But I'm fine with pyproject.toml parsing not being the default.

zsol:
> AFAICT, gen_cache is meant to do I/O, and does I/O in other providers (e.g. TypeInferenceProvider).

That's right. I propose to do the `pyproject.toml` finding in FQNP's `gen_cache`, and then pass in a `root_path` to `calculate_module_and_package` based on that.

My main concern is that `calculate_module_and_package` is not just an internal implementation detail, it's exposed in `libcst.helpers` and other code (outside of libcst) might be using it.

> Regarding the default behavior, I'd argue that the existing behavior of inferring the package from the file path relative to the repo root is also a heuristic, and for a repo that contains pyproject.toml files it is less likely to give the intended results than what I am proposing. But I'm fine with pyproject.toml parsing not being the default.

Agreed that the existing behavior is also just a heuristic, but I'm not convinced that the `pyproject.toml`-based heuristic is strictly better (or worse, it's just different), and so I don't think it warrants a backwards incompatible change. Let's make it configurable.

camillol:
Sounds good. I don't see an existing API for this kind of global configuration. Is it ok if I keep it simple and just make it a global variable?

camillol:
Oh, I'm not using the CLI at all. I'm using LibCST as a library from another program, so that config file doesn't apply.

`gen_cache` already has an optional argument that most implementations ignore (`timeout`). I could generalize that as `kwargs`, and add a keyword argument `use_pyproject`, provided by `FullRepoManager`.

zsol:
Sounds good, then!

camillol:
@zsol I think it's done, PTAL!

zsol (approved):
Thanks!

Issue #457, opened 2021-02-09
Remove `with self.assertRaises(Exception):`


jimmylai:
`with self.assertRaises(Exception):` is used in a couple places for testing an exception is raised.
It can catch any Exception which can result in catching unexpected error.
Change to use a more specific exception or use `self.assertRaisesRegex` instead.

srisatya1197:
@jimmylai will work on this issue.

PR #1168, opened 2024-07-04
Avoid raising bare Exception

zaicruvoir1rominet:
## Summary

Before tackling issue #457, there are some bare Exceptions thrown from within the code.

Here is the logic for changes in Exceptions:
* use `ValueError` when issue caused by user input
* use `CSTValidationError` & `TypeError` when issue occurs during the CST validation process
* use `ParserSyntaxError` when issue occurs during the parsing process 
* use `CSTLogicError` to replace all Exception which contained "Logic error" in the cause message
* remove Exception catch+re-raise when not needed

*I'm all for changes if some Exceptions changes are not OK,  
In particular, there are some `ParserSyntaxError` which may not be meaningful - with raw_line=0, raw_column=0 params*

## Test Plan

Keep the current test plan.

### DIFF
--- a/libcst/_excep.py
+++ b/libcst/_excep.py
@@ -1,8 +1,0 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
-#
-# This source code is licensed under the MIT license found in the
-# LICENSE file in the root directory of this source tree.
-
-
-class CSTLogicError(Exception):
-    pass
--- a/libcst/_excep.py
+++ b/libcst/_excep.py
@@ -1,0 +1,2 @@
+class CSTLogicError(Exception):
+    pass
--- a/libcst/_excep.py
+++ b/libcst/_excep.py
@@ -1,2 +1,7 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
 class CSTLogicError(Exception):
     pass
--- a/libcst/_excep.py
+++ b/libcst/_excep.py
@@ -3,5 +3,6 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+
 class CSTLogicError(Exception):
     pass
--- a/libcst/codemod/visitors/_add_imports.py
+++ b/libcst/codemod/visitors/_add_imports.py
@@ -7,7 +7,7 @@ from collections import defaultdict
 from typing import Dict, List, Optional, Sequence, Set, Tuple, Union
 
 import libcst
-from libcst import matchers as m, parse_statement
+from libcst import matchers as m, parse_statement, CSTLogicError
 from libcst._nodes.statement import Import, ImportFrom, SimpleStatementLine
 from libcst.codemod._context import CodemodContext
 from libcst.codemod._visitor import ContextAwareTransformer
@@ -107,7 +107,7 @@ class AddImportsVisitor(ContextAwareTransformer):
     ) -> List[ImportItem]:
         imports = context.scratch.get(AddImportsVisitor.CONTEXT_KEY, [])
         if not isinstance(imports, list):
-            raise Exception("Logic error!")
+            raise CSTLogicError()
         return imports
 
     @staticmethod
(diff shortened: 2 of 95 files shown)


kiri11:
This seems like an improvement, thank you. It's really hard to catch a specific `Exception`, I have to rely on it's message. I hope this gets merged.

libcst/_excep.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+class CSTLogicError(Exception):
+    pass
```
zsol, line 8: This should go in `_exceptions.py`, and please write a docstring

libcst/codemod/visitors/_add_imports.py
```diff
@@ -136,7 +136,7 @@ def add_needed_import(
         """
 
         if module == "__future__" and obj is None:
-            raise Exception("Cannot import __future__ directly!")
+            raise ImportError("Cannot import __future__ directly!")
```
zsol, line 139: I believe this (and the two others below) should be `ValueError`s

libcst/codemod/visitors/_remove_imports.py
```diff
@@ -45,7 +46,7 @@ def _remove_imports_from_importfrom_stmt(
             self.context.full_package_name, import_node
         )
         if module_name is None:
-            raise Exception("Cannot look up absolute module from relative import!")
```
zsol, line 48: these two should also be `ValueErrors` IMO

libcst/codemod/visitors/_remove_imports.py
```diff
@@ -255,7 +258,9 @@ def remove_unused_import_by_node(
                 context.full_package_name, node
             )
             if module_name is None:
-                raise Exception("Cannot look up absolute module from relative import!")
+                raise ImportError(
```
zsol, line 261: `ValueError`

libcst/metadata/type_inference_provider.py
```diff
@@ -60,17 +60,15 @@ def gen_cache(
...
-            raise Exception(f"stderr:\n {stderr}\nstdout:\n {stdout}")
+        result = subprocess.run(cmd_args, capture_output=True, timeout=timeout)
+        result.check_returncode()
+        stdout, stderr = result.stdout.decode(), result.stderr.decode()
+
         try:
             resp = json.loads(stdout)["response"]
         except Exception as e:
-            raise Exception(f"{e}\n\nstderr:\n {stderr}\nstdout:\n {stdout}")
+            raise Exception(f"{e}\n\nstderr:\n {stderr}\nstdout:\n {stdout}") from e
```
zsol, line 71: We could have an exception class just for type inference provider here

zsol (changes requested):
Thanks for this! I've got a few minor comments but overall looks great!

libcst/_excep.py
zaicruvoir1rominet: Thanks for the review !

I'm going to need help on this one though:

- It's in a separate `_excep.py` file, since `_exceptions.py` already imports internals, which in turn may raise a `CSTLogicError`.  
-

zaicruvoir1rominet:
Thanks for the support @kiri11 !
You also made me realize some users were perhaps using exception messages in order to manage errors, so I kept all original Exception messages in order to avoid breaking changes.

I hope everything is now ready (apart from that `CSTLogicError` file location & docstring)

kiri11:
You're welcome! (although I didn't do anything) 
Keeping the exception messages backwards-compatible is a great idea, thank you!

libcst/_excep.py
zaicruvoir1rominet: @zsol 
Any updates ?

zsol: > _exceptions.py already imports internals, which in turn may raise a CSTLogicError.

Let's move `get_expected_str` out of `_exceptions.py`. That should help.

> For the docstring, I couldn't think of anything meaningful

How about:
```
General purpose internal error within LibCST itself.
```

zsol (approved):

PR #1169, opened 2024-07-04
Mention codemod -x flag in docs

kiri11:
## Summary

After https://github.com/Instagram/LibCST/pull/879 got implemented, we can assume that most user codemods are loaded in this way instead of editing the LibCST source (it would be just pip installed). So `-x` flag is very important and should be mentioned in docs. 

I am open to suggestions on how to improve the wording and also mention the built-in codemods and `libcst.tool list` better.

## Test Plan

#shipit (just docs)


zsol (approved):
Sounds good, thanks! Note that you could always have loaded Codemods from arbitrary locations as long as you had a config file in the root of your repo and then listed the locations of the Codemods there.

But yeah with `-x` even that is not necessary anymore.

PR #1173, opened 2024-07-18
Enable writing module code directly to file

asvarga-sedaro:
## Summary

Enables generating code directly into files:

```py
with open('code.py', 'w') as f:
    module.write_code(f)
```

rather than

```py
code = module.code
with open('code.py', 'w') as f:
    f.write(code)
```

It does this with a subclass of `CodegenState` that avoids building in memory:

- the `tokens: List[str]` field of `CodegenState`
- the `"".join(state.tokens)` string of code

## Test Plan

- `hatch run test` passes
- we could add some tests that compare the contents of written files written directly vs indirectly

## Notes

- `CodegenState.pop_trailing_newline()` isn't supported because the token is already written to the file. this is fixable if desired.

### DIFF
--- a/libcst/_nodes/internal.py
+++ b/libcst/_nodes/internal.py
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 
+import io
 from contextlib import contextmanager
 from dataclasses import dataclass, field
 from typing import Iterable, Iterator, List, Optional, Sequence, TYPE_CHECKING, Union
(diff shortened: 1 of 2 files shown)


libcst/_nodes/internal.py
```diff
@@ -70,6 +71,30 @@ def record_syntactic_position(
         yield
 
 
+@add_slots
+@dataclass(frozen=False)
+class CodegenWriter(CodegenState):
```
zsol, line 76: Seems like the class hierarchy should be the other way around. `CodegenState` should specialize `CodegenWriter`.

Then we wouldn't have to have an awkward default value for `writer`.

libcst/_nodes/internal.py
```diff
@@ -70,6 +71,30 @@ def record_syntactic_position(
...
+
+    def add_indent_tokens(self) -> None:
+        for token in self.indent_tokens:
+            self.writer.write(token)
+
+    def add_token(self, value: str) -> None:
+        self.writer.write(value)
+
+    def pop_trailing_newline(self) -> None:
+        pass
```
zsol, line 95: this is incorrect, code that doesn't end in a trailing newline won't be serialized properly (for example the python module `""` will be serialized as `"\n"`)

zsol (changes requested):
Please write tests exercising the new code

PR #1177, opened 2024-07-27
Add validation for If node

kiri11:
Don't allow no space no parentheses after `if`.

## Summary

There should be a validation for `If` node similar to others such as `While` or `Assert` to not allow generating incorrect code.
There is even a TODO left for that reason.

Let me know if there should be other validations for `If`.

### Before:

With the current version of LibCST it is possible to generate the code like this with an `If` node followed by `Name('x')` without space:
```python
ifx:
    pass
```
which isn't correct Syntax. 

The new test fails:
```python
======================================================================
FAIL: test_invalid_0 (libcst._nodes.tests.test_if.IfTest.test_invalid_0)
  File "libcst\testing\utils.py", line 88, in new_test
    return member(self, *data)
           ^^^^^^^^^^^^^^^^^^^
  File "libcst\_nodes\tests\test_if.py", line 149, in test_invalid
    self.assert_invalid(get_node, expected_re)
  File "libcst\_nodes\tests\base.py", line 95, in assert_invalid
    with self.assertRaisesRegex(cst.CSTValidationError, expected_re):
AssertionError: CSTValidationError not raised

```

### After:
Attempting to generate incorrect code raises an exception preventing it:
```
libcst._nodes.base.CSTValidationError: Must have at least one space after 'if' keyword.
```

The new test passes:
```python
Ran 2076 tests in 62.315s

OK (skipped=18, expected failures=2)
```

### DIFF
--- a/libcst/_nodes/statement.py
+++ b/libcst/_nodes/statement.py
@@ -599,7 +599,12 @@ class If(BaseCompoundStatement):
     #: The whitespace appearing after the test expression but before the colon.
     whitespace_after_test: SimpleWhitespace = SimpleWhitespace.field("")
 
-    # TODO: _validate
+    def _validate(self) -> None:
+        if (
+            self.whitespace_before_test.empty
+            and not self.test._safe_to_use_with_word_operator(ExpressionPosition.RIGHT)
+        ):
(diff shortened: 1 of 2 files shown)


zsol:
✨ Thanks! ✨

Issue #1183, opened 2024-08-01
Warnings persist in Context


jvansanten:
I've been playing around with using libcst to copy (newly added) annotations from base class stubs to subclasses in a large legacy codebase. It's pretty neat!

There are some edge cases in this that I can't handle, and so kick them up to the user with `Codemod.warn()`. There is, however a minor UX issue when processing multiple files in one invocation: once I `warn()`, the warning persists in the `Context` for all subsequent files processed by that instance of the `Codemod`, and so I see spurious warnings attached to the wrong filename:
```console
> python3 -m libcst.tool codemod --hide-blacklisted-warnings -x icetray_codemod.ApplyStubsToSubclasses --stubs ../codemod/stubs.pkl --exclude-patterns 'icecube\.gulliver' sim-services
Calculating full-repo metadata...
Executing codemod...
1.82s 6% complete, 27.87s estimated for 46 files to go...Found 1 error (1 fixed, 0 remaining).
Codemodding /Users/jakob/Documents/IceCube/metaprojects/icetray/src/sim-services/python/gcd_validation/gcd_sanity_checker.py
WARNING: Mismatched arguments. icecube.icetray.I3Module.Finish has 1/0/0 pos-only/normal/kw-only params, I3GCDSanityChecker.Finish has 0/2/0 
Successfully codemodded /Users/jakob/Documents/IceCube/metaprojects/icetray/src/sim-services/python/gcd_validation/gcd_sanity_checker.py with warnings

Codemodding /Users/jakob/Documents/IceCube/metaprojects/icetray/src/sim-services/python/gcd_validation/hit_generator.py
WARNING: Mismatched arguments. icecube.icetray.I3Module.Finish has 1/0/0 pos-only/normal/kw-only params, I3GCDSanityChecker.Finish has 0/2/0 
Successfully codemodded /Users/jakob/Documents/IceCube/metaprojects/icetray/src/sim-services/python/gcd_validation/hit_generator.py with warnings

Codemodding /Users/jakob/Documents/IceCube/metaprojects/icetray/src/sim-services/python/gcd_validation/pe_generator.py
WARNING: Mismatched arguments. icecube.icetray.I3Module.Finish has 1/0/0 pos-only/normal/kw-only params, I3GCDSanityChecker.Finish has 0/2/0 
Successfully codemodded /Users/jakob/Documents/IceCube/metaprojects/icetray/src/sim-services/python/gcd_validation/pe_generator.py with warnings

[...]

PR #1184, opened 2024-08-03
Clear warnings for each file in codemod cli

kiri11:
## Summary

Original idea belongs to @jschavesr. I am just resubmitting his abandoned [PR](https://github.com/Instagram/LibCST/pull/665).
Fixes the issue: https://github.com/Instagram/LibCST/issues/1183

Context warning messages were not being cleaned between files, so warnings from one files could be passed to other files in the context. Because of that warnings were being shown wrongly in the summary and the total number of warnings was wrong as well.

I also [encountered ](https://github.com/Instagram/LibCST/actions/runs/10231310110/job/28307117717?pr=1184)another error in Windows CI. Seems like codemodding small files happens faster than the timer counts more than zero (probably less precision on Windows). So fixing that, too.

```python
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "D:\a\LibCST\LibCST\libcst\tool.py", line 689, in <module>
    main(os.environ.get("LIBCST_TOOL_COMMAND_NAME", "libcst.tool"), sys.argv[1:])
  File "D:\a\LibCST\LibCST\libcst\tool.py", line 684, in main
    return lookup.get(args.action or None, _invalid_command)(proc_name, command_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\a\LibCST\LibCST\libcst\tool.py", line 420, in _codemod_impl
    result = parallel_exec_transform_with_prettyprint(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\a\LibCST\LibCST\libcst\codemod\_cli.py", line 659, in parallel_exec_transform_with_prettyprint
    progress.print(successes + failures + skips)
  File "D:\a\LibCST\LibCST\libcst\codemod\_cli.py", line 398, in print
    f"{self.ERASE_CURRENT_LINE}{self._human_seconds(elapsed_time)} {percent:.{self.pretty_precision}f}% complete, {self.estimate_completion(elapsed_time, finished, left)} estimated for {left} files to go...",
                                                                                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\a\LibCST\LibCST\libcst\codemod\_cli.py", line 430, in estimate_completion
    fps = files_finished / elapsed_seconds
          ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
ZeroDivisionError: float division by zero
```

### Before:

Running `convert_format_to_fstring.ConvertFormatStringCommand` on three files with identical contents:
```python
def baz() -> str:
    return "{}: {}".format(*baz)
```
All three are the same, so they should get one warning each.
But instead every following file gets more warnings copied from previous files:
```shell
LibCST> python -m libcst.tool codemod convert_format_to_fstring.ConvertFormatStringCommand tmp
Calculating full-repo metadata...
Executing codemod...
Codemodding tmp\mod1.py
WARNING: Unsupported field_name 0 in format() call
Successfully codemodded tmp\mod1.py with warnings

Codemodding tmp\mod2.py
WARNING: Unsupported field_name 0 in format() call
WARNING: Unsupported field_name 0 in format() call
Successfully codemodded tmp\mod2.py with warnings

Codemodding tmp\mod3.py
WARNING: Unsupported field_name 0 in format() call
WARNING: Unsupported field_name 0 in format() call
WARNING: Unsupported field_name 0 in format() call
Successfully codemodded tmp\mod3.py with warnings

Finished codemodding 3 files!
 - Transformed 3 files successfully.
 - Skipped 0 files.
 - Failed to codemod 0 files.
 - 6 warnings were generated.
```

The new test fails:

```python
FAIL: test_warning_messages_several_files (libcst.codemod.tests.test_codemod_cli.TestCodemodCLI.test_warning_messages_several_files)
  File "libcst\codemod\tests\test_codemod_cli.py", line 92, in test_warning_messages_several_files
    self.assertIn(
AssertionError: '- 3 warnings were generated.' not found in "Calculating full-repo metadata... ... \n - 6 warnings were generated.\n"
```

### After:

Each file generates one warning correctly:
```shell
LibCST> python -m libcst.tool codemod convert_format_to_fstring.ConvertFormatStringCommand tmp
Calculating full-repo metadata...
Executing codemod...
Codemodding tmp\mod1.py
WARNING: Unsupported field_name 0 in format() call
Successfully codemodded tmp\mod1.py with warnings

Codemodding tmp\mod2.py
WARNING: Unsupported field_name 0 in format() call
Successfully codemodded tmp\mod2.py with warnings

Codemodding tmp\mod3.py
WARNING: Unsupported field_name 0 in format() call
Successfully codemodded tmp\mod3.py with warnings

Finished codemodding 3 files!
 - Transformed 3 files successfully.
 - Skipped 0 files.
 - Failed to codemod 0 files.
 - 3 warnings were generated.
```

And the new test passes.

### DIFF
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -423,7 +423,7 @@ class Progress:
         operations still to do.
         """
 
-        if files_finished <= 0:
+        if files_finished <= 0 or elapsed_seconds == 0:
             # Technically infinite but calculating sounds better.
             return "[calculating]"
 
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -22,6 +22,7 @@ from typing import Any, AnyStr, cast, Dict, List, Optional, Sequence, Union
 
 from libcst import parse_module, PartialParserConfig
 from libcst.codemod._codemod import Codemod
+from libcst.codemod._context import CodemodContext
 from libcst.codemod._dummy_pool import DummyPool
 from libcst.codemod._runner import (
     SkipFile,
@@ -246,33 +247,28 @@ def _execute_transform(  # noqa: C901
                 ),
             )
 
-        # Somewhat gross hack to provide the filename in the transform's context.
-        # We do this after the fork so that a context that was initialized with
-        # some defaults before calling parallel_exec_transform_with_prettyprint
-        # will be updated per-file.
-        # Clean the warnings as well, otherwise they will be
-        # passed from the previous file
-        transformer.context = replace(
-            transformer.context,
-            filename=filename,
-            scratch=deepcopy(scratch),
-            warnings=[],
-        )
-
         # determine the module and package name for this file
         try:
             module_name_and_package = calculate_module_and_package(
                 config.repo_root or ".", filename
             )
-            transformer.context = replace(
-                transformer.context,
-                full_module_name=module_name_and_package.name,
-                full_package_name=module_name_and_package.package,
-            )
+            mod_name = module_name_and_package.name
+            pkg_name = module_name_and_package.package
         except ValueError as ex:
             print(
                 f"Failed to determine module name for {filename}: {ex}", file=sys.stderr
             )
+            mod_name = None
+            pkg_name = None
+
+        # Apart from metadata_manager, every field of context should be reset per file
+        transformer.context = CodemodContext(
+            scratch=deepcopy(scratch),
+            filename=filename,
+            full_module_name=mod_name,
+            full_package_name=pkg_name,
+            metadata_manager=transformer.context.metadata_manager,
+        )
 
         # Run the transform, bail if we failed or if we aren't formatting code
         try:
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -250,10 +250,13 @@ def _execute_transform(  # noqa: C901
         # We do this after the fork so that a context that was initialized with
         # some defaults before calling parallel_exec_transform_with_prettyprint
         # will be updated per-file.
+        # Clean the warnings as well, otherwise they will be
+        # passed from the previous file
         transformer.context = replace(
             transformer.context,
             filename=filename,
             scratch=deepcopy(scratch),
+            warnings=[],
         )
 
         # determine the module and package name for this file
--- a/libcst/codemod/tests/test_codemod_cli.py
+++ b/libcst/codemod/tests/test_codemod_cli.py
@@ -8,10 +8,12 @@
 import platform
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 from unittest import skipIf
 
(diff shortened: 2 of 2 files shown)


libcst/codemod/_cli.py
```diff
@@ -250,10 +250,13 @@ def _execute_transform(  # noqa: C901
         # We do this after the fork so that a context that was initialized with
         # some defaults before calling parallel_exec_transform_with_prettyprint
         # will be updated per-file.
+        # Clean the warnings as well, otherwise they will be
+        # passed from the previous file
         transformer.context = replace(
             transformer.context,
             filename=filename,
             scratch=deepcopy(scratch),
+            warnings=[],
```
zsol, line 259: While this change will fix the issue, there's a deeper problem here. Apart from `metadata_manager`, every other field of `context` should be reset per file.
Concretely, I'm surprised we change `filename` here but not `full_module_name` or `full_package_name`.

Maybe we should make this explicit and have a global context that persists between files, and a local one that gets reset. Then this assignment would look something like:

```python
try:
    mod_and_pkg = calculate_module_and_package(config.repo_root or ".", filename)
    mod_name = mod_and_pkg.name
    pkg_name = mod_and_pkg.package
except ValueError as exc:
    print(f"Failed to determine module name for {filename}: {ex}", file=sys.stderr)
    mod_name = None
    pkg_name = None
    
transformer.context = CodemodContext(scratch=deepcopy(scratch), filename=filename, full_module_name=mod_name, full_package_name=pkg_name, metadata_manager=transformer.context.metadata_manager)

zsol (approved):

libcst/codemod/_cli.py
zsol: Having said that, I'm happy to merge the PR as is because it's definitely an improvement over the current state. Let me know if you wanna roll this change into this PR or not

kiri11: `full_module_name` or `full_package_name` are changed below, so currently context replaced twice. And if it isn't possible to determine the module/package name, it will persist from the previous file. The `wrapper` also persists. That would be changing in the new commit I'm submitting.

**Before:**
- `full_module_name`/`full_package_name` persist if can't get them for the new file
- `wrapper` persists across files

**After:**
- `full_module_name`/`full_package_name` both set to `None` if can't figure them out.
- `wrapper`  is set to `None` on every file.

Is that okay to not set the `wrapper`?

zsol: Yep, it should be

zsol:
Thanks!

PR #1192, opened 2024-08-21
Drop codecov from CI and readme

amyreese:
## Summary

Codecov keeps failing in CI. Chatted with @zsol and it doesn't seem worth the effort.

## Test Plan

CI should actually pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,12 +31,17 @@ jobs:
           toolchain: stable
       - name: Build LibCST
         run: hatch -vv env create
-      - name: Tests
+      - name: Native Parser Tests
         run: hatch run test
       - name: Pure Parser Tests
         env:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -89,13 +89,6 @@ jobs:
...
           hatch run coverage run setup.py test
           hatch run coverage xml -i
-      - uses: codecov/codecov-action@v4
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-        with:
-          files: coverage.xml
-          fail_ci_if_error: true
-          verbose: true
       - name: Archive Coverage
```
zsol, line 92: Will we ever look at this? Maybe worth deleting, plus the xml step above (but maybe keep the normal coverage run)

Issue #1181, opened 2024-07-31
[Windows, Mac] Can't use call_if_inside/call_if_not_inside matcher decorators with multiprocessing


kiri11:
This issue has been reported before (https://github.com/Instagram/LibCST/issues/848, https://github.com/Instagram/LibCST/issues/629, https://github.com/Instagram/LibCST/issues/435), but I debugged it and found the root cause, so decided to create another one with more information.

### Example traceback:
```python
Traceback (most recent call last):
  File "libcst/codemod/_cli.py", line 285, in _execute_transform
    output_tree = transformer.transform_module(input_tree)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "libcst/codemod/_command.py", line 71, in transform_module
    tree = super().transform_module(tree)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "libcst/codemod/_codemod.py", line 108, in transform_module
    return self.transform_module_impl(tree_with_metadata)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "libcst/codemod/_visitor.py", line 32, in transform_module_impl
    return tree.visit(self)
           ^^^^^^^^^^^^^^^^
  File "libcst/_nodes/module.py", line 89, in visit
    result = super(Module, self).visit(visitor)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "libcst/_nodes/base.py", line 227, in visit
    _CSTNodeSelfT, self._visit_and_replace_children(visitor)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "libcst/_nodes/statement.py", line 1814, in _visit_and_replace_children
    body=visit_required(self, "body", self.body, visitor),
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "libcst/_nodes/internal.py", line 81, in visit_required
    result = node.visit(visitor)
             ^^^^^^^^^^^^^^^^^^^
  File "libcst/_nodes/base.py", line 236, in visit
    leave_result = visitor.on_leave(self, with_updated_children)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1204, opened 2024-09-02
fix certain matchers breaking under multiprocessing by initializing them late

kiri11:
Support matcher decorators with multiprocessing on Windows/macOS.

See issue #1181 for more context.

## Summary

- Call `_gather_matchers` later in the code (first time when it's needed) to ensure we initialize them in a process AFTER the "fork".
- Proactively check if attributes are not properties instead of trying to evaluate them and suppressing all errors.

### Before
```
> hatch run python -m unittest libcst.codemod.tests.test_codemod_cli.TestCodemodCLI.test_matcher_decorators_multiprocessing

FAILED (failures=1)
```

### After
```
> hatch run python -m unittest libcst.codemod.tests.test_codemod_cli.TestCodemodCLI.test_matcher_decorators_multiprocessing     
.
Ran 1 test in 0.973s

OK
```

### DIFF
--- a/libcst/matchers/_visitors.py
+++ b/libcst/matchers/_visitors.py
@@ -270,20 +270,22 @@ def _check_types(
             )
 
 
-def _gather_matchers(obj: object) -> Set[BaseMatcherNode]:
-    visit_matchers: Set[BaseMatcherNode] = set()
+def _gather_matchers(obj: object) -> Dict[BaseMatcherNode, Optional[cst.CSTNode]]:
+    """
+    Set of gating matchers that we need to track and evaluate. We use these
+    in conjunction with the call_if_inside and call_if_not_inside decorators
+    to determine whether to call a visit/leave function.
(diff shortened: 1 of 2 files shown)


libcst/matchers/_visitors.py
```diff
@@ -366,6 +372,8 @@ def _visit_matchers(
     node: cst.CSTNode,
     metadata_resolver: cst.MetadataDependent,
 ) -> Dict[BaseMatcherNode, Optional[cst.CSTNode]]:
+    # Populate matchers dict if it's empty
+    matchers = matchers or _gather_matchers(metadata_resolver)
```
zsol, line 376: Looks like we'll pay extra on every `on_visit()` now if there are no matchers declared in the visitor/transformer. Would be worth eliminating that

zsol (approved):
Thank you!

PR #1213, opened 2024-09-25
Make rust build deterministic

bmwiedemann:
## Summary

We disable LTO and use `codegen-units = 1` for deterministic build results of the rust code.
See https://reproducible-builds.org/ for why this is good.

https://github.com/rust-lang/rust/issues/128675 states that the default `codegen-units = 16` should be deterministic, but here it clearly is not.

### DIFF
--- a/native/Cargo.toml
+++ b/native/Cargo.toml
@@ -4,3 +4,8 @@ members = [
     "libcst",
     "libcst_derive",
 ]
+
+# needed for reproducible builds from some hard-to-grasp issue in rust or LLVM
+# = https://github.com/rust-lang/rust/issues/128675
+[profile.release]
+lto = "off"


zsol:
Hey @bmwiedemann thanks for the PR!

> here it clearly is not

Could you help me understand why not? How can I reproduce this nondeterminism?

bmwiedemann:
I just did two clean builds as similar as possible. Same CPU, same username, same path, same timezone, locale, umask, date... and it still differed in the `/usr/lib64/python3.12/site-packages/libcst/native.cpython-312-x86_64-linux-gnu.so`

https://rb.zq1.de/compare.factory-20240904/diffs/python-libcst-compare.out shows a (partially filtered) diff.

bmwiedemann:
hmm. For some reason the variations are gone now with rust-1.81.0

PR #1229, opened 2024-10-17
`libcst.tool`: add support Python 3.9, 3.10, 3.11, 3.12 and 3.13

jorenham:
Currently, running e.g. `libcst.tool codemod --python-version=3.9 ...` will raise a `ValueError` with the message:

```
ValueError: LibCST can only parse code using one of the following versions of Python's grammar: 3.0, 3.1, 3.3, 3.5, 3.6, 3.7, 3.8. More versions may be supported by future releases.
```

This adds the missing supported python version strings.

### DIFF
--- a/libcst/_parser/types/config.py
+++ b/libcst/_parser/types/config.py
@@ -44,7 +44,7 @@ class AutoConfig(Enum):
 
 
 # This list should be kept in sorted order.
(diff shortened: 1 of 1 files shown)


zsol:
Unfortunately that flag doesn't actually do anything anymore by default[^1], so I'd rather remove it.

[^1]: It only controls the behavior of the old, unmaintained pure python parser which you can enable with an environment variable. That parser doesn't support 3.9 or newer.

PR #1257, opened 2024-12-07
Add a codemod for dataclass changes in 3.11

zsol:
Add a codemod for dataclass changes in 3.11

Python 3.11 changed the field default mutability check for dataclasses to only allow defaults which are hashable.

This codemod helps with the migration by changing all default values that aren't obviously hashable to use `default_factory` instead.

Note: it's impossible to accurately determine if a particular expression produces a hashable value in a codemod, so the codemod significantly over-approximates what's unhashable.

### DIFF
--- a/libcst/codemod/commands/dataclass_default_factory.py
+++ b/libcst/codemod/commands/dataclass_default_factory.py
@@ -1,0 +1,140 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+# pyre-strict
+import builtins
+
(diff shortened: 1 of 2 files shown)


itamaro (approved):
Is it too cumbersome to identify zero-args-calls used as mutable defaults and drop the lambda in those cases? 
Similarly for empty list/dict literals, using `list` / `dict` instead of the lambda

zsol:
@itamaro no, not at all. If you wanna take a stab at it, feel free to commandeer this PR :)

itamaro:
> If you wanna take a stab at it, feel free to commandeer this PR :)

don't have the time right now, and it's useful enough to merge as is!
we can iterate on it in future PRs.

PR #1267, opened 2024-12-30
Add codemod to rename typing aliases of builtins

yangdanny97:
## Summary

This PR adds a codemod for Python 3.9+ that updates references like `typing.List` to just `list`. It includes all the builtins that have an alias in the typing module.

We can also do the same thing for some non-builtin types in the collections module (the full list is in https://peps.python.org/pep-0585/#implementation) but I'm not sure if that should be a separate codemod or added to this one.

I also had to update the RenameCommand to avoid adding imports if the new name is a builtin.

## Test Plan

hatch run python -m unittest libcst.codemod.commands.tests.test_rename
hatch run python -m unittest libcst.codemod.commands.tests.test_rename_typing_generic_aliases


zsol:
Thanks! :)

PR #1274, opened 2025-01-08
Expose TypeAlias and TypeVar related structs in rust library

Crozzers:
## Summary

This PR re-exports a number of structs related to type aliases and type vars in the rust library. The use case for this is making it easier to traverse a parsed `Module`.

You can traverse `SmallStatement::TypeAlias(value) => value.type_parameters[0].param` all without directly referring to those types (ie: not needing to import them) but `TypeParam.param` is an enum. You can't do much with that without matching against the enum values to determine the actual type, which requires them to be imported.

## Test Plan

Tested locally and was able to use the new exports.


zsol:
Thanks!

PR #1289, opened 2025-01-28
Update for Pyo3 0.23

ngoldbaum:
## Summary

Updates the native library to use PyO3 0.23. This is the first step towards declaring support for the free-threaded interpreter.

I could have gone further in ripping out `TryIntoPy` in favor of `IntoPyObject`. If you'd prefer not to leave this project in a state with an unnecessary wrapper trait I'm happy to implement that, I just wanted to make the smallest possible working diff.

## Test Plan

This passes the tests for me locally. Let's see if CI spots and issues.

### DIFF
--- a/native/libcst/src/nodes/traits.rs
+++ b/native/libcst/src/nodes/traits.rs
@@ -118,7 +118,7 @@ impl<'a, T: Inflate<'a>> Inflate<'a> for Vec<T> {
 }
 #[cfg(feature = "py")]
 pub mod py {
-    use pyo3::{types::PyAny, types::PyTuple, IntoPy, PyObject, PyResult, Python};
+    use pyo3::{types::PyTuple, IntoPyObjectExt, PyObject, PyResult, Python};
 
     // TODO: replace with upstream implementation once
     // https://github.com/PyO3/pyo3/issues/1813 is resolved
@@ -135,7 +135,7 @@ pub mod py {
 
     impl TryIntoPy<PyObject> for bool {
         fn try_into_py(self, py: Python) -> PyResult<PyObject> {
-            Ok(self.into_py(py))
+            self.into_py_any(py)
         }
     }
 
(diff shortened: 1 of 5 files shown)


ngoldbaum:
I signed the CLA.

native/libcst/src/nodes/traits.rs
```diff
@@ -170,28 +170,13 @@ pub mod py {
                 .map(|x| x.try_into_py(py))
                 .collect::<PyResult<Vec<_>>>()?
                 .into_iter();
-            Ok(PyTuple::new_bound(py, converted).into())
-        }
-    }
-
-    impl TryIntoPy<PyObject> for PyTuple {
```
ngoldbaum, line 177: I found that with IntoPyObject that some of these impls aren't needed at all. I didn't look too closely to try to understand why though so there could be something subtle happening.

native/libcst/src/parser/errors.rs
```diff
@@ -59,36 +52,21 @@ mod py_error {
...
+                        ("raw_line", (line + 1).into_pyobject(py)?.into_any()),
+                        ("raw_column", col.into_pyobject(py)?.into_any()),
+                    ]
+                    .into_py_dict(py)?;
+                    let libcst = PyModule::import(py, "libcst")?;
+                    libcst.getattr("ParserSyntaxError")?.call((), Some(&kwargs))
+                }() {
+                    Ok(py_err_value) => PyErr::from_value(py_err_value),
+                    Err(e) => e,
+                }
```
ngoldbaum, line 68: Since ultimately we want a PyErr in both the success and error cases, we can make use of a closure that returns a result and `?` to avoid `unwrap` or `expect` in the new `into_pyobject()` calls that can return errors and at the same time clean up some unnecessary pre-existing `expect` use to turn some panics into errors. This is a neat Rust trick, unfortuntely it's only really convenient in error handling code.

native/libcst/src/parser/errors.rs
```diff
@@ -59,36 +52,21 @@ mod py_error {
...
             })
         }
     }
-
-    impl<'a> PyErrArguments for Details {
```
ngoldbaum, line 82: Both this and `Details` are unused and the compiler warns about them.

native/libcst/src/nodes/traits.rs
zsol: Great, I was hoping to get rid of these.

native/libcst/src/nodes/traits.rs
```diff
@@ -170,28 +170,13 @@ pub mod py {
...
-        fn try_into_py(self, py: Python) -> PyResult<PyObject> {
-            Ok(self.into_py(py))
+            PyTuple::new(py, converted).unwrap().into_py_any(py)
```
zsol, line 173: Anything we can do to avoid the `unwrap()` and potential panic here?

zsol (approved):
LGTM in general, thanks for this

zsol:
Ugh, apologies for the churn, my tools messed with the PR - I did my best to revert my changes

ngoldbaum:
Sorry for the delay here - I accidentally marked the TODO item I had for this as "done". Will try to deal with the `unwrap()` calls today.

zsol:
🎉 thanks!

PR #1294, opened 2025-02-22
Replace multiprocessing with ProcessPoolExecutor

zsol:
Replace multiprocessing with ProcessPoolExecutor

Instead of relying on `multiprocessing.Pool`, this PR replaces the implementation of `parallel_exec_transform_with_prettyprint` with `concurrent.futures.ProcessPoolExecutor`

### DIFF
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -14,16 +14,17 @@ import subprocess
 import sys
 import time
 import traceback
+from concurrent.futures import as_completed, Executor, ProcessPoolExecutor
 from copy import deepcopy
 from dataclasses import dataclass, replace
-from multiprocessing import cpu_count, Pool
+from multiprocessing import cpu_count
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


amyreese (approved):
Seems reasonable. 

Do you think it would be worth dumping the custom "walk paths" and "run things on a executor" bits in favor of using [trailrunner](https://pypi.org/project/trailrunner/).  This is what we use for usort/ufmt/fixit/etc, and it ensures consistent behavior across OS (ie, spawn-only, no fork on Linux), obeys gitignore, etc.  The "dummy pool" stuff could just be a matter of passing `ThreadPoolExecutor` as the default executor if needed, but we could also add an optimization in trailrunner instead.

zsol:
I don't know if it's worth the effort - I found it's really nice not to have any dependencies. On the other hand, this frontend piece could use a lot of rework to make it easier to bootstrap your own single purpose codemod tool

PR #1295, opened 2025-02-22
Enable support for free-threading

zsol:
Enable support for free-threading

This PR:
1. marks the `libcst.native` module as free-threading-compatible
2. replaces the use of ProcessPoolExecutor with ThreadPoolExecutor if free-threaded CPython is detected at runtime

This depends on #1294 and #1289.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/Instagram/LibCST/pull/1295).
* #1333
* __->__ #1295

### DIFF
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -8,18 +8,19 @@ Provides helpers for CLI interaction.
 """
 
 import difflib
+import functools
 import os.path
 import re
 import subprocess
 import sys
 import time
 import traceback
-from concurrent.futures import as_completed, Executor, ProcessPoolExecutor
+from concurrent.futures import as_completed, Executor
 from copy import deepcopy
 from dataclasses import dataclass
 from multiprocessing import cpu_count
 from pathlib import Path
-from typing import AnyStr, cast, Dict, List, Optional, Sequence, Type, Union
+from typing import AnyStr, Callable, cast, Dict, List, Optional, Sequence, Type, Union
 from warnings import warn
 
 from libcst import parse_module, PartialParserConfig
@@ -624,14 +625,20 @@ def parallel_exec_transform_with_prettyprint(  # noqa: C901
         python_version=python_version,
     )
 
-    pool_impl: type[Executor]
+    pool_impl: Callable[[], Executor]
     if total == 1 or jobs == 1:
         # Simple case, we should not pay for process overhead.
         # Let's just use a dummy synchronous executor.
         jobs = 1
         pool_impl = DummyExecutor
+    elif getattr(sys, "_is_gil_enabled", lambda: False)():  # pyre-ignore[16]
+        from concurrent.futures import ThreadPoolExecutor
+
+        pool_impl = functools.partial(ThreadPoolExecutor, max_workers=jobs)
     else:
-        pool_impl = ProcessPoolExecutor
+        from concurrent.futures import ProcessPoolExecutor
+
+        pool_impl = functools.partial(ProcessPoolExecutor, max_workers=jobs)
         # Warm the parser, pre-fork.
         parse_module(
             "",
@@ -650,7 +657,7 @@ def parallel_exec_transform_with_prettyprint(  # noqa: C901
         deepcopy(transform.context.scratch) if isinstance(transform, Codemod) else {}
     )
 
-    with pool_impl(max_workers=jobs) as executor:  # type: ignore
+    with pool_impl() as executor:  # type: ignore
         try:
             futures = [
                 executor.submit(
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,13 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.13t"]
     steps:
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+      - name: Install hatch
+        run: pip install -U hatch
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
@@ -24,9 +29,6 @@ jobs:
           cache: pip
           cache-dependency-path: "pyproject.toml"
           python-version: ${{ matrix.python-version }}
-      - name: Install hatch
-        run: |
-          pip install -U hatch
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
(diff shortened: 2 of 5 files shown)


libcst/codemod/_cli.py
```diff
@@ -629,7 +638,7 @@ def parallel_exec_transform_with_prettyprint(  # noqa: C901
     warnings: int = 0
     skips: int = 0
 
-    with pool_impl(processes=jobs) as p:  # type: ignore
+    with pool_impl() as executor:  # type: ignore
         args = [
             {
                 "transformer": transform,
```
ngoldbaum, line 644: won't `transform.context` get clobbered here:

https://github.com/Instagram/LibCST/blob/985cec808e6bf2ed859e3f7cdc2da7710ba32776/libcst/codemod/_cli.py#L264-L271

This is mostly safe with the GIL, so long as python doesn't switch to another thread that also owns a reference to `transform`, but in free-threading with the sort of parallel thread pool setup you have here it's much more likely to happen.

In general - how do you want to go about adding multithreaded tests? It looks like there's a decent amount of stateful objects that get passed around and updated. How much should we worry about threads simultaneously updating stateful objects?

zsol: oh yeah absolutely - we'll probably have to change how state is passed around between threads. In fact, it might be better to just make a new transformer instance for each file and only pass immutable data structures between threads/processes.

> In general - how do you want to go about adding multithreaded tests?

We could run the test suite with something like `unittest-ft`, but the main thing I'd look for is the concurrent execution of the same set of visitors/transformers on multiple input files (basically what this PR sketches out), and then any bugs we find we can write targeted tests. How does that sound?

ngoldbaum: > How does that sound?

Sounds great, that's more or less what I've been doing with `pyca/cryptography` this week: https://github.com/pyca/cryptography/pull/12555

Another thing I learned there is moving code form pure python to Rust can also be a way to avoid thread safety issues. If we move state from Python to Rust then we have much more control over concurrent multithreaded access.

I'll look at this next week - thanks for merging my other PR!

ngoldbaum:
> I'll look at this next week - thanks for merging my other PR!

Some other things came up but I'm going to come back to this next week.

ngoldbaum:
I ran the test suite with `unittest-ft` and there are *a lot* of errors and failures. Unfortunately the default settings for `unittest-ft` run all the tests simultaneously, which makes it a little difficult for me to understand exactly where the issues are. https://github.com/amyreese/unittest-ft/issues/21 might help.

I see some tracebacks inside `unittest.mock`, so there's probably thread safety issues in the way `mock` is used in the tests.

All that to say, I think I agree with https://github.com/Instagram/LibCST/pull/1295#discussion_r1985334823 that we should focus on multithreaded stress tests rather than trying to use the existing test suite as a stress test - we'll likely end up spending most of our effort on fixing races in the test suite itself.

For now I'll focus on writing a stress test that uses a thread pool to process many files simultaneously.

ngoldbaum:
I doubt it'll be very elucidating (it's 1.5 MB...) but here's the full output from running `unittest-ft` on this branch, after rebasing on `main`: https://gist.githubusercontent.com/ngoldbaum/0390a6afadd0fa8ebd5d5dbefe0cfd88/raw/c6d1a205ffb19240fa7de8807ca92ef3c3082ad2/gistfile1.txt

libcst/codemod/_cli.py
```diff
@@ -607,13 +609,20 @@ def parallel_exec_transform_with_prettyprint(  # noqa: C901
...
 
+    pool_impl: Callable[[], Executor]
     if total == 1 or jobs == 1:
         # Simple case, we should not pay for process overhead.
-        # Let's just use a dummy synchronous pool.
+        # Let's just use a dummy synchronous executor.
         jobs = 1
-        pool_impl = DummyPool
+        pool_impl = DummyExecutor
+    elif getattr(sys, "_is_gil_enabled", lambda: False)():
```
ngoldbaum, line 618:
```suggestion
    elif not getattr(sys, "_is_gil_enabled", lambda: True)():
```

The way you have this written now it'll use a process pool if the GIL is disabled.

That said, it looks like there are some scaling bottlenecks:

```
PYTHON_GIL=1 python3 -m libcst.tool codemod --no-format  .  307.24s user 3.77s system 979% cpu 31.750 total
```

vs

```
PYTHON_GIL=0 python3 -m libcst.tool codemod --no-format  .  561.45s user 1195.46s system 820% cpu 3:34.01 total
```

ngoldbaum: (this is using the format-to-fstring codemod on the numpy repo - I'm gonna make a NumPy PR for that cleanup!)

libcst/codemod/_cli.py
ngoldbaum: After looking at the code a bit, it occurs to me that we could also look at storing state in a [`contextvar.ContextVar`](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar) of a [`threading.local`](https://docs.python.org/3/library/threading.html#thread-local-data).

Using context-local or thread-local state might make sense of it turns out creating a transform for each file is expensive.

Making it so people need to explicitly pass around a `context` along with a `transform` is a big refactor and it implies API changes so that's probably a last resort.

libcst/codemod/_cli.py
zsol: > The way you have this written now it'll use a process pool if the GIL is disabled.

That's the intention. We should only use threadpool if we do have a `sys._is_gil_enabled` function and it returns true. Otherwise fall back to ProcessPool which will be much faster

ngoldbaum: I’m confused, why do you expect a process pool to be faster than a thread pool when the GIL is disabled?

zsol: Boolean algebra is my nemesis, that's why

ngoldbaum: You aren’t the first and won’t be the last to be tripped up by how they decided to spell that in the `sys` API.

zsol: Oh and it's not even my first trip-up! 😆 I consistently mess this up

libcst/codemod/_cli.py
zsol: After coming back to this, I think I only see terrible ways forward:

1. Change the API of `parallel_exec_transform_with_prettyprint` to take a `CodemodCommand` factory, and instantiate it per thread (but with multiprocessing, we'd have to instantiate it once and send the instance over)
2. Change the API of `Codemod` so that `Context` is being passed into `transform_module()` instead of in the constructor. This is a **lot** of churn.
3. Change how `context` is stored inside `Codemod` and add a getter/setter that would call `ContextVar` APIs. I'm a little unclear on how this would work, and a bit worried about perf - that `context` attribute can be pretty hot in a codemod.
4. Screw all of this and add a Threadpool-specific alternative API for `parallel_exec_transform_with_prettyprint` that would do the right thing(tm) but would not support multiprocessing

zsol: I did 1. in #1334

Issue #1242, opened 2024-11-11
Publish cp313t (free-threaded) wheels


ngoldbaum:
Currently `pip install libcst` will do a from-source build on free-threaded 3.13.0:

```
goldbaum at Nathans-MBP in ~/Documents/cryptography on main!
± pip install libcst
Collecting libcst
  Downloading libcst-1.5.0.tar.gz (772 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 773.0/773.0 kB 22.9 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting pyyaml>=5.2 (from libcst)
  Using cached PyYAML-6.0.2-cp313-cp313t-macosx_14_0_arm64.whl
Building wheels for collected packages: libcst
  Building wheel for libcst (pyproject.toml) ... done
  Created wheel for libcst: filename=libcst-1.5.0-cp313-cp313t-macosx_14_0_arm64.whl size=2027094 sha256=ec7f1a08268640717734e28b8454ca23043112fb2f002b46694494c82cf6b6bf
  Stored in directory: /Users/goldbaum/Library/Caches/pip/wheels/af/db/82/f05ed1829e60f978a9af19520c78cb52df1bed762e579d7f84
Successfully built libcst
Installing collected packages: pyyaml, libcst
Successfully installed libcst-1.5.0 pyyaml-6.0.2

[notice] A new release of pip is available: 24.2 -> 24.3.1
[notice] To update, run: pip install --upgrade pip

goldbaum at Nathans-MBP in ~/Documents/cryptography on main!
± python -VV
Python 3.13.0 experimental free-threading build (main, Nov  5 2024, 16:45:19) [Clang 16.0.0 (clang-1600.0.26.3)]
```

The build succeeds, but it probably shouldn't, and immediately segfaults if I try to import anything that imports the native module:

```
$  python
Python 3.13.0 experimental free-threading build (main, Nov  5 2024, 16:45:19) [Clang 16.0.0 (clang-1600.0.26.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import libcst.codegen.gen_matcher_classes as matcher_codegen
[1]    33127 segmentation fault  python
```

[...]

PR #1312, opened 2025-03-12
add free-threaded CI

ngoldbaum:
xref #1242 

passes CI on my fork: https://github.com/ngoldbaum/LibCST/actions/runs/13817416645

Is adding `dev-without-jupyter` in `pyproject.toml` ok? It seemed like the cleanest way to handle avoiding jupyter. You could also split `dev` into `tests` and `docs` but I'm not sure how to sort the dependencies.

The windows 3.13t CI is pretty slow but I think most of that is just installing Python and once the cache populates it should be faster.

I'm pinning to an unreleased version of `setup-python` - hopefully Github does a release with free-threaded support soon.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,51 @@ jobs:
           hatch run coverage combine .coverage.pure
           hatch run coverage report
 
+  # TODO:
+  # merge into regular CI once hatch has support for creating environments on
+  # the free-threaded build: https://github.com/pypa/hatch/issues/1931
+  free-threaded-tests:
+    name: "test (${{ matrix.os }}, 3.13t)"
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -44,6 +44,51 @@ jobs:
...
+        with:
+          cache: pip
+          cache-dependency-path: "pyproject.toml"
+          python-version: '3.13t'
+      - name: Build LibCST
+        run: |
+          # Install build-system.requires dependencies
+          pip install setuptools setuptools-scm setuptools-rust wheel
+          # Jupyter is annoying to install on free-threaded Python
+          pip install -e .[dev-without-jupyter] --no-build-isolation
```
zsol, line 73: oh i just noticed this, would CI still work without `--no-build-isolation`?

PR #1316, opened 2025-03-20
ci: enable macos intel wheels

### LINKED ISSUE #1279: 1.6.0 has no intel macos wheels
Hello,

I encountered an issue when attempting to install libcst==1.6.0 using Python 3.10. The installation fails with the following error message:

```vbnet
error: can't find Rust compiler

If you are using an outdated pip version, it is possible a prebuilt wheel is available for this package but pip is not able to install from it. Installing from the wheel would avoid the need for a Rust compiler.

To update pip, run:

    pip install --upgrade pip

and then retry package installation.

If you did intend to build this package from source, try installing a Rust compiler from your system package manager and ensure it is on the PATH during installation. Alternatively, rustup (available at https://rustup.rs) is the recommended way to download and update the Rust compiler toolchain.
```

This appears to be related to the upgrade of PyO3 to version 0.22 in PR #1180. It seems that the dependency on a Rust compiler was introduced with this change.

This issue makes it impossible to install libcst without a Rust compiler, which could be problematic for environments that do not have Rust installed.

Could you please clarify whether this behavior is intended? If not, is there a plan to provide prebuilt wheels for libcst==1.6.0 that would not require a Rust compiler?

Thank you!

hadialqattan:
## Summary

Basically fixes #1279 by bumping minimum MacOS version from `10.9` (`cibuildwheel` default) to `10.12`.

`Rustc` requires `MACOSX_DEPLOYMENT_TARGET >= 3.12` while `cibuildwheel` sets the `MACOSX_DEPLOYMENT_TARGET` to `10.9` by default, if not specified by the user.

ref 0: [cibuildwheel/macos.py](https://github.com/pypa/cibuildwheel/blob/main/cibuildwheel/macos.py#L285-L306)

ref 1: [rustc/src/platform-support/apple-darwin.md](https://github.com/rust-lang/rust/blob/d8e44b722a93e55cbc9a8188dfbfe3faf1226096/src/doc/rustc/src/platform-support/apple-darwin.md?plain=1#L33C1-L37C10)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,6 @@ jobs:
     env:
       SCCACHE_VERSION: 0.2.13
       GITHUB_WORKSPACE: "${{github.workspace}}"
-      MACOSX_DEPLOYMENT_TARGET: "10.12"
     steps:
       - uses: actions/checkout@v4
         with:
@@ -26,6 +25,10 @@ jobs:
           cache: pip
           cache-dependency-path: "pyproject.toml"
           python-version: "3.12"
+      - name: Set MACOSX_DEPLOYMENT_TARGET for Intel MacOS
+        if: matrix.os == 'macos-13'
+        run: >-
+          echo MACOSX_DEPLOYMENT_TARGET=10.13 >> $GITHUB_ENV
       - name: Disable scmtools local scheme
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
         run: >-
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Set MACOSX_DEPLOYMENT_TARGET for Intel MacOS
         if: matrix.os == 'macos-13'
         run: >-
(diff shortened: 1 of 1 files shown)


hadialqattan:
Hey @zsol,

Kindly, could you help to to finalize this PR?

.github/workflows/build.yml
```diff
@@ -10,10 +10,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, ubuntu-latest, ubuntu-24.04-arm, windows-latest]
+        # macos-13 is an intel runner, macos-latest is apple silicon
+        os: [macos-13, macos-latest, ubuntu-latest, ubuntu-24.04-arm, windows-latest]
     env:
       SCCACHE_VERSION: 0.2.13
       GITHUB_WORKSPACE: "${{github.workspace}}"
+      MACOSX_DEPLOYMENT_TARGET: "10.12"
```
zsol, line 18: We should only do this on the Intel runners, right?

hadialqattan:
Btw, should I document this somewhere (e.g. changelog)?

zsol:
> Btw, should I document this somewhere (e.g. changelog)?

I'll take care of that. Thank you for the PR!

PR #1323, opened 2025-04-08
Don't share transforms between thread pool tasks

ngoldbaum:
Note that this PR targets the `pr1295` branch.

Unfortunately I'm still seeing a substantial slowdown using threading compared with multiprocessing. I don't see anything obvious in a profile generated from `samply`. I want to try to test with Python 3.14 to see if it's a performance issue that's been fixed upstream already. I'll report back if I find anything, I might need to update the rust code a little to work with the latest version of PyO3.

### DIFF
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -627,7 +627,6 @@ def parallel_exec_transform_with_prettyprint(  # noqa: C901
         pool_impl = DummyExecutor
     elif not getattr(sys, "_is_gil_enabled", lambda: True)():
         from concurrent.futures import ThreadPoolExecutor
-
         pool_impl = functools.partial(ThreadPoolExecutor, max_workers=jobs)
     else:
         from concurrent.futures import ProcessPoolExecutor
--- a/libcst/codemod/_cli.py
+++ b/libcst/codemod/_cli.py
@@ -8,19 +8,30 @@ Provides helpers for CLI interaction.
 """
 
 import difflib
+import functools
 import os.path
 import re
-import functools
 import subprocess
 import sys
 import time
 import traceback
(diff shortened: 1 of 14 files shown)


libcst/codemod/_cli.py
```diff
@@ -215,11 +226,13 @@ class ExecutionConfig:
 
 
 def _execute_transform(  # noqa: C901
-    transformer: Codemod,
+    codemod_class: Type[Codemod],
```
ngoldbaum, line 229: pyre complains with:

```
libcst/codemod/_cli.py:234:18 Invalid class instantiation [45]: Cannot instantiate abstract class `Codemod` with abstract method `transform_module_impl`.
```

Not totally sure how to fix that.

ngoldbaum:
I finally managed to get a usable profile to play with. It turns out `samply` doesn't handle running under a VM very well.

Here's the profile: https://share.firefox.dev/4jq1Bv2

We seem to be getting a lot of lock contention in several methods in `CSTTransformer` as well as `MatcherDecoratableTransformer`.

@nascheme, who has been spending a lot of time in `typeobject.c` lately, tells me that the calls to `PyType_LookupRef` are suspicious and maybe LibCST is defeating an optimization in CPython by getting dynamically generated function names at runtime. If that happens concurrently, there's a chance you'll need to acquire a global lock that all python types share.

I'll need to understand how the visitor pattern is used inside libcst to know whether it makes any sense to refactor libcst rather than report a bug to CPython, but I wonder if we can use thread-local state instead of touching global type objects to avoid the contention inside the interpreter.

(also ping @colesbury since this might be interesting to you)

ngoldbaum:
This is now diverging quite a bit so I'm going to close it and reopen targeting the `main` branch

PR #1324, opened 2025-04-09
update pyo3 to 0.25

ngoldbaum:
Fixing the deprecations is needed to try testing with Python 3.14 on a pending PR to pyo3's main branch.


amyreese:
Rebased onto #1330

PR #1331, opened 2025-05-19
Run CI tests on 3.14

amyreese:
## Summary

Adds 3.14 to the CI matrix

## Test Plan

Look at CI results


zsol:
Looks like the tests are failing

ngoldbaum:
I can't reproduce the crash on the Mac build but I do see the failure the other runners also saw.

The exception comes from here:

https://github.com/Instagram/LibCST/blob/be0b668d089f1e189ebd122f0f7beca31c8d8867/libcst/codegen/gen_matcher_classes.py#L394-L400

If I modify that error to tell me about the type that's getting passed in, I get

```
Exception: Logic error, unexpected top level type! Found BinaryOperation(
    left=SimpleString(
        value="'BaseExpression'",
        lpar=[],
        rpar=[],
    ),
    operator=BitOr(
        whitespace_before=SimpleWhitespace(
            value=' ',
        ),
        whitespace_after=SimpleWhitespace(
            value=' ',
        ),
    ),
    right=Name(
        value='None',
        lpar=[],
        rpar=[],
    ),
    lpar=[],
    rpar=[],
) with type <class 'libcst._nodes.expression.BinaryOperation'>
```

In a debugger, going up the call stack, it's dying trying to parse the type string `'libcst._nodes.expression.BaseExpression | None'`.

I don't know enough about libcst to have any idea why this only happens on 3.14.

zsol:
Looks like this is a genuine bug, [deferred annotations](https://peps.python.org/pep-0649/) is causing the type string to appear as `BaseExpression | None` but in 3.13 and before, it used to be `Optional[BaseExpression]`. I'll look at fixing this

zsol:
#1336 should help

zsol (approved):

PR #1353, opened 2025-06-06
Upgrade PyYAML-ft version and use new module name

lysnikolaou:
## Summary

Upgrade to PyYAML-ft 8.0.0 (not yet released) and use new `yaml_ft` module name.


zsol:
Thanks for the quick help! This is now released in libcst version 1.8.1

PR #1358, opened 2025-06-11
fix(dependency): add back typing-extensions for 3.9

Lee-W:
## Summary
Missing typing-extensions breaks "from libcst.codemod import CodemodContext"

This can be verified by 

```shell
uv run --python 3.9 --with libcst==1.8.1 python -c "from libcst.codemod import CodemodContext"
```

it raises the following error

```
    from typing_extensions import ParamSpec
ModuleNotFoundError: No module named 'typing_extensions'
```

while it works fine for 3.10

```shell
uv run --python 3.10 --with libcst==1.8.1 python -c "from libcst.codemod import CodemodContext"
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ requires-python = ">=3.9"
 dependencies = [
     "pyyaml>=5.2; python_version < '3.13'",
     "pyyaml-ft>=8.0.0; python_version >= '3.13'",
+    "typing-extensions; python_version < '3.10'",
 ]
 
 [project.urls]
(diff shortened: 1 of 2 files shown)


zsol:
Hmm looks like this was broken in 1.8.0, but why didn't the tests catch it?

zsol:
Ah I see, the test dependencies transitively include `typing-extensions`. Unfortunate.

zsol:
Thanks for the fix!

zsol:
I've put up https://github.com/Instagram/LibCST/pull/1359 to make sure this never happens again

Lee-W:
Thanks for helping us merge it! It seems we excluded 1.8.0 last time (I wasn't aware), and this time 1.8.1 breaks my PR, which is why I find it. Glad to know 1.8.2 would work fine!

PR #1359, opened 2025-06-12
ci: test built wheels

zsol:
This PR does two things:
1. makes sure the built wheels are exercised (by running a simple codemod on libcst sources themselves)
2. builds the wheels and runs the above tests on PRs to get early feedback


zsol:
This crash on 3.14 looks like it's because of https://github.com/PyO3/pyo3/issues/5175

Issue #1322, opened 2025-04-07
PositionProvider does not provide position for `match` statements


Otto-AA:
I'm using a `PositionProvider` to get the lines of all nodes in a Visitor. My assumption would be that the metadata exists for all nodes, however when visiting a `match` statement this is not true.

The following program results in a `KeyError: SimpleWhitespace(value='',)` on the metadata access:

```python
import libcst as cst
from libcst.metadata import PositionProvider, MetadataWrapper

def main():
    code = """
match status:
    case 400:
        return "Bad request"
    case 404:
        return "Not found"
    case 418:
        return "I'm a teapot"
    case _:
        return "Something's wrong with the internet"
"""
    module = cst.parse_module(code)

    metadata_wrapper = MetadataWrapper(module)
    visitor = Visitor()
    module = metadata_wrapper.visit(visitor)


class Visitor(cst.CSTVisitor):
    METADATA_DEPENDENCIES = (PositionProvider,)

    def on_visit(self, node):
        print(self.get_metadata(PositionProvider, node).start.line)
        return super().on_visit(node)



if __name__ == "__main__":
    main()
```

I'm using libcst version 1.7.0.

zsol:
Looks like this is a bug, PositionProvider needs to be updated with match and a few other new CST nodes

PR #1369, opened 2025-07-01
fixes match statements to work with PositionProvider

patreon-steve:
## Summary

Fixes [issue 1322](https://github.com/Instagram/LibCST/issues/1322) where `PositionProvider` does not provide positions for whitespace. The issue seems to be that when there is no pattern for whitespace nodes during codegen, then they are not included in the position metadata.

## Test Plan

- Added a test
- Verified this test failed
- Implemented fix and verified fix

### DIFF
--- a/libcst/_nodes/statement.py
+++ b/libcst/_nodes/statement.py
@@ -2886,6 +2886,9 @@ class MatchCase(CSTNode):
                 state.add_token("if")
                 self.whitespace_after_if._codegen(state)
                 guard._codegen(state)
+            else:
+                self.whitespace_before_if._codegen(state)
+                self.whitespace_after_if._codegen(state)
 
             self.whitespace_before_colon._codegen(state)
             state.add_token(":")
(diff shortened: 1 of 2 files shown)


drinkmorewaterr:
Hi @patreon-steve,
Thanks for the PR  :) In general it lgtm ! 

However there are a few failing tests, it seems like the python parser doesn't support match statements.
Luckily for us its marked for removal anyways... give me a couple days to see if we can either
1. Remove the old python parser completely
2. Add a way to mark tests as 'native-only'

patreon-steve:
> Hi @patreon-steve, Thanks for the PR :) In general it lgtm !
> 
> However there are a few failing tests, it seems like the python parser doesn't support match statements. Luckily for us its marked for removal anyways... give me a couple days to see if we can either
> 
> 1. Remove the old python parser completely
> 2. Add a way to mark tests as 'native-only'

Awesome. Thank you

drinkmorewaterr:
Hi @patreon-steve,
Sorry for the delay, can you try rebasing this PR on master :)

PR #1379, opened 2025-07-16
Fix `noqa` comments

thomas-serre-sonarsource:
## Summary

At Sonar, we are working on suppressing linter issues on lines containing comments starting with `noqa`.

Doing so, and testing it on your source code, it seems that two of the `noqa` comments are missing `:` before the rule ID. 
Then, all issues on these lines are suppressed, rather than only the issue with the specific rule ID. I guess this was not the intent, and I propose to correct it.


zsol:
Whoops I accidentally marked this as ready for review,sorry. In any case, thanks for this, lgtm

PR #1389, opened 2025-08-04
fixes match statements to work with PositionProvider

imsut:
## Summary

(This is a rebase of https://github.com/Instagram/LibCST/pull/1369)

Fixes https://github.com/Instagram/LibCST/issues/1322 where PositionProvider does not provide positions for whitespace. The issue seems to be that when there is no pattern for whitespace nodes during codegen, then they are not included in the position metadata.

## Test Plan

* Added a test
* Verified this test failed
* Implemented fix and verified fix


imsut:
@drinkmorewaterr -- I'd like to try this fix. Is there a plan to cut a release v1.8.3?

PR #1454, opened 2026-05-29
add support for Python 3.15

AlexWaygood:
## Summary

Add support for running and building LibCST on Python 3.15 and 3.15t.

PyO3 0.28 is needed to build the native extension on 3.15, and requires bumping the minimum Rust version to 1.83. This also updates uv and cibuildwheel for the new prerelease interpreters.

Finally, refresh coverage.py: the previously locked version crashes while tracing tests on 3.15.

Following the precedent set in [#1203, #1331 and #1345, this is a CI-focussed change that does not add support for parsing new Python 3.15 syntax. That would need to be done separately.

Fixes #1445.

## Test plan

I ran all tests that I could locally; CI should hopefully confirm whether this looks in good shape!

PR #1455, opened 2026-06-01
bump pyo3 to 0.28, swap PyObject for Py<PyAny> in derive macro

thatch:
## Summary

It appears that pyo3 0.26 is incompatible with 3.15t, so this updates and now it builds.

## Test Plan

Wait for CI


thatch:
I don't know how I missed #1454 existing when I created this.

PR #1465, opened 2026-07-31
Allow FlattenSentinel in matcher decorator return annotations

### LINKED ISSUE #1075: `FlattenSentinel` can't be used as return type annotation when parametrized.
Consider the following code:

```python
MATCHER = m.FunctionDef(name=m.Name("__init__"))

class MyVisitor(...):
    ...
    @m.leave(MATCHER)
    def mutate_init(
        self,
        original_node: cst.FunctionDef,
        updated_node: cst.FunctionDef
    ) -> cst.FlattenSentinel[cst.FunctionDef]:
        ...
```

This raises the following exception: `TypeError: issubclass() arg 1 must be a class`

https://github.com/Instagram/LibCST/blob/43a27b1222993f29086748ef6a08a57801c536aa/libcst/matchers/_visitors.py#L155-L157

Here, `ret` is `cst.FlattenSentinel[cst.FunctionDef]` (an instance of `GenericAlias`), `annotation` is `libcst._nodes.statement.BaseStatement`.

Works as expected without parametrizing the class

Sanjays2402:
Closes #1075

## Summary

Annotating a `@leave`-decorated method with `FlattenSentinel` raised `TypeError: issubclass() arg 1 must be a class`, even though returning a `FlattenSentinel` from `leave_<Node>` is valid.

Two causes: the generated `TYPED_FUNCTION_RETURN_MAPPING` omitted `FlattenSentinel` entirely (unlike the `leave_` signatures `gen_visitor_functions` emits), and `_get_possible_annotated_classes` passed the parameterized generic `FlattenSentinel[BaseStatement]` straight to `issubclass()`, which only accepts classes. `gen_type_mapping` now mirrors `gen_visitor_functions` for sequence-valued nodes, and generic aliases are unwrapped to their origin class before the subclass check. `_return_types.py` is regenerated output.

## Test Plan

Two tests in `libcst/matchers/tests/test_visitors.py` (parameterized and bare `FlattenSentinel`) fail on main with the reported `TypeError` and pass with the fix. `libcst/matchers/tests` 145 passed / 1 skipped; `libcst/codemod` unchanged from baseline; `ufmt check` clean on the touched files.

This change was prepared with AI assistance; the regression tests were run locally and fail without the fix.


Sanjays2402:
Gentle ping — CI here is still pending workflow approval, so the checks haven't run yet. If a maintainer could kick that off I'd appreciate it. Glad to make changes if the approach needs adjusting.

PR #1466, opened 2026-07-31
fix: key inherited metadata dependency cache per class

### LINKED ISSUE #1459: Bare CSTTransformer metadata pass poisons the inherited-dependency cache for all later subclasses
## Summary

`MetadataDependent.get_inherited_dependencies` memoizes into a plain class attribute and reads it back with ordinary attribute lookup. Because that lookup walks the MRO, running a metadata pass with a **bare** `CSTTransformer`/`CSTVisitor` caches an empty `frozenset` onto the base class, and every subclass that has not yet computed its own cache then inherits that empty set — raising `METADATA_DEPENDENCIES` `KeyError`s even though the subclass declares the provider correctly.

The poisoning is global and permanent for the process, and it surfaces arbitrarily far from the cause.

## Reproduction

```python
import libcst as cst
from libcst.metadata import MetadataWrapper, PositionProvider

class PositionAware(cst.CSTTransformer):
    METADATA_DEPENDENCIES = (PositionProvider,)
    def leave_Pass(self, original_node, updated_node):
        self.get_metadata(PositionProvider, original_node)
        return updated_node

def run(visitor):
    MetadataWrapper(cst.parse_module("pass\n")).visit(visitor)

run(PositionAware())          # fine
run(cst.CSTTransformer())     # <-- poisons the base class
run(PositionAware())          # KeyError
```

```
KeyError: 'PositionProvider is not declared as a dependency in PositionAware.METADATA_DEPENDENCIES.'
```

Observed on libcst 1.8.6 (current release) with CPython 3.14.6; the same code is present on `main`.

## Cause

```python
try:
    return cls._INHERITED_METADATA_DEPENDENCIES_CACHE
except AttributeError:
    ...
    cls._INHERITED_METADATA_DEPENDENCIES_CACHE = frozenset(dependencies)
    return cls._INHERITED_METADATA_DEPENDENCIES_CACHE
```

`cls._INHERITED_METADATA_DEPENDENCIES_CACHE` resolves through the MRO, so the `try` succeeds against a *base class* entry. Resolving `CSTTransformer` itself writes `frozenset()` there, since the base declares no dependencies. Every later subclass without its own entry reads it.

Note the trigger is **resolution**, not construction — `cst.CSTTransformer()` alone is harmless; the cache is only written when `get_inherited_dependencies()` runs during a metadata pass.

## Impact

This is hard to diagnose in a test suite. One helper that ran a bare transformer broke 38 tests across five unrelated files in our codebase, and bisecting by file or by single test finds nothing, because the poison needs one specific test to have run first and is then permanent for the interpreter. The reported error also points at the innocent subclass, which reads as a bug in that class rather than in shared state.

## Suggested fix

Read only the class's own entry, so a base-class cache can never be inherited:

```python
@classmethod
def get_inherited_dependencies(cls) -> Collection["ProviderT"]:
    cache = cls.__dict__.get("_INHERITED_METADATA_DEPENDENCIES_CACHE")
    if cache is None:
        dependencies = set()
        for c in inspect.getmro(cls):
            if issubclass(c, MetadataDependent):
                dependencies.update(c.METADATA_DEPENDENCIES)
        cache = frozenset(dependencies)
        cls._INHERITED_METADATA_DEPENDENCIES_CACHE = cache
    return cache
```

This keeps the memoization and its per-class semantics, and costs one dict lookup. Happy to send this as a PR if the approach looks right.

Sanjays2402:
Closes #1459

## Summary

`MetadataDependent.get_inherited_dependencies()` memoized into a plain class attribute and read it back with ordinary attribute lookup. That lookup walks the MRO, so resolving a bare `CSTTransformer`/`CSTVisitor` cached an empty `frozenset` onto the base class, and every subclass that had not yet computed its own cache inherited it — raising `METADATA_DEPENDENCIES` `KeyError`s even though the subclass declares the provider correctly. Reading the cache from `cls.__dict__` instead means each class only ever sees its own memoized value.

## Test Plan

Added `test_bare_visitor_does_not_poison_subclass_dependencies` to `libcst/metadata/tests/test_metadata_provider.py`, which resolves a bare `CSTTransformer` first and then asserts the subclass still reports its declared dependency. It fails on the unpatched source (`frozenset() != {ProviderA}`) and passes with the fix; `libcst/metadata/tests/`, `test_visitor.py` and `test_batched_visitor.py` are green (178 passed, only the pre-existing pyre-binary errors in `test_type_inference_provider.py`).

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/libcst/_metadata_dependent.py
+++ b/libcst/_metadata_dependent.py
@@ -82,17 +82,21 @@ class MetadataDependent(ABC):
 
         Recursively searches the MRO of the subclass for metadata dependencies.
         """
-        try:
-            # pyre-fixme[16]: use a hidden attribute to cache the property
-            return cls._INHERITED_METADATA_DEPENDENCIES_CACHE
-        except AttributeError:
-            dependencies = set()
-            for c in inspect.getmro(cls):
-                if issubclass(c, MetadataDependent):
-                    dependencies.update(c.METADATA_DEPENDENCIES)
-            # pyre-fixme[16]: use a hidden attribute to cache the property
-            cls._INHERITED_METADATA_DEPENDENCIES_CACHE = frozenset(dependencies)
-            return cls._INHERITED_METADATA_DEPENDENCIES_CACHE
+        # Read the cache out of ``cls.__dict__`` rather than with attribute
+        # lookup: a plain attribute read walks the MRO, so a value cached on a
+        # base class would be returned for every subclass that has not computed
+        # its own cache yet.
+        cache = cls.__dict__.get("_INHERITED_METADATA_DEPENDENCIES_CACHE")
+        if cache is not None:
+            return cache
+        dependencies = set()
+        for c in inspect.getmro(cls):
+            if issubclass(c, MetadataDependent):
+                dependencies.update(c.METADATA_DEPENDENCIES)
+        cache = frozenset(dependencies)
+        # pyre-fixme[16]: use a hidden attribute to cache the property
+        cls._INHERITED_METADATA_DEPENDENCIES_CACHE = cache
+        return cache
 
     @contextmanager
     def resolve(self, wrapper: "MetadataWrapper") -> Iterator[None]:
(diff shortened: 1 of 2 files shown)


Sanjays2402:
Superseded by #1468, which fixes the same issue (#1459) with a tighter diff. Closing this one to avoid duplicate review effort — sorry for the noise.

PR #1467, opened 2026-08-02
Respect overwrite_existing_annotations for attributes and globals

### LINKED ISSUE #1159: `ApplyTypeAnnotationsVisitor` does not respect `overwrite_existing_annotations` for attributes
Consider this example:

```py
class X:
    x: int
```

used to patch:

```py
class X:
    x: str
```

`ApplyTypeAnnotationsVisitor` will *not* change the type annotation on `X.x` to `int` even if you set `overwrite_existing_annotations` to True. This is the case even if you change it to be an assign (as in `x: str = "a"`.)

Sanjays2402:
Closes #1159

## Summary

`ApplyTypeAnnotationsVisitor` only overwrote existing annotations on function parameters and returns. An attribute or global that already carries an annotation is an `AnnAssign`, which the transformer never visited, so `overwrite_existing_annotations=True` had no effect there — patching `class X: x: str` with a stub `class X: x: int` left `x: str` unchanged.

Adds `leave_AnnAssign`, which (only when overwriting is requested) looks up the qualified name in the stub's attribute annotations and swaps in the stub annotation, recording it in the annotation counts.

## Test Plan

New `test_annotate_attributes_with_existing_annotations` in `libcst/codemod/visitors/tests/test_apply_type_annotations.py` covers a class attribute, an attribute with a value, and a module-level global. All three fail without the source change and pass with it; `libcst/codemod/` is otherwise unchanged (389 passed, with one pre-existing `test_codemod_formatter_error_input` failure present on a clean checkout too).

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/libcst/codemod/visitors/_apply_type_annotations.py
+++ b/libcst/codemod/visitors/_apply_type_annotations.py
@@ -1266,6 +1266,35 @@ class ApplyTypeAnnotationsVisitor(ContextAwareTransformer):
         else:
             return self._annotate_single_target(original_node, updated_node)
 
+    def leave_AnnAssign(
+        self,
+        original_node: cst.AnnAssign,
+        updated_node: cst.AnnAssign,
+    ) -> cst.AnnAssign:
+        # An attribute or global that already carries an annotation is only
+        # touched when we are explicitly told to overwrite existing annotations.
+        if not self.overwrite_existing_annotations:
+            return updated_node
+        target = original_node.target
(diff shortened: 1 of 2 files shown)


Sanjays2402:
Ping — this one's waiting on workflow approval before CI can run. Would appreciate a maintainer enabling the run whenever there's a spare moment. Open to feedback on the approach.

